PR #2, opened 2025-08-01
[spring-ecommerce-order - Step-1] Submit Vito's mission

VittorioDeMarzi:
# Spring E-commerce Order - Step 1 PR Overview

Hi PK 👋,

This PR is for **Step 1** of the `spring-ecommerce-order` mission. It builds on the previous `spring-ecommerce-product` project by refactoring the codebase to use Spring Data JPA instead of JDBC, adding pagination and sorting, and implementing product options with all required validations and admin functionality.

Below is a quick overview of the features implemented, followed by some questions where I'd really appreciate your feedback.

## Title
`Step 1 - Entity Mapping, Pagination, Product Options`

### Refactor JDBC → Spring Data JPA
- All CRUD operations now use Spring Data JPA with dedicated repositories
- Entity relationships mapped using `@OneToMany`, `@ManyToOne`, etc.

### Entity Mapping
- Created entities `Product`, `Option`, `Cart`, `CartItem`, `Member` with direct references between them
- Conversion between entities and DTOs implemented via `toDto()` and `toEntity()` methods

### Pagination & Sorting
- Implemented pagination for the product list and wishlist, with support for multiple sorting fields and ascending/descending order

### Product Options
- Added option management for each product (name, quantity, validation constraints)
- Implemented all validation rules (allowed characters, min/max quantity, unique name per product)
- Added method to decrease option quantity directly in the entity (reusable in future use cases)
- Admin interface to add new options via `/add/option/{productId}`

### 1. E2E Tests & Data.sql
Currently, all end-to-end tests rely on `data.sql` (H2 in-memory).
- Is it fine to keep it this way for now?
- When migrating to PostgreSQL/MySQL, is it considered good practice to still have a dedicated test environment with an H2-based data file?

### 2. Entity ↔ DTO Mapping
I'm handling conversion using `toDto()` and `toEntity()` **directly inside entity classes** (e.g., `Product`, `CartItem`, `Option`).
- Is this considered good practice, or would it be better to centralize mapping in separate mapper classes?

### 3. Product Creation Approach
Currently creating a `Product` with its `Options` directly from the DTO:

```java
productJpaRepository.save(productRequest.toEntity())
```

- Do you think this is a clean approach, or should I follow a different pattern to better separate responsibilities?

### 4. @SqlResultSetMapping / @NamedNativeQuery for statistics
I mapped native query results directly to DTOs using these annotations.
- I'm sure there are more modern or optimized approaches. Could you give me some hints on what to look into?

### 5. Validation on Entities
If I'm already validating request DTOs with `@Valid`, does it make sense to also validate the entities themselves?

### 6. Testing
I'm still missing some unit tests for Controllers and Services, as well as integration tests. I'll spend time this weekend working on them.

---

Thanks a lot for your time and any detailed feedback! 🙏

**Vittorio**


src/main/resources/data.sql
```diff
@@ -0,0 +1,55 @@
+INSERT INTO PRODUCT (name, price, image_url)
```
pkeugine, line 1: **Q. Currently, all end-to-end tests rely on data.sql (H2 in-memory). Is it fine to keep it this way for now?**

`data.sql` isn't limited to H2, it works with other database vendors too, as long as the SQL syntax is compatible. If the concern is "will this approach stop working if we switch to a different database?" then no worries, it's still valid. I'm actually curious what prompted the question. If you're thinking about changing something, it'd be great to discuss that. It might open up a chance to align or improve the setup.

src/main/resources/data.sql
```diff
@@ -0,0 +1,55 @@
...
+VALUES ('Espresso', 2.50, 'https://upload.wikimedia.org/wikipedia/commons/4/45/A_small_cup_of_coffee.JPG'),
```
pkeugine, line 2: **Q. When migrating to PostgreSQL/MySQL, is it considered good practice to still have a dedicated test environment with an H2-based data file?**

What are the other options we could choose from, and what do you think best fits your needs? Let's talk about it. 🙂  And just like my previous comment, if the SQL syntax is compatible, `data.sql` isn't limited to H2.

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,22 @@
...
+    val email: String,
+    val password: String,
+    val role: String = "USER",
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    fun toDto(): MemberDto {
+        return MemberDto(id, email, role)
+    }
```
pkeugine: **Q. I'm handling conversion using `toDto()` and `toEntity()` directly inside entity classes. Is this considered good practice, or would it be better to centralize mapping in separate mapper classes?**

Interesting! What led you to put the conversion logic inside the entity classes?

Also, if you're considering moving it to separate mapper classes, is there a specific reason like code reusability? Would be great to understand your thought process so we can weigh the trade-offs together.

src/main/kotlin/ecommerce/dto/ProductRequest.kt
```diff
@@ -0,0 +1,39 @@
...
+    val options: List<OptionDto>,
+)
+
+fun ProductRequest.toEntity(): Product {
+    return Product(
+        name = name,
+        price = price,
+        imageUrl = imageUrl,
+        options = options.map { it.toEntity() }.toMutableList(),
+    )
```
pkeugine: **Q. Currently creating a Product with its Options directly from the DTO. Do you think this is a clean approach, or should I follow a different pattern to better separate responsibilities?**

I get the sense this builds on the previous question!

One fun thought I’ve had is that `toEntity()` and `toDto()` actually serve very different roles, especially when you consider where they’re typically used in a layered architecture.

I don’t want to spoil my full take just yet! I’d love for you to reflect on it first, but here’s a hint :

> “Domain logic should not be affected by changes in controllers or views.”
> So… if that's the case, should the domain have any dependency on controller-level structures like DTOs?

Just like the previous question, I’d love to hear what led you to place `toEntity()` inside the DTO (and I saw you're using an extension function! 😮 👍 ). Curious what problem you were solving, or what felt clean to you about that approach.

src/main/kotlin/ecommerce/model/CartHistory.kt
```diff
@@ -0,0 +1,79 @@
...
+        ConstructorResult(
+            targetClass = TopProductStats::class,
+            columns = [
+                ColumnResult(name = "productName", type = String::class),
+                ColumnResult(name = "addedCount", type = Int::class),
+                ColumnResult(name = "lastAddedAt", type = Timestamp::class),
+            ],
+        ),
+    ],
+)
```
pkeugine: **Q. I mapped native query results directly to DTOs using these annotations: `@SqlResultSetMapping` / `@NamedNativeQuery` for statistics. I'm sure there are more modern or optimized approaches. Could you give me some hints on what to look into?**

I'm always curious about the thought process behind these decisions about what led you to go with this approach!

From my point of view, since the query doesn’t seem to rely on any database-specific features, we could likely achieve the same result using the `@Query` annotation (even for native queries). It’s more concise and plays well with constructor-based DTOs in Spring Data JPA.

And hey, we don’t *always* have to use JPA either! We’ve got another solid tool in our kit: `JdbcTemplate` 😎  

It might not **look as cool**, but it’s widely used in production projects thanks to its `predictability` and `ease of maintenance`, especially for reporting or aggregation-heavy queries.

src/main/kotlin/ecommerce/dto/ProductPatchRequest.kt
```diff
@@ -0,0 +1,20 @@
...
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Positive
+import jakarta.validation.constraints.Size
+
+data class ProductPatchRequest(
+    @field:NotBlank
+    @field:Size(min = 1, max = 15, message = "Name must have at maximum 15 characters")
+    @field:Pattern(regexp = "^[a-zA-Z0-9()\\[\\]+\\-&/_ ]+$", message = "Invalid characters")
+    var name: String? = null,
```
pkeugine, line 12: **Q. If I'm already validating request DTOs with `@Valid`, does it make sense to also validate the entities themselves?**

I would like you to tell me your take on this first! Since this topic doesn't have a one perfect answer, I think it's important for you to build your own thoughts around it. Which also means you can tell me your opinion without pressure :)

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
```diff
@@ -0,0 +1,164 @@
+package ecommerce.endToEnd
```
pkeugine: Let's change the package name so [it follows the general convention](https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names).
Though I remember thinking that it's a bit weird after seeing this convention for the first time, it's pretty interesting if we find out the motive behind it. It's even in [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-6.html)!

src/main/kotlin/ecommerce/repository/CartHistoryJpaRepository.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import ecommerce.dto.ActiveUsersResponse
+import ecommerce.dto.TopProductStats
+import ecommerce.model.CartHistory
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+
+interface CartHistoryJpaRepository : JpaRepository<CartHistory, Long> {
+    @Query(nativeQuery = true)
+    fun getTopProducts(): List<TopProductStats>
```
pkeugine: I see you've used `@Query` 👍 
Just like my comment above, why don't we utilize this annotation to it's full potential?
There's an option in `@Query` annotation called `value` where you can put SQL in it. We'll be able to maintain queries here if we use it.

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+                .then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.OK.value())
+
+        val productName = response.body().jsonPath().getString("name")
+        assertThat(productName).isEqualTo("updatedTest")
+    }
+
+    @Test
+    fun deleteProduct() {
```
pkeugine: There's a failing test in this PR.

Screenshot 2025-08-02 at 3 55 26 PM

Before making a PR, make sure to check :  
* the project builds successfully
* there's no failing tests

I would like you to fix this asap.

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,33 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+import jakarta.persistence.OneToOne
+
+@Entity
+data class Cart(
+    @OneToOne(cascade = [CascadeType.PERSIST])
+    val member: Member,
+    @OneToMany(cascade = [CascadeType.ALL], mappedBy = "cart")
+    val cartProducts: MutableList<CartItem> = mutableListOf<CartItem>(),
```
pkeugine:
```suggestion
    val cartProducts: MutableList<CartItem> = mutableListOf(),
```

We can let this be implicit :)
Since this topic came up, it would be great to learn about generics if you have the time.
Generics in Kotlin could be quite hard, but it's used in logics very often.

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,33 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
```
pkeugine, line 19: So many new kinds of annotations were added as we migrated to JPA right? 🥲 
You did well adding in `@Id` and `@GeneratedValue` here.

I'm curious, what does `GenerationType.IDENTITY` do, and what are the other options?
Also, what made you choose `0L` to be the default value for id? I guess there are two things we could choose from : `null` and `0L`. If we chose null, id would be `Long?`.

src/main/kotlin/ecommerce/model/Cart.kt
pkeugine: I see you've used bidirectional mapping between `Cart` and `CartItem`!
What were the pros and cons in this decision? I'd like to know your experience :)

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,37 @@
...
+    var cart: Cart,
+    @ManyToOne
+    @JoinColumn(name = "product_id")
+    val product: Product,
+    var quantity: Int,
+    @CreationTimestamp
+    var createdAt: LocalDateTime = LocalDateTime.now(),
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
```
pkeugine:
```suggestion
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    val id: Long = 0L,

    @ManyToOne
    @JoinColumn(name = "cart_id")
    var cart: Cart,

    @ManyToOne
    @JoinColumn(name = "product_id")
    val product: Product,

    var quantity: Int,

    @CreationTimestamp
    var createdAt: LocalDateTime = LocalDateTime.now(),
```

I personally prefer keeping `id` field on the very top, since that (to me) is the most important value when it comes to entity.

Also, when there's multiple annotations added, I like to keep some line breaks in between fields, as it helps increase the readability. (this is also just my opinion 😂 )

src/main/kotlin/ecommerce/model/Cart.kt
pkeugine: Two different `CascadeType`s used! What are the differences between them, and what made you choose these for `member` and `cartProducts`?

src/main/kotlin/ecommerce/repository/ProductJpaRepository.kt
```diff
@@ -0,0 +1,19 @@
...
+    findByIdOrNull(id)
+        ?: throw ProductNotFoundException("Product not found")
+
+fun ProductJpaRepository.existsByNameOrThrow(name: String) {
+    if (existsByName(name)) throw ProductAlreadyInDBException("Product already exists with name: $name")
+}
+
+interface ProductJpaRepository : JpaRepository<Product, Long> {
+    fun existsByName(name: String): Boolean
+}
```
pkeugine: I noticed that `MemberJpaRepository` returns `Member?` when looking for the member entity, while some other repositories like `ProductJpaRepository` handles not-found logic inside itself. I'm curious about this decision!

As for me, I would actually unify the way repositories handle not-found logic; for example, I would return `Product?` in this repository just like `MemberJpaRepository`, and keep it that way for all other ones.

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    fun getCartItems(
+        memberId: Long,
+        pageable: Pageable,
+    ): Page<CartItemResponse> {
+        val cart = cartJpaRepository.getByMemberId(memberId)
+        val products = cart.cartProducts
+        val start = pageable.offset.toInt()
+        val end = Math.min((start + pageable.pageSize), products.size) // replace by coerce kotlin?
```
pkeugine: Why don't we replace this with `coerceAtMost`?
It would be great to remove the comment even if we don't replace it, since we wouldn't want this commit to be merged.

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,68 @@
...
+        return ResponseEntity.ok(addToCartResult)
+    }
+
+    @GetMapping
+    fun getCartItems(
+        @LoginMember member: MemberDto,
+        @RequestParam(defaultValue = "0") page: Int,
+        @RequestParam(defaultValue = "5") size: Int,
+        @RequestParam(defaultValue = "id") sortBy: String,
+        @RequestParam(defaultValue = "true") ascending: Boolean,
```
pkeugine, line 42: It's great to see default values backing up the requests 👍

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,56 @@
...
+        val productsInPage = products.subList(start, end)
```
pkeugine: Since `cartProducts` in `Cart` is a `mutableList` and we're `subList`ing it, it would be great if we could some how make them immutable once we made the list of our needs. It's always great to keep things immutable as much as we can. In other words, let them be immutable as soon as we're done mutating it.

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    @Test
+    fun save() {
+        val option = Option("test", 10)
+        val expected = Product("test", mutableListOf(option), 10.0, "https://google.com")
+        val actual = productJpaRepository.save(expected)
+        assertThat(actual.id).isNotNull
+        assertThat(actual.id).isNotZero
+        assertThat(actual.name).isEqualTo(expected.name)
+    }
```
pkeugine: Nice learning test! 👍 

To make it even more better, why don't we assert option as well?
Since we used a special data for option, it would be great to check if that value was also persisted as well :)

pkeugine (changes requested):
Hello Vito 🙌 
Migrating to JPA must have been a challenging work, but I'm glad you've managed to make it work well 👍 

To make it perfect, I would like you to check the tests, since I see a failing one currently.

Other than that, it's all good.
I left some questions, answers to your questions, and some feedbacks that we can refactor. There are some 'ask back's to your questions, and this is because most of the questions you've asked were very interesting. It would be great to build your own thoughts on it first, and then have a chat with me. I believe this would be a more meaningful mission that way :)

If you have any questions or opinions, feel free to DM me or leave a comment here.
Happy Hacking! 🔥 🔥 🔥 

**UPDATE :** I would also like you to check [PR that Alex requested](https://github.com/herotechcourse/spring-ecommerce-order/pull/7), since I'm the reviewer for Alex as well. Since you two have the exact same codebase, I've extended the review. Please apply the suggestion I put up there, and if you want to talk about the topics that I mentioned there, feel free to leave a comment in this PR. Cheers!

src/main/resources/data.sql
VittorioDeMarzi: The reason I asked is because I was thinking ahead to when we might migrate from H2 to PostgreSQL/MySQL, and I wasn’t sure whether it’s still considered best practice to keep data.sql-based initialization for E2E tests, or if it’s more common to set up a dedicated test database for those scenarios.
After doing a bit of research on the options for E2E testing, here’s my current thinking.

For now, I believe it’s appropriate to continue with data.sql and H2 in-memory because:
- Test speed – H2 in-memory runs tests much faster.
- Isolation – Every test starts with a clean, predictable state.
- SQL compatibility – The SQL we’re using is fairly standard (basic INSERTs), so compatibility shouldn’t be an immediate concern.

Later on, if we start relying on database-specific features, I’d consider switching to Testcontainers with PostgreSQL/MySQL to ensure full compatibility with production.

src/main/resources/data.sql
VittorioDeMarzi: Just to clarify, my question wasn’t about whether data.sql can work with another database. I understand it’s compatible as long as the SQL syntax matches.

What I meant was: once we migrate production to something like PostgreSQL, is it still considered good practice to keep a separate test environment using H2 + data.sql, so we can run tests with a predictable, pre-populated state?

Here’s how I currently see the options:
H2 + data.sql (current approach)
- Pro: Fast, simple, isolated
- Cons: Potential SQL dialect differences

Testcontainers with PostgreSQL/MySQL
- Pro: Identical to production
- Cons: Slower, requires Docker

src/main/kotlin/ecommerce/model/Member.kt
VittorioDeMarzi: The main reason I placed the conversion logic (toDto() / toEntity()) inside the entity classes was simplicity and proximity, it felt convenient to have the transformation logic right next to the fields it operates on, especially while the project is still small.

However, looking at it more critically now, I can see several issues with this approach:
- tight coupling between entity and DTO.
- Less flexible if DTOs change later.
- Entity classes can get bigger and harder to read.
- Testing complexity: Harder to unit test mapping logic in isolation
- Scalability: As the project grows, entities could become bloated with multiple DTO conversions

Better approach I'm considering:
Moving to extension functions in separate mapper files - this gives me the best of both worlds:
- Clean, idiomatic Kotlin syntax (member.toDto())
- Proper separation of concerns
- Easy to test and maintain
- Entities stay focused on their core responsibility

Something like:
```
// In MemberMapper.kt
fun Member.toDto(): MemberDto = MemberDto(id, email, role)
```

This way I keep the convenient syntax but gain better architecture. For larger projects I might even consider MapStruct for performance and type safety.

What's your take on extension functions vs dedicated mapper classes for this use case?

src/main/kotlin/ecommerce/dto/ProductRequest.kt
VittorioDeMarzi: What I was solving:
- Reducing boilerplate code for DTO ↔ Entity conversions
- Keeping conversion logic close to the data structure
- Making the service layer cleaner by avoiding manual mapping

What felt clean:
- Extension functions felt more Kotlin-idiomatic
- DTOs "knowing" how to convert themselves seemed intuitive
- One place to maintain the mapping logic

But now I see the architectural issue...
The hint about "Domain logic should not be affected by changes in controllers" is key!

My domain entity Product is now indirectly coupled to my controller layer through the DTO. If I change the DTO structure, I might need to change domain logic.

Better approach: move the logic to separate mapper classes!

src/main/kotlin/ecommerce/model/CartHistory.kt
VittorioDeMarzi: Before the migration to JPA, our initial approach was actually using JdbcTemplate. Curios to try something new we asked for hints and we were advised to look at @NamedNativeQuery with @SqlResultSetMapping. We got it working and kept this approach, but since the query doesn’t rely on any database-specific features, we always felt that this solution was:
- verbose in terms of annotations,
- less readable,
- and putting the query directly inside the CartHistory entity felt like it broke separation of responsibilities.

Here’s how I see the options now:
1. @**NamedNativeQuery**
    - Most verbose and least flexible.

2. @**Query** with Native SQL
    - More concise than annotations.
    - Lives in the repository, keeping data access logic in one place.
    - Possible errors in runtime while mapping
    - Manual mapping not type-safe
    
3. @**Query** with **JPQL** + Constructor Expression
     - Type-safe constructor-based mapping.
     - Database-agnostic.Concise and readable.

4. **JPA Projections** (Interface-based)
     - Flexible and automatic mapping by name.
     - Requires an extra interface.
     - Less control over mapping and possible performance limitations for complex queries.

5. **JdbcTemplate**
     - Maximum predictability and maintainability.
     - Very easy to debug.
     - Excellent performance.
     - More boilerplate, but full control.

For this specific case, I think JdbcTemplate is still one of the best options. But since in this case we hare a domain simple query, to explore something new and keep some of Spring’s “magic,” I’ll first try implementing it with Query with JPQL + Constructor Expression 😊

src/main/kotlin/ecommerce/dto/ProductPatchRequest.kt
VittorioDeMarzi: My take is that (like always 😄) it depends on the context and on what we want to protect.

If the main goal of the validations is to protect the API from invalid input, then validating only the request DTOs might be enough, since that’s the first entry point for the data.

But if the entity can also be created or modified from other sources, then adding validations at the entity level gives an extra layer of safety.

In this project, all data currently comes through the API and is validated with @Valid on DTOs, so having the same validations on the entity might feel redundant and could cause duplication.
Still, I can see the benefit of having some critical constraints (like @NotNull for required fields) directly on the entity, to make sure the domain model is always in a valid state, even if it’s not coming from the controller.

So my current opinion is:
- Keep API-specific rules (like max length for UI input) in DTOs.
- Keep domain integrity rules (like required fields) in entities.

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
VittorioDeMarzi: Got it! I’ll rename the package to follow the standard convention (e2e in lowercase).
I’m actually curious now about the Java Language Specification part, I’ll check it out to understand the reasoning behind the naming convention.

src/main/kotlin/ecommerce/model/Cart.kt
VittorioDeMarzi: Let me break down the JPA annotation details:

## Generation Type
JPA provides several generation strategies:
- **`IDENTITY`**
tells JPA to use the database's auto-increment feature to generate primary key values. When you insert a new entity, the database automatically assigns the next available ID value.
    - The ID is generated at INSERT time by the database.
    - Hibernate doesn’t know the ID until after the insert happens.
    - Great for MySQL and databases with AUTO_INCREMENT.
 
Downside: Less efficient for batch inserts, because each insert must be executed individually to get the generated ID.
- **`SEQUENCE`** - Uses database sequences
    - Hibernate asks the database for the next sequence value before the insert.
    - Works perfectly with PostgreSQL and Oracle.

 Advantage: Very efficient for batch inserts because Hibernate knows all IDs before inserting.

- **`TABLE`** - Stores the last used ID in a dedicated table and increments it manually (Slower and rarely used in production.)
- **`AUTO`** - Lets JPA choose the best strategy for your database
- **`UUID`** - alternative when you want to generate a universally unique identifier instead of a simple incremental Long
    - Globally unique – No risk of collision, even across different databases.
    - Database-independent – Doesn’t rely on sequences or identity, so it’s portable.
    - Security – IDs don’t reveal the number of records or insertion order (useful in public APIs).
    - Client-side generation – You can create the ID before saving the entity. 
    - Takes up 16 bytes (compared to 8 bytes for a Long).

## Why 0L vs null?

The choice between 0L and null (making it Long?) comes down to how you want to handle new entities:

**Using 0L (non-nullable Long):**
- Simpler type system - no null checks needed
- 0L acts as a sentinel value meaning "not yet persisted"
-  JPA ignores the default value anyway when generating IDs

**Using null (nullable Long?):**
- More semantically correct (null = "no ID yet")
- Requires null safety handling throughout your code
- Clearer intent that unpersisted entities have no ID

Do you  personally prefer one of the two approaches?

src/main/kotlin/ecommerce/model/Cart.kt
VittorioDeMarzi: Here’s a quick overview of all `CascadeType` options in JPA:

* **PERSIST** – Saves new child entities when the parent is saved (initial persist only).
* **MERGE** – Updates child entities when the parent is updated, also merges detached entities.
* **REMOVE** – Deletes children when the parent is deleted.
* **REFRESH** – Reloads children from the database when the parent is refreshed.
* **DETACH** – Detaches children from the persistence context when the parent is detached.
* **ALL** – Applies all of the above.

In our case:

* `PERSIST` for `Cart → Member` because members are created elsewhere and not updated via the cart.
* `ALL` for `Cart → CartItem` because cart items are fully owned by the cart and should be persisted, updated, or removed together with it.

src/main/kotlin/ecommerce/model/CartItem.kt
VittorioDeMarzi: 😄 I actually had the id at the top initially as well, mostly because I was used to that style from working with Java.
But during the sessions with Json and Justin, they suggested keeping it at the bottom so we don’t have to include it in the constructor.

I found this especially convenient for creating entities in tests, since we can omit the id entirely when building test objects. So for me, it’s more practical and less boilerplate.

src/main/kotlin/ecommerce/model/Cart.kt
VittorioDeMarzi: Initially, my plan was to keep the relationship **unidirectional**, similar to `Product` → `Option`.
```
@OneToMany
val options: List<Option>
```
That approach allowed me to centralize the validation logic inside the Product entity, for example, ensuring that each Option has a unique name:
```
init {
    require(options.isNotEmpty())
    require(options.distinctBy { it.name }.size == options.size)
}
```

However, in that case (correct me if I’m wrong), Spring Data JPA automatically creates a **join table** between `Product` and `Option` because there is no `mappedBy` side and no foreign key in the child entity.

To avoid the join table and have direct foreign keys, we eventually chose a **bidirectional** mapping.
By adding `@JoinColumn` on one side and `mappedBy` on the other, we gained full control over the foreign keys and avoided the unnecessary join table, while still keeping navigation possible in both directions.

src/main/kotlin/ecommerce/repository/ProductJpaRepository.kt
VittorioDeMarzi: 😜😎 Initially, we were also returning nullable types in all repositories, just like in MemberJpaRepository.
However, this was actually a suggestion from Jason, and I personally find it very convenient.

By handling the not-found logic inside the repository itself, we can keep the service layer clean from null checks and explicit exception throwing, making the code more concise and easier to read.

src/main/kotlin/ecommerce/service/CartService.kt
VittorioDeMarzi: Yes, I can replace it with coerceAtMost, in Kotlin it makes the code cleaner and more idiomatic.
And I’ll also remove the comment to keep the commit clean 🙂

src/main/kotlin/ecommerce/service/CartService.kt
VittorioDeMarzi: Yes, I agree, it’s better to return an immutable list here to avoid accidental modifications.
Since cartProducts is a MutableList (which maybe we can still handle it as List actually), subList will still give a mutable view, so I can wrap it with toList() to make it immutable:
```
val productsInPage = products.subList(start, end).toList()
```
This way, once we’ve created the list we need, it’s guaranteed not to be modified elsewhere.

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
VittorioDeMarzi: Good idea, since the Option has a unique test value, it makes sense to verify that it’s also persisted correctly.
I extend the test like:
```
assertThat(actual.options).hasSize(1)
assertThat(actual.options[0].name).isEqualTo("test")
assertThat(actual.options[0].quantity).isEqualTo(10)
```
This way, we’re not only checking that the Product was saved, but also that its associated Option was persisted with the expected values.

VittorioDeMarzi:
Thanks a lot for the thorough review, PK 🙌

I’ve carefully gone through all your questions, they were super helpful in understanding more deeply what’s going on behind certain annotations and relationships.

Here’s what I’ve changed:
- Extracted the responsibility of creating entities and DTOs into separate mapper classes.
- Moved the CartHistory queries back into the repository, now using `@Query` with JPQL + Constructor Expression.
- Removed the `CartItem` relation from `CartHistory` to allow removing a cart item from a cart.
   Now `CartHistory` holds `Member`, `Product`, and `quantity`.
   To delete a product, I now first remove it from the history, then from any carts, and only then delete the product itself.
- Moved `Pageable` creation from the controller into the service layer.
- Added Spring’s `@Transactional` annotation to `ProductService` so that product creation, updates, and deletions are fully transactional.
- Synchronized the bidirectional relationship between `Product` and `Option` when creating a product and when adding an option.
- Added a logger to the `GlobalExceptionHandler` and included the cause when throwing `ProductDeleteException`, `ProductUpdateException`, and `ProductCreationException`.

I share even some answers to your questions in Alex PullRequest.

## 1. Persistence
**Let's remove commented out logic so it doesn't get merged into our final branch.
That said, this commented-out line is actually pretty interesting.
Even though we're not explicitly saving the `product`, the new option still ends up persisted in the DB... 🤯
Why is that happening?
(I would like you to explain to me how JPA's persistence context works 😉 )**

This happens because when we fetch `product` with `getByIdOrThrow()`, JPA returns a **managed entity** inside the current persistence context.

While we're still in the same transactional context, any changes made to that entity are automatically tracked (*dirty checking*).

When the transaction completes (or `flush()` is called), JPA generates the necessary `INSERT`/`UPDATE` statements to persist the changes, even without explicitly calling `save()`.

In our case, the `@OneToMany(cascade = CascadeType.ALL)` mapping means that adding a new `Option` to the `product` will also persist it automatically.

## 2. Exception cause

**As we work with more and more libraries, there will be plenty of situations where we need to chain exceptions, just like you did here. Nice job! 👍
Here's the fun part:
When chaining exceptions, I personally prefer including the original cause when throwing a custom exception, like this:**

```
} catch (e: Exception) {
    throw ProductUpdateException("Failed to update product, id: $id", e)
}
```

**This does mean we need to add another constructor to our custom exception class, but it gives us much more powerful options when it comes to exception handling.
Why do you think I prefer this pattern? 🙂
Would love to hear your thoughts!**

I see your point, and I agree, having a constructor that accepts the original cause makes debugging much easier.

I even managed to consolidate all status exception handlers into single methods adding buildErrorResponse()

Also, I integrated a logger (instead of just printing) to keep consistent error tracking and make it easier to investigate issues in production.

## 3. Transactional
**`@Transactional` annotation! 😮**
**Since you've added this annotation here, I would like you to tell me**
- **what this does**
- **what options it has**
- **and the difference between `org.springframework.transaction.annotation.Transactional` and `jakarta.transaction.Transactional`**

**This is a regular question in tech interviews, so we might as well go over it here :)**

@Transactional marks a method or class to run inside a database transaction.
Spring opens the transaction at the start, commits on success, and rolls back on runtime exceptions by default.
Key options include readOnly, rollbackFor, propagation, isolation, and timeout.

```
org.springframework.transaction.annotation.Transactional
```
is Spring’s extended version with more features, while
```
jakarta.transaction.Transactional
```

is the standard JTA annotation with fewer options.
In Spring projects, we usually prefer Spring’s version for better integration and flexibility.

### IN TEST
In **tests**, `@Transactional` works the same way as in application logic-Spring opens a transaction at the start of the test and rolls it back at the end.
The **key difference** is that in tests the rollback is automatic, so any DB changes made in the test don’t persist.
This keeps your test database clean without needing manual cleanup, and allows tests to be run repeatedly without side effects.

## 4. Bidirectional sync

**Let's go back to the [[topic about pros and cons of bidirectional mapping](https://github.com/herotechcourse/spring-ecommerce-order/pull/2#discussion_r2249151076)](https://github.com/herotechcourse/spring-ecommerce-order/pull/2#discussion_r2249151076).**
**I was going to keep this review for next time, but since we're having 2x review session (and a lot of higher level discussion here), might as well talk about this too**
**I think there's a missing logic here, which is CRUTIAL to bidirectional mapping.**
**This is actually a perfect example of bidirectional mapping's downside!**
**I'm not saying you did a bad job, I'm just excited that we're experiencing it rather than just talking about it :)**
**What should get added here? 🤔**

You're right — with bidirectional mapping, it's crucial to keep both sides of the relationship in sync.
In my current `addOption()` method, I’m only updating the `Product` side (`options.add(option)`), but I’m not setting the back-reference in `Option`.

I should update it to something like:
```
fun addOption(option: Option) {
    require(options.none { it.name == option.name })
    options.add(option)
    option.product = this
}
```
This way, `product.options` and `option.product` stay consistent in memory, avoiding issues where JPA might not persist the relationship correctly.
It also highlights one of the main downsides of bidirectional mapping, you always have to manually maintain both sides to prevent inconsistent state.


All tests are now passing ✅ - ready for another review 🚀
Let me know if you need any clarification.. happy to discuss further! 🙌

src/main/resources/data.sql
pkeugine: I agree with your point, and H2 based tests are still very popular and effective in real-life applications.
Like you said, when we have any database-specific features, we can think about bringing in Testcontainers.
It's great that you've thought of that option as well 👍

src/main/resources/data.sql
pkeugine: Like what we've discussed in the previous comment, H2 based tests are still very effective even when we use other database vendors. This is because test time and simplicity are two very important factors of tests. When you get to implement tests with Testcontainers, you'll notice a significant increase in test time, which will affect not even local development, but also teamwork such as building and maintaining the project in build pipeline.

src/main/kotlin/ecommerce/model/Member.kt
pkeugine: I wouldn't mind DTO changing itself to an entity by itself, but I would definitely avoid `toDto` in entity, since that would make entities depend on DTOs which are close to controller (view) layer.

You've given me a great explanation about mappers, and I agree with you. I personally would save it for later since like you said, the project isn't so big just yet. However, I wouldn't mind using mappers since I see the upsides. 👍

src/main/kotlin/ecommerce/dto/ProductPatchRequest.kt
pkeugine: Nice convention :)
I completely agree with you, and I personally like to think simply : `more validations, the better`, since I think objects should never trust what happened in the outside world (so entity should validate what's needed even if DTO validated it)

However, as the project gets bigger and bigger, it's very hard to keep up with all the validations, and if there are SO MANY, the important business logic becomes hard to find with all the thick validation logics hanging around. That's when I started to think `let's keep it so that objects trust each other, and think all the validations were made previously` 😂 

Conclusion : It really depends on the context and situation (like always 😄 ) with team convention being the most important factor at the end.

src/main/kotlin/ecommerce/model/Cart.kt
pkeugine: Wow, you've done a great job summarizing all these info! 👍 
I personally prefer keeping the value null when there isn't a value set for that property, since I like to keep things consistent across all fields. I say this because there are times where fields don't have a good 'default value' or 'value that is great to express nothing'. However, there are developers who prefer `0L` since they try their best to keep things non-null. So they each have their pros and cons.

src/main/kotlin/ecommerce/model/CartItem.kt
pkeugine: Sounds great to me! 👍

pkeugine (approved):
Hello Vito!
That was very quick! There were additional reviews in another PR, but you've managed to check all of them and do a great refactoring. 👍 
The answers to my questions also look very solid, with nothing too much to fix.

I've added my personal thoughts in your additional questions, and checked the refactored project.
It seems like we're ready to move onto the next step! 👍 
If you have any questions or have something to discuss more, feel free to DM me any time like yesterday :)

Happy Hacking! 🔥 🔥 🔥

PR #3, opened 2025-08-01
[spring-ecommerce-order - Step-1] Submit SebasNadu's mission

SebasNadu:
## spring-ecommerce-order Step-1
Implement Step 1 Features: JPA Integration, Pagination, Wishlist, and Product Options

## Description
This pull request implements the features outlined in Step 1 of the project roadmap:

### Step 1.1: JPA Integration
- Replaced `JdbcTemplate` with Spring Data JPA for database operations.
- Updated entities and services to use JPA.
- Added logging for JPA operations to improve traceability.

### Step 1.2: Pagination and Wishlist
- Added pagination support for products.
- Implemented wishlist functionality.
- Added pagination support for wishlist items.

### Step 1.3: Product Options
- Created the `Option` entity with the following constraints:
  - Name limited to 50 characters.
  - Quantity must be between 1 and 99,999,999.
  - Prevented duplicate option names within the same product.
  - Created subtract method in Option.
- Added admin functionality to create new options for products.
- Created an endpoint in the `AdminController` for adding new options to products.

All features have been tested and verified.

## Questions
1.	I’m not entirely sure when to use @Transactional in service classes. Is my current usage appropriate, or are there cases where it’s unnecessary or even discouraged?
2.	For my @OneToMany relationships, I always use CascadeType.ALL. Do you think that’s a good practice, or would it be better to use more specific cascade types depending on the context?


README.md
```diff
@@ -1 +1,23 @@
...
+- [x] The entity should contain a name up to 50 characters
+- [x] The Option quantity must be at least 1 and less than 100_000_000
+- [x] Duplicated option names are not allowed within the same product to prevent confusion during purchase.
+
+- [x] Add Admin functionality to create new option for product
+- [x] Create an endpoint in the admin controller for adding new options to products
+
+
+### Future work (maybe?)
+- [ ] Subtract the user option choice from stock for a product option
```
Hyeon9mak: nice habit 💯 
and future work... of course, it would be great to do it 😄 

But, since there are more essential things to take away from the HeroTechCourse missions, it might be better to take your time and think it through slowly. 😉

build.gradle.kts
```diff
@@ -19,16 +20,26 @@ repositories {
     mavenCentral()
 }
 
+val assureVersion = "5.5.5"
+val h2Version = "2.2.224"
+val jwtVersion = "0.12.6"
 dependencies {
-    implementation("org.springframework.boot:spring-boot-starter-jdbc")
+//    implementation("org.springframework.boot:spring-boot-starter-jdbc")
```
Hyeon9mak: remove comment of no more use library.
it makes confusion to ur fellow developers.

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
```diff
@@ -0,0 +1,219 @@
...
+        log.warn("DuplicateKeyException: $errorMessage", e)
+        val body =
+            mapOf(
+                "status" to HttpStatus.CONFLICT.value(),
+                "error" to "Duplicate Key Error",
+                "message" to errorMessage,
+                "timestamp" to Instant.now(),
+            )
+        return ResponseEntity.status(HttpStatus.CONFLICT).body(body)
+    }
```
Hyeon9mak: nice logging, but we need to think about.
is it sure WARN logging level of internal server error?

How about categorizing log levels like this?
- ERROR level for issues that developers need to fix immediately.
- WARN level for those that can be monitored and addressed later.

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
```diff
@@ -0,0 +1,26 @@
...
+class WebMvcConfiguration(
+    private val checkLogin: CheckLoginInterceptor,
+    private val checkRole: CheckRoleInterceptor,
+    private val loginMemberArgumentResolver: LoginMemberArgumentResolver,
+) : WebMvcConfigurer {
+    override fun addInterceptors(registry: InterceptorRegistry) {
+        registry.addInterceptor(checkLogin).addPathPatterns("/api/**", "/products", "/admin/**")
+            .excludePathPatterns("/api/members/**")
+        registry.addInterceptor(checkRole).addPathPatterns("/admin/**", "/api/products/**")
+    }
```
Hyeon9mak, line 21: Great job separating the interceptors by feature—this makes it much easier to define clear rules per route. 👍 
What changes do you think might arise if we also separated interceptors by role (e.g., Admin, Customer)?
In what situations do you think the pros and cons of each approach would become more noticeable? 😄

src/main/kotlin/ecommerce/controller/AdminController.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    @GetMapping("/active-members")
+    fun getActiveMembers(): List<ActiveMemberDTO> = adminService.findMembersWithRecentCartActivity()
+
+    @PostMapping("/newOption")
+    fun createOption(
+        @RequestBody optionDTO: OptionDTO,
+    ): ResponseEntity<Unit> {
+        adminService.createOption(optionDTO)
+        return ResponseEntity.ok().build()
```
Hyeon9mak: `newOption` camelCase alone! 👀 
Consistency leads to Readability,
Readability leads to Maintainability.

src/main/kotlin/ecommerce/controller/CartItemController.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @DeleteMapping
+    fun deleteCartItem(
+        @RequestBody cartItemRequestDTO: CartItemRequestDTO,
+        @LoginMember member: MemberDTO,
+    ): ResponseEntity<Unit> {
+        cartItemService.delete(cartItemRequestDTO, member.id!!)
+        return ResponseEntity.noContent().build()
+    }
```
Hyeon9mak: A successfully logged-in user always has an ID.
And since Kotlin encourages null-safety by design.

I think it might be better to create a separate `LoginMember` representation that includes only the necessary information, rather than reusing a more general-purpose `MemberDTO`.

What do you think about this approach? 😄

src/main/kotlin/ecommerce/controller/MemberController.kt
```diff
@@ -0,0 +1,51 @@
...
+) {
+    @PostMapping("/register")
+    fun register(
+        @Valid @RequestBody memberDTO: MemberDTO,
+    ): ResponseEntity<TokenResponseDTO> {
+        val memberDTO: MemberDTO = memberService.save(memberDTO)
+        val tokenResponse = authService.createToken(memberDTO)
+        return ResponseEntity.ok().body(tokenResponse)
+    }
+
```
Hyeon9mak: Like `LoginMember`, register api's `MemberDTO` also have unnecessary property `id`. 💭 
Maybe, it means "`MemberDTO` have too many responsibility."!

src/main/kotlin/ecommerce/controller/MemberController.kt
```diff
@@ -0,0 +1,51 @@
...
+    @PostMapping("/login")
+    fun login(
+        @Valid @RequestBody tokenRequestDTO: TokenRequestDTO,
+    ): ResponseEntity<TokenResponseDTO> {
+        if (authService.checkInvalidLogin(tokenRequestDTO)) throw ForbiddenException("Invalid email or password.")
+        val memberDTO = memberService.findByEmail(tokenRequestDTO.email)
+        val tokenResponse = authService.createToken(memberDTO)
+        return ResponseEntity.ok().body(tokenResponse)
+    }
```
Hyeon9mak: "Invalid member information" looks like a business logic, 
"find member by email" looks like a business logic.
There're not HTTP protocol's logics.

And Spring Controller layer's role is representation of HTTP protocol!
So, we can just ask to `authService`. 😄 

```kotlin
val tokenResponse = authService.login(...)
```

src/main/kotlin/ecommerce/controller/ProductViewController.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.stereotype.Controller
+import org.springframework.ui.Model
+import org.springframework.validation.BindingResult
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestMapping
+
+@Controller
+@RequestMapping
+class ProductViewController(private val productService: ProductService) {
```
Hyeon9mak: Is `@RequestMapping` actually working in this controller? 👀

src/main/kotlin/ecommerce/infrastructure/JwtTokenProvider.kt
```diff
@@ -0,0 +1,65 @@
...
+class JwtTokenProvider(
+    @Value("\${security.jwt.token.secret-key}")
+    secret: String,
+    @Value("\${security.jwt.token.expire-length}")
+    private val validityInMs: Long,
+) {
+    private val secretKey: SecretKey =
+        Keys.hmacShaKeyFor(secret.toByteArray(StandardCharsets.UTF_8))
+
+    fun createToken(
```
Hyeon9mak, line 23: (Just ask! feel free 😄 )
What's different between infrastructure package and config package ?

src/main/kotlin/ecommerce/model/MemberDTO.kt
```diff
@@ -0,0 +1,21 @@
...
+    var id: Long? = null,
+    @field:NotBlank(message = MEMBER_NAME_REQUIRED)
+    var name: String,
+    @field:NotBlank(message = EMAIL_BLANK)
+    @field:Email(message = EMAIL_INVALID)
+    var email: String,
+    @field:NotBlank(message = PASSWORD_BLANK)
+    var password: String,
+    var role: Member.Role = Member.Role.CUSTOMER,
+)
```
Hyeon9mak: What's different between DTO and entity?

src/main/kotlin/ecommerce/services/CartItemServiceImpl.kt
```diff
@@ -0,0 +1,106 @@
...
+            CartItemResponseDTO(
+                id = cartItem.id!!,
+                memberId = cartItem.member.id!!,
+                product = cartItem.product.toDTO(),
+                quantity = cartItem.quantity,
+                addedAt = cartItem.addedAt,
+            )
+        }
+    }
+
```
Hyeon9mak: The `@Transactional` annotation functionally guarantees the ACID properties described in https://en.wikipedia.org/wiki/ACID.

Ultimately, an annotation is just a simple marker, and how it is interpreted depends on each implementation.
In the case of JPA (Hibernate), this annotation acts as a mediator that decides whether or not to actually send queries to the database.

e.g.
When wrapping Create/Update/Delete operations with `@Transactional`, JPA optimizes actions in JVM memory and only sends the final query to the database at the very last moment, providing some additional benefits.

When wrapping read operations with `@Transactional(readOnly = true)`, even if the code attempts to send Create/Update/Delete queries internally, those queries are not actually sent to the database. It guarantees a read-only operation as the name implies.

There are many other features that JPA provides through the `@Transactional` annotation.
I believe it’s best for Sebas to study and discover them one by one. 😄 

Remember: the important thing is that the `@Transactional` annotation is basically a declaration that
“This is a unit of work that guarantees ACID properties through transaction!”

src/main/resources/templates/fragments/modalForms.html
```diff
@@ -0,0 +1,69 @@
...
+                    </div>
+                </div>
+                <div class="modal-footer">
+                    <button type="submit" class="btn btn-primary">Update Product</button>
+                    <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Cancel</button>
+                </div>
+            </form>
+        </div>
+    </div>
+</div>
```
Hyeon9mak: https://medium.com/@hyeon9mak/about-no-newline-at-end-of-file-warning-on-github-b42d14dccc93 😉

Hyeon9mak (changes requested):
Hi Sebas! Ur spring-ecommerce-order reviewer here!
U did a great job implementing this mission. 👍 

I’m sure you have many questions about JPA,
but for now, I left comments mainly focused on clarifying the concepts of Controller, DTO, and Entity.

Let's start by getting a good grasp of these concepts first, then we can take a deeper dive into JPA step by step. 😉 

Feel free to ask anytime if you have any questions!

README.md
SebasNadu: Yes, exactly. We left this TODO because we assumed that in the next step, we might need to use the subtract method, possibly to reduce stock when a product with a specific option is added to the cart or ordered. But I agree, we focused on completing all the requirements and understanding the core topics before adding extra features, which is why we didn’t implement it yet. That said, I understand it’s better to remove incomplete or speculative features before merging. I’ll take care of it 👍

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
SebasNadu: You’re right, that’s a good point. Logging DataAccessException with WARN might not be the best choice, especially since it results in an internal server error.

I’ll update the log levels accordingly, as you suggested. I have to admit I didn’t think too much about the logging strategy, my mind was mostly focused on the projects features. So thanks a lot for pointing this out!

Appreciate the suggestion!

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
SebasNadu: If we separate interceptors by role, such as AdminInterceptor or CustomerInterceptor, we could encapsulate both login and role checks specific to each user type. That might simplify route protection for individual roles and centralize role-specific logic in one place. This could be useful if roles have very distinct request flows.

However, the current separation by feature makes the interceptors more reusable and avoids duplication. For example, CheckLogin applies to any authenticated user, and CheckRole works across different roles depending on the endpoint. Given our current context, where all users, including admins, need to be authenticated, and only some routes require admin privileges, splitting by feature is more practical and flexible for me.

Since we don’t currently have routes that are exclusive to customers, I think the most maintainable and reusable solution is to keep interceptors separated by feature rather than by role.

Thanks for your question, I’d love to hear your perspective on when role-based separation becomes more suitable.

src/main/kotlin/ecommerce/controller/CartItemController.kt
SebasNadu: Yes, this is a solid idea. Defining a LoginMember type with only the authenticated user’s ID makes the contract clearer and safer.

I’ll refactor the argument resolver and controller to use a LoginMember representation instead. Thanks for pointing it out.

src/main/kotlin/ecommerce/controller/MemberController.kt
SebasNadu: You’re right. For me, the real issue is security and data inconsistency. Right now, it’s possible to send a POST request with an existing user ID, and instead of creating a new user, it updates the existing one. That’s risky and causes unexpected behavior. Thanks for pointing it out!

src/main/kotlin/ecommerce/controller/MemberController.kt
SebasNadu: Yes, again you’re completly right. Mixing business logic like checking credentials and fetching the member directly in the controller breaks the responsibility of the HTTP layer. Delegating the entire login flow to a service keeps the controller focused on HTTP concerns only. I’ll refactor it, thanks for the insight! 😄

src/main/kotlin/ecommerce/controller/ProductViewController.kt
SebasNadu: Yes, it’s meant to define the base path for all methods in the controller. I probably added it earlier for that reason but forgot to remove it later. Right now, it has no effect, it’s the same as not having it at all. It might even add unnecessary processing during controller creation. Thanks for pointing it out, I’ll remove it!

src/main/kotlin/ecommerce/infrastructure/JwtTokenProvider.kt
SebasNadu: Good question! 😄🤔
I see infrastructure as the place for concrete implementations that interact with external systems, things like JWT, databases, email senders, file storage, external APIs, etc. It’s where the “adapters” live.

On the other hand, config is meant for setting up Spring beans, loading properties, and wiring dependencies. In my case, I also use it to seed the H2 database. It’s purely configuration, no business logic, no external integrations.

So JwtTokenProvider fits well under infrastructure because it’s not just wiring, it implements token generation, which is external to the domain, and not part of the business logic itself.

src/main/kotlin/ecommerce/model/MemberDTO.kt
SebasNadu: Entity = Domain + Persistence
DTO = Data structure for input/output

Entities represent the actual domain model stored in the database. They are persistent and typically include JPA/Hibernate annotations, relationships, and business identity (@Id, @Version, etc.).

DTOs (Data Transfer Objects) are used to transfer data between layers (controller ↔ service) or across system boundaries (e.g., API). They usually don’t contain persistence annotations and often represent only a subset of the entity’s fields, or a combination of several entities.

In my case, the service layer is responsible for converting between DTOs and Entities. The repository layer should work only with Entities.

src/main/kotlin/ecommerce/services/CartItemServiceImpl.kt
SebasNadu: Thanks for the detailed explanation! That makes it much clearer.

I’ll definitely look deeper into how JPA handles transactions under the hood, especially the behavior around delayed writes and session flushing. This kind of control could be super helpful once you understand what’s actually happening behind the scenes, right now for me it's a blackbox. Thanks for the push! 😄

src/main/resources/templates/fragments/modalForms.html
SebasNadu: Thank you for sharing, and congrats on your article, Hyeon-Gu!
It was great to understand the reason behind the importance of the newline at the end of a file🤔.
The article was clear, concise, and easy to follow, really well written! 👏

SebasNadu:
Hi Hyeon-Gu👋,
Thanks for your review and thoughtful questions, I’ve implemented all your suggestions. They really helped me reflect on my understanding and continue growing.

✨ Features
- Implemented member registration and login using MemberRegisterDTO.

🛠 Refactors
- Replaced usage of MemberDTO with MemberLoginDTO in Cart and WishItem services for clarity and encapsulation.
- Removed unused DTO imports and cleaned up related test code.
- Removed redundant @RequestMapping annotations to improve controller clarity.

🐞 Fixes
- Changed log level from WARN to ERROR in exception handling, ensuring better visibility in logs.
- Added missing newlines at the end of several .html files for POSIX compliance and cleaner diffs.

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
Hyeon9mak: As you probably already know, the ultimate standard for good code is how easy it is to maintain later on.
But how can we predict whether something will be easy to maintain or not?
The key lies in deeply understanding the nature of our business.
That’s because the parts of the system that need to change frequently differ depending on the characteristics of the business itself.

This is why engineers shouldn’t just study technology — they also need to have a good grasp of the business.

src/main/kotlin/ecommerce/infrastructure/JwtTokenProvider.kt
Hyeon9mak: There can be many different opinions, and there’s no such thing as a perfect architecture.
What truly matters is having a clear intention behind your code — enough to explain it with confidence.
You did a great job answering that!

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
```diff
@@ -0,0 +1,219 @@
...
+        val body =
+            mapOf(
+                "status" to HttpStatus.NOT_FOUND.value(),
+                "error" to "Operation failed",
+                "message" to errorMessage,
+                "timestamp" to Instant.now(),
+            )
+        return ResponseEntity.status(HttpStatus.NOT_FOUND).body(body)
+    }
+
```
Hyeon9mak, line 43: What would happen if an exception occurs that we didn't explicitly handle?
Hint: inheritance exception.

src/main/kotlin/ecommerce/entities/CartItem.kt
```diff
@@ -0,0 +1,34 @@
...
+    @JoinColumn(name = "member_id", nullable = false)
+    val member: Member,
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "product_id", nullable = false)
+    val product: Product,
+    @Column(name = "quantity", nullable = false)
+    var quantity: Int,
+    @Column(name = "added_at", nullable = false)
+    val addedAt: LocalDateTime,
+)
```
Hyeon9mak: What was your reasoning for choosing `FetchType.LAZY`?

For example, when displaying CartItem on a shopping page,
it would be natural for the Product information to be shown together.
In that case, EAGER might actually be a better fit than LAZY. 😉 

How about considering fetch strategies from a business perspective like this? 😄

src/main/kotlin/ecommerce/entities/CartItem.kt
```diff
@@ -0,0 +1,34 @@
...
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import jakarta.persistence.UniqueConstraint
+import java.time.LocalDateTime
+
+@Entity
+@Table(
+    name = "cart_item",
+    uniqueConstraints = [UniqueConstraint(columnNames = ["member_id", "product_id"])],
+)
```
Hyeon9mak, line 19: notice unique column. 💯 👍 
it makes helpful to fellow developers.

src/main/kotlin/ecommerce/entities/Member.kt
```diff
@@ -0,0 +1,35 @@
...
+import jakarta.persistence.Table
+
+@Entity
+@Table(name = "member")
+class Member(
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long? = null,
+    @Column(name = "name", nullable = false)
+    val name: String,
```
Hyeon9mak: In SpringDataJpa default settings, null or zero(0L) value represent "it's not yet granted ID".
We can dodge from null!

check the `org.springframework.data.repository.core.support.AbstractEntityInformation.class`
image

So we can generate entity class like this

```kotlin
class Member(
    @Column(name = "email")
    val email: String,
) {
    @Id
    val id: Long = 0L
}
```

src/main/kotlin/ecommerce/entities/Member.kt
```diff
@@ -0,0 +1,35 @@
...
+    @Column(name = "email", nullable = false)
+    val email: String,
+    @Column(name = "password", nullable = false)
+    val password: String,
+    @Enumerated(EnumType.STRING)
+    @Column(name = "role", nullable = false)
+    val role: Role = Role.CUSTOMER,
```
Hyeon9mak: I think,emil, name, password, role need length limits in Table Create DDL.
So we can declare length limit in `@Column` annoation.
What do u think? 😄

src/main/kotlin/ecommerce/entities/Member.kt
```diff
@@ -0,0 +1,35 @@
...
+    @OneToMany(mappedBy = "member", cascade = [CascadeType.ALL], orphanRemoval = true)
+    val cartItems: Set<CartItem> = emptySet(),
+    @OneToMany(mappedBy = "member", cascade = [CascadeType.ALL], orphanRemoval = true)
+    val wishItems: Set<WishItem> = emptySet(),
```
Hyeon9mak: What's different between CascadeType.REMOVE and orphanRemoval ?

src/main/kotlin/ecommerce/entities/Product.kt
```diff
@@ -0,0 +1,80 @@
...
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.OneToMany
+import jakarta.persistence.Table
+
+@Entity
+@Table(name = "product")
+class Product(
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long? = null,
```
Hyeon9mak: Question about entity definition.

- id = 1, name = 'Sebas's MacBook'
- id = 1, name = 'Hyeon9mak's MacBook'

Should these be considered the same Product or different ones?

- id = 1, name = 'Sebas's MacBook'
- id = 30, name = 'Sebas's MacBook'

And how about these two? Same Product or different?

Your answer will influence how you define equality for the Entity! 😄

src/main/kotlin/ecommerce/services/AdminService.kt
```diff
@@ -0,0 +1,13 @@
...
+import ecommerce.model.OptionDTO
+import ecommerce.model.TopProductDTO
+
+interface AdminService {
+    fun findTopProductsAddedInList30Days(): List<TopProductDTO>
+
+    fun findMembersWithRecentCartActivity(): List<ActiveMemberDTO>
+
+    fun createOption(optionDTO: OptionDTO)
+}
```
Hyeon9mak: In Spring, using an interface is not just about ensuring polymorphism — it's also a clear declaration that you want to separate concerns between layers.

Right now, the controller appears to depend on the service interface, but in reality, it references the entire services package. This means it is effectively aware of the concrete implementation, which violates the intended abstraction.

To truly decouple using interfaces, the interface should reside not where it is used, but where it is called from — in this case, within the controller package.
(Just like how we can freely use List or Set without knowing their exact implementation details — we simply rely on the interface they expose.)

However, placing something like Service inside the controller package feels awkward and unnatural. That’s actually a sign that the interface is poorly named. By calling it a *Service, you’re leaking the implementation detail through the name itself.

A better approach is to use a more general and intention-revealing name that doesn't expose how it's implemented. For example, naming it something like AuthUseCase would express what it does, not how it does it.

What do u think about this? 😉

Hyeon9mak (changes requested):
Hi Sebas, sorry for late 🙏 
Did a great job reflect feedback comments! 💯 👍 

this time, I've left some comments about JPA!
Let's learn more deeper-!

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
SebasNadu: You’re absolutely right, maintainability depends on how well our design reflects the business reality.
If the business evolves and introduces distinct behaviors for roles (customers vs. admins handling different flows), then separating interceptors by role could align better with this flow.

So yes, anticipating change means staying close to the business story.
Thank you for sharing your insights!! they really made me reflect on how we define the “right” architecture, and how, as always, there are no fixed rules, it all depends on the context.

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
SebasNadu: If an exception occurs that isn’t explicitly handled by any @ExceptionHandler in my @RestControllerAdvice, Spring falls back to its default exception handling mechanism, which typically produces a generic 500 Internal Server Error response.

Thank you for your suggestion!, so to avoid this and maintain a consistent API style, I can implement a fallback exception handler that catches all unhandled exceptions. This allows me to provide clients with clearer and more structured error information. I’ll do this by adding a handler that catches the superclass Exception.

src/main/kotlin/ecommerce/entities/CartItem.kt
SebasNadu: You’re absolutely right, choosing FetchType.LAZY was a default-safe decision to me to avoid unnecessary data loading and circular reference risks, especially in serialization.

But I agree, from the real use case perspective, it makes more sense to load Product eagerly, since it’s almost always needed. That would reduce extra queries caused by LAZY loading in that context for example.

So a more tailored approach might be:
- Product: FetchType.EAGER
- Member: keep as LAZY, since the cart usually doesn’t need full member data

I’ll also review the other relationships, since I have to admit I didn’t give them much thought at this stage. Thanks for the great reminder to align technical decisions with actual use cases 😄👍👍

src/main/kotlin/ecommerce/entities/Member.kt
SebasNadu: That’s very interesting, thanks for the insight!

Using 0L as the default ID value allows us to avoid nullable ID fields while still relying on JPA’s identity generation strategy. I’ll update the entity definitions accordingly. It’ll require some refactoring, but it’s definitely worth it in the long run. Also, thanks for pointing to the source class, helped a lot to clarify the behavior.

src/main/kotlin/ecommerce/entities/Member.kt
SebasNadu: mmmm You’re right, now that JPA define the tables, adding length limits in the @Column annotation looks like the correct way to do this🤔. Thank you for your insight, I forgot that!

src/main/kotlin/ecommerce/entities/Member.kt
SebasNadu: CascadeType.REMOVE remove from parent to children. If I delete a Member also the cartItems should be deleted in the same operation.
With orphanRemoval true, it's removing children when they are removed from the collection, even if the parent is not deleted. if I call member.cartItems.remove(item), that item will be deleted from DB immediately.
So cascadeType.REMOVE is about parent removal, but orphanRemoval is about child removal due to the disassociation from parent.

src/main/kotlin/ecommerce/entities/Product.kt
SebasNadu: You’re right, how we define equality depends on how we interpret identity within the domain.

In this case, since we’re using @Id with GenerationType.IDENTITY, I consider the id field to be the sole source of identity. So:
- Same id, different names == same product
- Same name, different ids ==  different products

I’ll override equals and hashCode based only on id, treating transient entities (id == null) as unequal to any persisted entity. Now I can see that this becomes especially important when working with @OneToMany relationships, where entities are stored in sets or maps.

Thanks for raising this, it made me re-evaluate the domain more carefully!

src/main/kotlin/ecommerce/services/AdminService.kt
SebasNadu: mmm that’s a really interesting perspective. I hadn’t considered interface placement and naming in that way before. The idea of positioning interfaces closer to where they’re used, and naming them based on intention rather than implementation, sounds interesting. I’ll try to apply that approach by moving the use case interfaces to the caller layer and renaming them to better reflect what they actually do. Thanks for the insight, it definitely made me rethink how I structure responsibilities with a cleaner architectural mindset.

SebasNadu:
Hi Hyeon-Gu 👋,

I applied your suggestions, thank you again for the thoughtful review and all the questions, they really help me grow as a developer.

I’m curious about your thoughts on how I refactored the service layer. Also, do you have any recommendations regarding the structure of DTOs and entities, or preferences on how to organize this project in general?

Looking forward to your feedback!

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
```diff
@@ -0,0 +1,235 @@
...
+        val body =
+            mapOf(
+                "status" to HttpStatus.INTERNAL_SERVER_ERROR.value(),
+                "error" to "Internal Server Error",
+                "message" to "An unexpected error occurred",
+                "timestamp" to Instant.now(),
+            )
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(body)
+    }
+
```
Hyeon9mak, line 171: nice 👍

src/main/kotlin/ecommerce/entities/WishItem.kt
```diff
@@ -0,0 +1,44 @@
...
+class WishItem(
+    @Column(name = "added_at", nullable = false)
+    val addedAt: LocalDateTime,
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "member_id", nullable = false)
+    val member: Member,
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "product_id", nullable = false)
+    val product: Product,
+    @Id
```
Hyeon9mak, line 29: > That would reduce extra queries caused by LAZY loading in that context for example.

More accurately, EAGER loading doesn’t mean fewer queries. It doesn’t perform JOINs by default — it just decides whether related entities are fetched immediately or lazily when accessed.

The best way is to log the actual queries and observe the generated JPQL yourself. 😉 
Since this is something Sebas can enjoy firsthand, I’ll skip the detailed explanation. 😄

src/main/kotlin/ecommerce/entities/Member.kt
```diff
@@ -0,0 +1,44 @@
...
+    val email: String,
+    @Column(name = "password", nullable = false, length = 255)
+    val password: String,
+    @Enumerated(EnumType.STRING)
+    @Column(name = "role", nullable = false, length = 20)
+    val role: Role = Role.CUSTOMER,
+    @OneToMany(mappedBy = "member", cascade = [CascadeType.ALL], orphanRemoval = true)
+    val cartItems: MutableSet<CartItem> = mutableSetOf(),
+    @OneToMany(mappedBy = "member", cascade = [CascadeType.ALL], orphanRemoval = true)
+    val wishItems: MutableSet<WishItem> = mutableSetOf(),
```
Hyeon9mak, line 29: > So cascadeType.REMOVE is about parent removal, but orphanRemoval is about child removal due to the disassociation from parent.

The most important thing is to understand how it works and use it accordingly. 💯 👍

src/main/kotlin/ecommerce/entities/Member.kt
```diff
@@ -0,0 +1,44 @@
...
+    enum class Role { CUSTOMER, ADMIN }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is Member) return false
+        if (id == 0L || other.id == 0L) return false
+        return id == other.id
+    }
+
+    override fun hashCode(): Int = id.hashCode()
```
Hyeon9mak, line 43: Now we can finally call it a true entity 😄

src/main/kotlin/ecommerce/controller/admin/AdminController.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/admin")
+@CheckAdminOnly
+class AdminController(
+    private val createOptionUseCase: CreateOptionUseCase,
+    private val findTopProductsUseCase: FindTopProductsUseCase,
+    private val findMembersWithRecentCartActivityUseCase: FindMembersWithRecentCartActivityUseCase,
```
Hyeon9mak, line 23: Practicing this approach—placing interfaces close to where they are used and having the implementations depend on them—will be a great help later on in understanding one of Spring’s key concepts: IoC(Inversion of Control).

Surprisingly, many developers struggle to fully grasp this concept. I believe it’s mainly due to a lack of hands-on practice. Taking the time to go through it calmly and practicing step by step can really help. 👍 💪

src/main/kotlin/ecommerce/services/auth/AuthServiceImpl.kt
```diff
@@ -0,0 +1,39 @@
...
+    override fun login(tokenRequestDTO: TokenRequestDTO): TokenResponseDTO {
+        if (checkInvalidLogin(tokenRequestDTO)) throw ForbiddenException("Invalid email or password.")
+        val memberDTO = memberService.findByEmail(tokenRequestDTO.email)
+        return createToken(memberDTO)
+    }
+
+    override fun checkInvalidLogin(tokenRequestDTO: TokenRequestDTO): Boolean {
+        val memberDTO = memberService.findByEmail(tokenRequestDTO.email)
+        return tokenRequestDTO.email != memberDTO.email || tokenRequestDTO.password != memberDTO.password
+    }
```
Hyeon9mak, line 26: we learned by past missions, we can ask validation to domain model.
`memberDTO.validatePassword(password = password)`

src/main/kotlin/ecommerce/model/MemberDTO.kt
```diff
@@ -0,0 +1,28 @@
...
+import ecommerce.util.ValidationMessages.MEMBER_NAME_REQUIRED
+import ecommerce.util.ValidationMessages.MEMBER_NAME_SIZE
+import ecommerce.util.ValidationMessages.PASSWORD_BLANK
+import ecommerce.util.ValidationMessages.PASSWORD_SIZE
+import jakarta.validation.constraints.Email
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Size
+
+data class MemberDTO(
+    @field:NotBlank(message = MEMBER_NAME_REQUIRED)
```
Hyeon9mak, line 16: Now, a follow-up question!
What is the difference between a model and a DTO?

Currently, the package is named model, but all the class names include the DTO suffix.
I'm curious to hear your thoughts on this.

src/main/kotlin/ecommerce/services/auth/AuthServiceImpl.kt
```diff
@@ -0,0 +1,39 @@
...
+import ecommerce.infrastructure.JwtTokenProvider
+import ecommerce.model.MemberDTO
+import ecommerce.model.TokenRequestDTO
+import ecommerce.model.TokenResponseDTO
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+
+@Service
+class AuthServiceImpl(private val jwtTokenProvider: JwtTokenProvider, private val memberService: CrudMemberUseCase) :
+    AuthUseCase {
```
Hyeon9mak, line 15: The UseCase interface was probably created for the controller layer.
For this reason, using that interface when one service depends on another service can lead to overly complicated dependency structures.

Also, having one service layer depend on another can introduce significant maintenance challenges.
For example, if MemberService changes due to a policy update, then AuthService would also need to be updated accordingly.

In such cases, it might actually be simpler for AuthService to depend directly on MemberRepository.

As engineers, we often have a natural instinct to reduce code duplication — but the cost of slightly duplicated code is thousands of times cheaper than the cost of a developer struggling to understand overly complex code.

Hyeon9mak (approved):
Hi Sebas, U applying the feedback very quickly! 👍 
I've also left comments addressing the questions you asked.

It’s clear that you already have a almost understanding of JPA and that you reflect feedback well.
You also seem interested in architectural separation between layers,
so I included some comments on slightly more abstract concepts that might be a bit challenging. 😄 

Please take your time to think them through and feel free to refactor based on your own reasoning.
When you open your next PR, just make sure to explain why you chose to refactor things that way.

If you have any questions, don’t hesitate to ask.
See you in the next step!

src/main/kotlin/ecommerce/model/MemberDTO.kt
SebasNadu: You’re right to point out the naming inconsistency. Initially, all the classes in the model package were purely DTOs, used only for transferring data, mainly in the controller layer. That could cause confusion, so I decided to make the distinction explicit:
- I’ve created a new dto package for classes strictly used as dumb data carriers (mostly in the controller layer).
- If a class includes logic (like validateEmail() or validatePassword()) and is reused in the service layer, then I consider it more than a DTO. In that case, I place it in the model package, treating it as an application-level model—something between a domain entity and a pure DTO.
- The entities package continues to contain the JPA entities, which reflect the database structure.

So the structure now is:
- dto/      → controller input/output only
- model/    → reusable logic, validation, or used in service layer
- entities/ → JPA entities, tied to persistence

Let me know if you’d suggest a better separation. I’m open to refining this structure!

src/main/kotlin/ecommerce/services/auth/AuthServiceImpl.kt
SebasNadu: You're absolutely right, I initially reused `CrudMemberUseCase` to avoid duplication, but that created unnecessary coupling between services.

I'll refactor `AuthServiceImpl` to depend directly on `MemberRepository`, and now also it is extracted shared logic (like password validation) to the `Member` domain model. That way, responsibilities stay focused and services stay decoupled.

Thanks for the insight, it helped clarify the direction!

PR #4, opened 2025-08-02
[spring-ecommerce-order - Step-1] Submit AnnLvu's mission

AnnLvu:
## 🔄 Spring E-commerce-order - Step 1 PR Overview
Hi PK 👋,

This PR represents **Step 1** of our Spring E-commerce platform improvements. The main focus of this step was to **migrate from JDBC to Spring Data JPA**, implement **pagination**, and **add support for product options**.

#### 🔁  JDBC → Spring Data JPA
- Replaced all repository logic with Spring Data JPA interfaces
- Removed direct SQL/JdbcTemplate usage
- Introduced proper entity mapping using annotations:
  - `@Entity`, `@Id`, `@OneToMany`, `@ManyToOne`

#### 🧩 Entity Relationships
- Defined entities: `Product`, `Option`, `Cart`, `CartProduct`, `User` (Member/Admin)
- Maintained bidirectional relationships where necessary
- Added conversion logic between DTOs and entities (`toDTO()`, `toEntity()`)

#### 📄 Pagination Support
- Implemented pagination for product listings (both for guest and admin views)
- Controlled via `page` and `perPage` query parameters
- Default values configured and reusable

#### ⚙️ Product Options
- Each product now supports a list of options 
- Options include `name`, `price`, `quantity`, `imageUrl`
- Full CRUD support for options 
- Validations:
  - Unique option name per product
  - Quantity ≥ 0, valid price, etc.
- `PATCH` support to partially update option fields

#### 🔄 Question
In several places (for example: `Product → Option`, `Cart → CartProduct`), I'm using:

```kotlin
@OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true)

```
This allows managing child entities directly from the parent, including creation and removal.

**Questions:**

- Is it appropriate to use `CascadeType.ALL` in these contexts, or would it be better to specify only the required cascade types (PERSIST, MERGE, etc.) more explicitly?

- I'm also interested in whether I've correctly understood and applied the JPA relationships themselves (e.g., mappedBy, @JoinColumn, orphanRemoval, etc.). Are there any aspects of the mapping you think could be improved?

### ❗️Important:
To test the project, you need to create a .env file at the same level as build.gradle.kts.
The file should contain the following parameters:

```kotlin
JWT_SECRET=
JWT_TIME=

```
And put your own secret values there

Thanks a lot for your time!


src/main/resources/application.properties
```diff
@@ -1 +1,13 @@
...
+spring.datasource.username=${DB_USER_NAME}
+spring.datasource.password=${DB_PASSWORD}
+spring.h2.console.enabled=true
+spring.sql.init.mode=always
+security.jwt.token.secret-key=${JWT_SECRET}
+security.jwt.token.expire-length=${JWT_TIME}
+spring.jpa.hibernate.ddl-auto=create
+spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.format_sql=true
+#spring.jpa.open-in-view=false
```
pkeugine: Although these aren’t application code files, some are missing trailing newlines.
Let’s add them and make sure our IDE is set to include them automatically. 👍

This helps avoid unnecessary diffs in git.
Otherwise, teammates will keep seeing these lines as "changed," which clutters the git history.

pkeugine: I think it would be better to remove the property instead of commenting it out if we're not going to use it.

Although it's interesting that you've tried this out, what's this `open-in-view` setting? 🙂

src/main/kotlin/ecommerce/dto/products/OptionDTO.kt
```diff
@@ -0,0 +1,29 @@
+package ecommerce.dto.products
+
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Positive
+import jakarta.validation.constraints.Size
+import org.hibernate.validator.constraints.Range
+
+class OptionDTO(
```
pkeugine: Class names should follow [the general camel case format, even when acronyms are included](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).
In this case, `OptionDTO` could be changed to `OptionDto`. Why don't we change it? :)

Same review applies to function names as well!

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,57 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+
+@Entity
+class Cart(
+    @OneToMany(mappedBy = "cart", cascade = [CascadeType.ALL], orphanRemoval = true)
+    val items: MutableList<CartProduct> = mutableListOf(),
```
pkeugine, line 14: I see you've used bidirectional mapping between `Cart` and `CartProduct`!
I'm curious about your experience in maintaining this structure.
What were the pros and cons?

I ask this because a lot of developers have different thoughts on how to manage this relationship.

(I'm not pointing out that this is wrong, so please share your thoughts without pressure 🙂 )

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,57 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
```
pkeugine, line 17: So many new kinds of annotations were added as we migrated to JPA right? 🥲 
You did well adding in `@Id` and `@GeneratedValue` here.

I'm curious, what does `GenerationType.IDENTITY` do, and what are the other options?
Also, what made you choose `0L` to be the default value for id? I guess there are two things we could choose from : `null` and `0L`. If we chose null, id would be `Long?`.

src/main/kotlin/ecommerce/model/CartProduct.kt
```diff
@@ -0,0 +1,31 @@
...
+    var cart: Cart,
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "option_id", nullable = false)
+    var option: Option,
+    @Column(name = "quantity", nullable = false)
+    var quantity: Int = 1,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+)
```
pkeugine, line 31:
```suggestion
class CartProduct(
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    val id: Long = 0L,

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn(name = "cart_id", nullable = false)
    var cart: Cart,

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn(name = "option_id", nullable = false)
    var option: Option,

    @Column(name = "quantity", nullable = false)
    var quantity: Int = 1,
)
```

I personally prefer keeping `id` field on the very top, since that (to me) is the most important value when it comes to entity.

Also, when there's multiple annotations added, I like to keep some line breaks in between fields, as it helps increase the readability. (this is also just my opinion 😂 )

src/main/kotlin/ecommerce/repository/CartStatisticRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+    WHERE cs.action = :action
+      AND cs.createdAt >= :since
+    GROUP BY cs.option.name
+    ORDER BY COUNT(cs) DESC, MAX(cs.createdAt) DESC
+    """,
+    )
+    fun findTopProducts(
+        @Param("action") action: CartAction,
+        @Param("since") since: LocalDateTime,
+        pageable: org.springframework.data.domain.Pageable,
```
pkeugine:
```suggestion
        pageable: Pageable,
```

Why don't we simplify type names such as `org.springframework.data.domain.Pageable` to just `Pageable`? It would make reading the logic a lot easier!

Also, it would be interesting to find a way to reduce unnecessary parts in SQL, such as `new ecommerce.dto.cartStatistics.TopAddedProductsDTO`. There are multiple ways to do it, so feel free to look up some options and tell me what you think about them.

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,57 @@
...
+        val existing = findProduct(option)
+        return if (existing != null) {
+            existing.quantity += quantity
+            existing
+        } else {
+            val newItem = CartProduct(this, option, quantity)
+            items.add(newItem)
+            newItem
+        }
+    }
```
pkeugine, line 32: Great job not forgetting about this logic as these models are linked as bidirectional. 👍 

Since there are domain logics, it would be great to have unit tests for these logics.
Same applies to all the `require` logics in other models (domains)!

src/main/kotlin/ecommerce/utils/extensions/OptionUtil.kt
```diff
@@ -0,0 +1,24 @@
...
+
+fun Option.toDTO(): OptionResponseDTO {
+    return OptionResponseDTO(
+        id,
+        name,
+        price,
+        quantity,
+        imageUrl,
+    )
+}
```
pkeugine: I personally think the current `util` package is having too much things in it : `custom exceptions`, `dto - entity mappers`, `custom annotation that is related to Spring MVC`, and `JwtProvider`.

This would be a great time for us to think about `why do we separate packages`, and `when would we need one, while the other might not be needed`.

src/main/kotlin/ecommerce/dto/cartProduct/CartProductResponse.kt
```diff
@@ -0,0 +1,5 @@
+package ecommerce.dto.cartProduct
+
+class CartProductResponse(
```
pkeugine, line 3: I'm curious about how you manage class names, especially DTO names!
Is there your own convention?

I'm asking this question, because to me, it's not so straight forward on how to manage the names if I were to manage this project together. For example, there are names like :  
* `CartProductDTO`
* `CartProductResponse`
* `ProductResponseDTO`
* (if we look at other domain) `AuthTokenPayload`

I personally think class, method, and even variable names are extremely important, since when named well, they deliver their intention and purpose very well. If named without detail, we'll be the ones to look up their intentions every time we have to use them.

Why don't we refactor the names  so that they have some kind of convention that we can easily understand? :)
(If you had the convention but I didn't catch it, I'm sorry 🥲 )

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,57 @@
...
+import jakarta.persistence.CascadeType
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+
+@Entity
+class Cart(
+    @OneToMany(mappedBy = "cart", cascade = [CascadeType.ALL], orphanRemoval = true)
```
pkeugine, line 13: **Q. Is it appropriate to use CascadeType.ALL in these contexts, or would it be better to specify only the required cascade types (PERSIST, MERGE, etc.) more explicitly?**

Nice question!
I would stick with `ALL` for now, since `Cart` owns the full lifecycle of `CartProduct`, and `Product` owns `Option`'s.

It would be more fun to ask 
* `so then, when would an entity not own child's full lifecycle?`
*  or `even if it did, when would we not use CascadeType.ALL?`. 

I would like to know your take on this. 😄

src/main/kotlin/ecommerce/controller/admin/AdminProductController.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    @PutMapping("/{productId}/options/{optionId}")
+    fun updateOption(
+        @PathVariable("productId") productId: Long,
+        @PathVariable("optionId") optionId: Long,
+        @RequestBody @Valid optionDTO: OptionDTO,
+    ): ResponseEntity<MessageResponse> {
+        adminProductService.updateOption(productId, optionId, optionDTO)
+        return ResponseEntity.ok(MessageResponse("Option updated"))
+    }
```
pkeugine: As two entities get involved, we can't just say `id` since we don't know which `id` we're referring to. That made us explicitly declare `productId` and `optionId` in our path.

Although this might seem not too important, I like to keep a convention on this. Rather declare all the id names explicitly even when only one entity is getting used, or allow us to use `id` for the entity that's closely related to the api.

How do you think about this? It's fine to not apply this review, but I'm curious about your thoughts.

src/main/kotlin/ecommerce/service/AdminProductService.kt
```diff
@@ -0,0 +1,182 @@
...
+
+@Service
+@Transactional
+class AdminProductService(
+    private val productRepository: ProductRepository,
+    private val optionRepository: OptionRepository,
+) {
+    fun getAllProducts(
+        page: Int = DEFAULT_PAGE,
+        perPage: Int = PER_PAGE,
```
pkeugine, line 28: The service currently depends on controller-related information, which I generally try to avoid.

In a typical layered architecture, the flow of logic goes **top-down** like `controller → service → persistence`. To keep the architecture flexible and maintainable, it's best if each layer doesn't depend on the layer above it. That way, changes in the controller won’t ripple down and affect the service layer unnecessarily.

What do you think about this approach? 😄
No pressure at all, feel free to share your thoughts!

src/main/kotlin/ecommerce/service/AdminProductService.kt
```diff
@@ -0,0 +1,182 @@
...
+import ecommerce.utils.extensions.getPaginatedDTOs
+import ecommerce.utils.extensions.toEntity
+import ecommerce.utils.extensions.toProductDTO
+import jakarta.transaction.Transactional
+import org.springframework.data.domain.Page
+import org.springframework.stereotype.Service
+import java.net.URI
+
+@Service
+@Transactional
```
pkeugine, line 21: It's interesting that you've used `@Transactional` annotation!
What does this do, and what are the options we can use for this annotation?
(This question is pretty important, as it's a regular in tech interviews)

update : after finding out which `@Transactional` annotation you've added, I want you to tell me the difference between different `@Transactional` annotations!

src/main/kotlin/ecommerce/service/AdminProductService.kt
```diff
@@ -0,0 +1,182 @@
...
+
+        if (product.options.find { it.name == optionDTO.name } != null) {
+            throw DuplicateProductNameException("Duplicate option not accepted")
+        }
+
+        val newOption = optionRepository.save(optionDTO.toEntity())
+        product.options.add(newOption)
+
+        return URI.create("/products/${product.id}/options/${newOption.id}")
+    }
```
pkeugine: As we introduced `Option` to our project, I think this `AdminProductService` became thick, meaning it's holding onto multiple domains and is ready to be separated. Controller feels the same too, but I feel like service is more important since it's closer to our application logic.

How do you feel about maintaining `Option`s in their own service?

src/main/kotlin/ecommerce/service/AdminProductService.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    fun getProductOptions(productId: Long): ProductResponseDTO {
+        val product = getValidProduct(productId)
+        return product.toProductDTO()
+    }
+
+    fun createOption(
+        productId: Long,
+        optionDTO: OptionDTO,
+    ): URI {
```
pkeugine: I see service is returning `URI` as a result for adding a new entity.
What if we added `OutputView` (remember this guy? 😄 ) and added support for consoles?
Let's say we don't provide HTTP related information to console application. Then, we would have to make a new function for creating entities, or update this logic whereas if this function returned something that could be used in any controller, we would just make `OutputView` and get it over with.

I think this kind of thoughts add up and allow us to make a maintainable, flexible-to-expansion projects. How do you think about this? 😄

src/main/kotlin/ecommerce/service/AdminProductService.kt
```diff
@@ -0,0 +1,182 @@
...
+import ecommerce.model.Option
+import ecommerce.model.Product
+import ecommerce.repository.OptionRepository
+import ecommerce.repository.ProductRepository
+import ecommerce.utils.exception.DuplicateProductNameException
+import ecommerce.utils.exception.EntityNotFoundException
+import ecommerce.utils.extensions.getPaginatedDTOs
+import ecommerce.utils.extensions.toEntity
+import ecommerce.utils.extensions.toProductDTO
+import jakarta.transaction.Transactional
```
pkeugine: Are you sure you want to import this `Transactional` instead of `org.springframework.transaction.annotation.Transactional`? 😄

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,30 @@
...
+
+@Entity
+class Product(
+    @Column(name = "name", nullable = false, unique = true)
+    var name: String,
+    @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true)
+    @JoinColumn(name = "product_id")
+    var options: MutableList<Option>,
+    @Column(name = "created_at", nullable = false)
+    var createdAt: LocalDateTime = LocalDateTime.now(),
```
pkeugine, line 21: **Q. I'm also interested in whether I've correctly understood and applied the JPA relationships themselves (e.g., mappedBy, joincolumn, orphanRemoval, etc.). Are there any aspects of the mapping you think could be improved?**

I think this is a topic that can't be answered with just one comment.
I'm not saying there's a lot of wrong implementations here, what I'm trying to say is that this mission is made to let you think and experience it yourself :)

You've done a great job and I'll be sure to alert you if things feel like they need directions 👍

pkeugine (changes requested):
Hello Ann!
Learning Spring Boot is hard already, but learning JPA as well?
That requires intense focus, and you did well on it 👍

I left some reviews, questions, and answers to your questions.
Feel free to request for another review once you're done with all of them.
Some answers might feel vague, but I tried my best to let you experience this mission to it's fullest, rather than me giving you all the answers (they may not even be an answer anyways since many developers have their own point of views).

Feel free to ask back any questions via DM or comments!
Happy Hacking! 🚀 🚀 🚀

src/main/resources/application.properties
AnnLvu: Hi PK! :relaxed:
I’ve enabled the "Ensure line feed at file end on Save" option in IntelliJ IDEA.
Now, the IDE automatically adds a newline at the end of files when saving, which helps avoid unnecessary diffs in Git. Also, I  removed unnecessary property.

src/main/kotlin/ecommerce/dto/products/OptionDTO.kt
AnnLvu: Thanks for pointing that out. I’ve gone ahead and renamed all `*DTO` to `*Dto` across the project to follow the proper CamelCase convention.
Sorry about that - I initially kept DTO in all caps because I considered it an abbreviation and thought it would be more readable that way.

src/main/resources/application.properties
AnnLvu: Yeah, we actually talked about this with Jason - he said it’s usually better to keep `spring.jpa.open-in-view=true` in production, since it helps avoid `LazyInitializationException` when entities are accessed outside a transaction.
We were just experimenting locally with turning it off and totally forgot to remove the commented line afterwards.
Cleaned it up now - thanks for catching it! 😄

src/main/kotlin/ecommerce/model/Cart.kt
AnnLvu: Initially, we kept the cart structure unidirectional - `CartProduct` just had a `cart_id`, and `Cart` itself didn’t hold a collection of items.
It worked, but it meant that any cart operations required explicitly working with the `CartProduct `repository. 
Later, we switched to the bidirectional mapping you see now:
```
// That's  Cart
@OneToMany(mappedBy = "cart", cascade = [CascadeType.ALL], orphanRemoval = true)
val items: MutableList<CartProduct> = mutableListOf()

// That's CartProduct
@ManyToOne(fetch = FetchType.LAZY)
@JoinColumn(name = "cart_id", nullable = false)
var cart: Cart
```
We changed this because:

- It allows us to manage the cart directly through `cart.items`, like in `addProduct()` and `decrementProduct()` in our `Cart` class:

```
fun addProduct(
        option: Option,
        quantity: Int = 1,
    ): CartProduct {
        val existing = findProduct(option)
        return if (existing != null) {
            existing.quantity += quantity
            existing
        } else {
            val newItem = CartProduct(this, option, quantity)
            items.add(newItem)
            newItem
        }
    }
```
- `orphanRemoval = true` automatically removes `CartProduct` rows when we clear the list,
so `CartService.clearCart()` can simply do:
```
 fun clear() {
        items.clear()
    }
```
**Trade‑off:**

We need to keep both sides in sync and be mindful of lazy loading,
but for a frequently updated cart this approach keeps the service layer clean and avoids manual repository calls.

src/main/kotlin/ecommerce/model/Cart.kt
AnnLvu: I used `@GeneratedValue(strategy = GenerationType.IDENTITY`) because Justin showed it in the lecture, so I decided to stick with it. It basically means the database auto‑generates the id for each new row.
I also checked the other strategies:

- `AUTO` - JPA picks depending on the DB

- `SEQUENCE` - Uses a database sequence (common in Postgres/Oracle)

- `TABLE` - Uses a separate table (rarely used now)

I kept the default `id = 0L` so it’s non‑nullable, and JPA will just replace 0 with the real id on insert.
If I used `Long?` with `null`, I’d have to handle nulls everywhere.

src/main/kotlin/ecommerce/model/CartProduct.kt
AnnLvu: I actually keep the id field at the end for a practical reason:
in Kotlin, if the first constructor parameter has a default value (0L), you still need to use named arguments for all the following parameters when creating the object.
For example, if `id` were first, creating an `Option` in tests would look like this:
```
Option(
    id = 0L,
    name = "test",
    price = 10.2,
    quantity = 52,
    imageUrl = "http://localhost:8080/image/upload/product1.png",
)
```
By keeping `id` at the end, I can just rely on its default value and create test much more cleanly:
```
Option(
    "test",
    10.2,
    52,
    "http://localhost:8080/image/upload/product1.png",
)
```
It also feels more natural for me to read the business fields first, and the technical id last. 😄

But I get your point about readability - I’ve now added some line breaks between fields with multiple annotations to make it easier to scan visually.
Thanks for the tip!

AnnLvu: Justin actually showed us this Kotlin trick in the lecture - if you move the id to the end with a default value, creating test  becomes so much cleaner.

I used to always keep `id` first, but after that lecture I went through and moved all `id`s to the end.😅

src/main/kotlin/ecommerce/repository/CartStatisticRepository.kt
AnnLvu: Thanks for the tips!
I went ahead and:

- Imported types like Pageable so the method signatures are shorter and easier to read
- Switched JPQL to interface projections instead of those long new `ecommerce.dto.cartStatistics.TopAddedProductsDto(...)` lines

In the service I just map the interfaces to our existing DTOs, so everything outside the repo stays the same.

It definitely feels cleaner now - thanks for pointing it out!

src/main/kotlin/ecommerce/model/Cart.kt
AnnLvu: I added unit tests for the domain logic:

- Cart (addProduct, decrementProduct, clear and exception cases)

- Option (all require validations for name, price, quantity, and URL)

- Product (checks for at least one option and uniqueness of options)

This really helped me double‑check the business rules.
Thanks for the reminder!

src/main/kotlin/ecommerce/utils/extensions/OptionUtil.kt
AnnLvu: You were right - utils really turned into a “dumping ground”.
I went ahead and split it up:
exception - all custom exceptions

- mapper -  entity mappers & repository extensions

- annotation - LoginMember

- security - JwtProvider

Now the structure feels way cleaner and easier to navigate.
Thanks for nudging me to tidy it up! 🙌

src/main/kotlin/ecommerce/dto/cartProduct/CartProductResponse.kt
AnnLvu: You were right - our DTO names were a bit all over the place.
I went through and made them consistent:

- Requests now end with RequestDto
- Responses now end with ResponseDto
- Patch DTOs keep PatchDto
Thanks for pointing it out - it feels much cleaner now! 🙌

src/main/kotlin/ecommerce/controller/admin/AdminProductController.kt
AnnLvu: Thanks for your answer!
I think `CascadeType.ALL` is good here because Cart really owns its` CartProducts` -
if cart deleted or cleared, items must be deleted too. Same with Product and Option.

About your question - when entity doesn’t own full lifecycle:
maybe when children can live alone or belong to other parents too.
In this case, using only `PERSIST` and `MERGE` is better to avoid deleting children by accident.

To be honest, I never used other cascade types before, only `ALL`,
and I’m not sure when exactly better to use different ones.

src/main/kotlin/ecommerce/service/AdminProductService.kt
AnnLvu: I created a new PaginationConstants object inside the config package to hold `DEFAULT_PAGE` and `PER_PAGE`.
Now both the controller and service use these constants, so the service doesn’t depend on the controller anymore.

src/main/kotlin/ecommerce/service/AdminProductService.kt
AnnLvu: I changed from jakarta.transaction.Transactional to Spring’s @Transactional because:

Spring’s annotation is designed specifically for Spring apps and works better with its tools.

It lets us control transactions more carefully - for example, how they start, how they behave if something goes wrong, and how nested transactions work.

jakarta.transaction.Transactional is simpler and part of a Java standard, but it doesn’t have as many features.

src/main/kotlin/ecommerce/service/AdminProductService.kt
AnnLvu: I changed from jakarta.transaction.Transactional to Spring’s @transactional.
It was mistake :)

src/main/kotlin/ecommerce/service/AdminProductService.kt
AnnLvu: @Transactional is an annotation that helps manage database operations.
It means: “All the code inside this method should run as one single operation.”
If something goes wrong, all changes are undone to keep the data safe and consistent.

This is important to make sure the database stays correct and nothing breaks.

src/main/kotlin/ecommerce/service/AdminProductService.kt
AnnLvu: I created a separate `AdminOptionService` to handle all Option-related operations. This way, each service has a clear single responsibility, and the code is easier to maintain and extend.

src/main/kotlin/ecommerce/service/AdminProductService.kt
AnnLvu: I created a new DTO called `CreateOptionResult` to return meaningful data from the service instead of just URI.
Also updated the service and controller to use this new DTO.

src/main/kotlin/ecommerce/model/CartProduct.kt
AnnLvu: I use KtlinFormat to keep the code neat, and it removes empty lines. I need to use it to pass the Check validation. So, I can’t leave empty lines for separation if I want the Check to pass.🤣

src/main/kotlin/ecommerce/model/Cart.kt
pkeugine: Perfect explanation! 💯

src/main/kotlin/ecommerce/model/Cart.kt
pkeugine: I personally prefer keeping the value null when there isn't a value set for that property, since I like to keep things consistent across all fields. I say this because there are times where fields don't have a good 'default value' or 'value that is great to express nothing'. However like your thought, there are developers who prefer `0L` since they try their best to keep things non-null. So they each have their pros and cons. Great explanation!

src/main/kotlin/ecommerce/model/CartProduct.kt
pkeugine: Great! Since this is a suggestion, I understand your point and keeping it as is looks fine 👍

src/main/kotlin/ecommerce/controller/admin/AdminProductController.kt
pkeugine: I think your answer is placed in a different conversation 😂 
But your answer seems right on track 👍 
When child can live by itself or belongs to other parents, that's when we'll be thinking about using other cascade strategy. You say you've never met those cases, but soon you'll meet them (the time will come).
Until then, keeping these things in mind would definitely help.

Nice explanation! 👍

pkeugine (approved):
Hello Ann!
You did a great job refactoring this project, and all the answers you've left to my questions look great 😮 👍 

I think we are ready to move onto the next step!
I added some reactions and my thoughts to your answers so check them out when you have the time. Honestly, the answers were very solid so there weren't much to add.

If you have any questions or thoughts, feel free to DM me anytime :)
We're almost at the end of this level!
Happy Hacking! 👍 👍 👍

PR #6, opened 2025-08-02
[spring-ecommerce-order Step-1] Submit Bojana's mission

bojana-petroska:
**Hi Suno!**

This PR refactors the existing data access layer from `JdbcTemplate` to `Spring Data JPA`, aligning the project with a more idiomatic Spring persistence strategy. 

#### Key changes introduced in this refactor include:

- Domain Model Refactoring (Entities):
   - `Product`, `Member`, `CartItem` have been refactored into `JPA @Entity Classes`.
   - A new `Cart Entity` has been introduced to explicitly represent the shopping cart concept.
   - JPA annotations (`@Entity, @Table, @Id, @GeneratedValue, @Column`) are used for database mapping.
   - Relationships (`@OneToOne, @OneToMany, @ManyToOne`) have been established between entities.
   - `Cascade Types` are managed through JPA annotations for related entity deletions.
- Repository Layer Refactoring:
   - All existing repositories (`CartItemRepository, ProductRepository, MemberRepository`) have been refactored to extend `JpaRepository`.
   - A new `CartRepository` has been introduced, extending `JpaRepository`.
   - Spring Data JPA's derived query methods are utilized for common lookups (e.g., `findByEmail, findByName`).
- Service Layer Adjustments:
   - Service constructors have been updated to inject the new `Spring Data JPA repository interfaces`.
   - Methods interacting with repositories (e.g., `create, update, delete, findById`) have been adjusted to use `JpaRepository's save(), deleteById(), and findById().orElseThrow()` patterns.
   - Validation logic (`require()`) remains in the service layer.
- Testing Infrastructure:
   - New repository test classes (e.g., `ProductRepositoryJpaTest, MemberRepositoryJpaTest, CartRepositoryJpaTest, CartItemRepositoryJpaTest, CartEventRepositoryJpaTest`) have been introduced.
   - These tests leverage `@DataJpaTest` for efficient and isolated testing of JPA repositories, benefiting from automatic transaction rollback.
- Configuration & Schema:
   - `spring-boot-starter-data-jpa` has been added to build.gradle.kts.
   - `application.properties` has been configured with `JPA-specific` settings (ddl-auto, show-sql).
   - The manual schema.sql content has been removed, as JPA now handles schema generation.
   - Visualization of our db diagram: [dbdiagram](https://dbdiagram.io/d/Schema-6888be9fcca18e685c413eeb)

I am looking forward to your input and thank you very much for your time!
Bojana


src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
```diff
@@ -36,6 +36,14 @@ class GlobalExceptionHandler {
         return ResponseEntity.internalServerError().build()
     }
 
+    @ExceptionHandler(IllegalArgumentException::class)
+    fun handlerIllegalArgumentException(e: Exception): ResponseEntity<Map<String, Any>> {
+        println("IllegalArgumentException occurred: " + e.message)
+        val error = mapOf("page" to e.message)
+        val errorBody = mapOf("errors" to error)
+        return ResponseEntity.badRequest().body(errorBody)
```
Wordbe: Instead of using a Map, create a predefined error object to return consistent error responses. 🙂

src/main/kotlin/ecommerce/controller/api/CartController.kt
```diff
@@ -25,10 +28,18 @@ class CartController(
...
+        @RequestParam(defaultValue = "10") pageSize: Int,
+        @RequestParam(defaultValue = "name") sortBy: String,
+    ): ResponseEntity<Page<CartItemResponse>> {
         val memberId = member.id
-        val cartItems = cartItemService.getCartItemsByMemberId(memberId)
-        return ResponseEntity.ok(cartItems)
+        val cartItemPage =
+            when (sortBy.isEmpty()) {
+                true -> cartItemService.getCartItemsByMemberId(memberId, pageNumber, pageSize)
+                false -> cartItemService.getCartItemsByMemberId(memberId, pageNumber, pageSize, sortBy)
```
Wordbe: `sortBy` isn’t being used in the `CartItemService`. Please check it! 🙂

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
```diff
@@ -1,33 +1,35 @@
...
-//    ): ResponseEntity<List<ActiveMemberInfo>> {
-//        member.id ?: throw InternalServerErrorException(MESSAGE_AUTH_FAILED)
-//        val members = cartItemRepository.getActiveMembersInLast7Days()
-//        return ResponseEntity.ok(members)
-//    }
+    @GetMapping("/top5-products")
+    fun getTop5Products(
+        @LoginMember member: Member,
+    ): ResponseEntity<List<TopProductStats>> {
+        member.id
```
Wordbe: This isn’t being used!

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
```diff
@@ -1,33 +1,35 @@
...
+        return ResponseEntity.ok(stats)
+    }
 
-    companion object {
-        const val MESSAGE_AUTH_FAILED = "auth failed"
+    @GetMapping("/active-members")
+    fun getActiveMembers(
+        @LoginMember member: Member,
+    ): ResponseEntity<List<ActiveMemberInfo>> {
+        member.id
```
Wordbe: Same here.

src/main/kotlin/ecommerce/controller/api/ProductController.kt
```diff
@@ -23,16 +27,26 @@ class ProductController(private val productService: ProductService) {
...
+    fun getProducts(
+        @RequestParam(defaultValue = "0") pageNumber: Int,
+        @RequestParam(defaultValue = "10") pageSize: Int,
+        @RequestParam(defaultValue = "name") sortBy: String,
+    ): ResponseEntity<Page<ProductResponse>> {
+        val productPage =
+            when (sortBy.isEmpty()) {
+                true -> productService.getPaginatedProducts(pageNumber, pageSize)
+                false -> productService.getPaginatedProducts(pageNumber, pageSize, sortBy)
+            }
```
Wordbe: What about letting the `Service` handle this part? Since the `Controller` is mainly responsible for receiving external input, and this logic varies based on that input, it feels more appropriate as business logic.

src/main/kotlin/ecommerce/dto/CartItemResponse.kt
```diff
@@ -0,0 +1,10 @@
+package ecommerce.dto
+
+import ecommerce.model.Product
+import java.time.LocalDateTime
+
+data class CartItemResponse(
```
Wordbe, line 6: Is there a specific reason you chose to use a `data class`?

src/main/kotlin/ecommerce/dto/OptionResponse.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+data class OptionResponse(
```
Wordbe, line 3: .

src/main/kotlin/ecommerce/dto/ProductResponse.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.dto
+
+data class ProductResponse(
```
Wordbe, line 3: .

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -30,4 +31,10 @@ class CartItem(
     fun changeQuantity(quantity: Int) {
         this.quantity += quantity
     }
+
+    companion object {
+        fun to(cartItem: CartItem): CartItemResponse {
+            return CartItemResponse(cartItem.product, cartItem.quantity, cartItem.createdAt)
+        }
+    }
```
Wordbe: Try to ensure that domain objects don’t depend on DTOs.
If you've established a layered architecture, keeping dependencies flowing in a single direction has clear advantages. 🙂

src/main/kotlin/ecommerce/model/Option.kt
```diff
@@ -0,0 +1,33 @@
...
+class Option(
+    @Column(nullable = false)
+    var name: String,
+    @Column(nullable = false)
+    var quantity: Int,
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    @ManyToOne(optional = false)
+    var product: Product? = null
```
Wordbe, line 22: Is there a reason this wasn’t placed in the primary constructor? 🙂

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -16,9 +18,19 @@ class Product(
     var price: Double,
     @Column(nullable = false)
     var imageUrl: String,
+    options: List<Option> = emptyList(),
     @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
     val id: Long = 0L,
 ) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST])
+    val options: MutableList<Option> = options.toMutableList()
```
Wordbe, line 26: Is there a reason this wasn’t placed in the primary constructor? (Same question as in `Option`)

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -4,22 +4,39 @@ import ecommerce.dto.ProductForm
...
+    private val productRepository: ProductRepository,
+    private val optionRepository: OptionRepository,
+) {
     fun insert(form: ProductForm): Product {
-        checkProductNameExists(form.name)
-        val product = ProductForm.toProduct(form)
+        nameExists(form.name)
+        val option = Option(name = "none", quantity = 1)
+        val product = ProductForm.toProduct(form, listOf(option))
+        option.product = product
```
Wordbe: I don't think this in unnecessary because when you make `Product`, it is setting itself to the option in the constructor (`init` block)

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -4,22 +4,39 @@ import ecommerce.dto.ProductForm
...
         val savedProduct = productRepository.save(product)
+        optionRepository.save(option)
```
Wordbe: You already added the `CascadeType.PERSIST` on the `options` in the `Product`. 🙂  Was this still necessary in the end?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -46,16 +63,17 @@ class ProductService(private val productRepository: ProductRepository) {
         productRepository.delete(product)
     }
 
-    private fun checkProductNameExists(
+    fun nameExists(
```
Wordbe: This seems to be a method that's only used within `ProductService`.

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -46,16 +63,17 @@ class ProductService(private val productRepository: ProductRepository) {
...
         name: String,
         originalName: String? = null,
-    ) {
+    ): Boolean {
```
Wordbe: Why does it return `Boolean` ? Nowhere uses it now. 🙂

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -31,8 +48,8 @@ class ProductService(private val productRepository: ProductRepository) {
             productRepository.findByIdOrNull(id)
                 ?: throw InternalServerErrorException("ProductService.update() - Product with ID $id not found")
         val originalName = originalProduct.name
-        checkProductNameExists(form.name, originalName)
-        val product = ProductForm.toEntity(form, id)
+        nameExists(form.name, originalName)
+        val product = productRepository.findById(id).get()
```
Wordbe: The get() function throws an exception if the value is absent.
How about explicitly handling the exception instead—for better readability and control? 😌

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -31,12 +43,9 @@ class Product(
...
-        fun toEntity(
-            product: Product,
-            id: Long,
-        ): Product {
-            return Product(product.name, product.price, product.imageUrl, id = id)
-        }
+    fun addOption(option: Option) {
+        require(options.size == options.map { it.name }.distinct().size)
+        this.options.add(option)
+        option.product = this
```
Wordbe: It’s not being used—may I ask why you created it? 🤔

src/main/resources/sql/product.sql
```diff
@@ -1,17 +1,10 @@
...
-
-INSERT INTO product (name, price, imageUrl) VALUES ('Iron Man', 1000, 'https://alexnsan.comics/imageurl/1');
-INSERT INTO product (name, price, imageUrl) VALUES ('X-men', 1000, 'https://alexnsan.comics/imageurl/2');
-INSERT INTO product (name, price, imageUrl) VALUES ('Superman', 1000, 'https://alexnsan.comics/imageurl/3');
-INSERT INTO product (name, price, imageUrl) VALUES ('Naruto', 1000, 'https://alexnsan.comics/imageurl/4');
-INSERT INTO product (name, price, imageUrl) VALUES ('Full Metal Alchemist', 1000, 'https://alexnsan.comics/imageurl/5');
-INSERT INTO product (name, price, imageUrl) VALUES ('Batman', 1000, 'https://alexnsan.comics/imageurl/6');
+TRUNCATE TABLE products;
+;
+INSERT INTO products (name, price, image_url) VALUES ('Iron Man', 1000, 'https://alexnsan.comics/imageurl/1');
```
Wordbe, line 4: There is no DDL for options. It would be good to manage it together. 🙂

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -4,22 +4,39 @@ import ecommerce.dto.ProductForm
...
-        val product = ProductForm.toProduct(form)
+        nameExists(form.name)
+        val option = Option(name = "none", quantity = 1)
+        val product = ProductForm.toProduct(form, listOf(option))
+        option.product = product
         val savedProduct = productRepository.save(product)
+        optionRepository.save(option)
         return productRepository.findByIdOrNull(savedProduct.id)
             ?: throw InternalServerErrorException("ProductService.insert() - Product with ID ${savedProduct.id} not found")
     }
```
Wordbe: If `savedProduct` has been returned, that means it’s already created, so I don’t think we need to look it up again. What do you think? 🤔

src/test/kotlin/ecommerce/service/ProductServiceTest.kt
```diff
@@ -5,14 +5,13 @@ import ecommerce.exception.ProductNameAlreadyExistsException
...
 import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.context.SpringBootTest
 import org.springframework.transaction.annotation.Transactional
 
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 @Transactional
 class ProductServiceTest(
```
Wordbe: Please also enhance the tests in `ProductTest`, not just `ProductService`. Try to keep the logic cohesive within this domain as much as possible. 🙂

Wordbe (changes requested):
Hi Bo!

Thank you for writing the code so thoroughly.
I've left comments in the review. Please take a look and apply any suggestions that are feasible.

Also, please refer to and consider the following points:
1. A `.` means the same comment applies to similar parts of the code.
2. The product list API works well, but it's not displaying correctly on the actual screen. It would be great if you could address this. 

Thank you! 😊

src/main/kotlin/ecommerce/model/CartItem.kt
bojana-petroska: Great Point, it makes sense : ) I changed it!

src/main/kotlin/ecommerce/service/ProductService.kt
bojana-petroska: Yes, correct. I updated the whole update because it made couple of redundant calls to the repository.

src/main/kotlin/ecommerce/dto/CartItemResponse.kt
bojana-petroska: We used data classes for the responses `CartItemResponse`, `OptionResponse`, `ProductResponse` because they are DTOs only for holding data, that are used for transferring of it between layers. They don't contain any business logic or any complex behavior. For these reasons we find them most suitable for the response objects.
They're easy to work with when we use the data to build our UI or present a response. Their clear structure makes it simple to access properties like `product.name` or `cartItem.quantity`.

src/main/kotlin/ecommerce/model/Option.kt
bojana-petroska: We wanted to avoid the circular dependencies. Instead we decided to use the init block. The Option is created first and after that passed to the constructor of the Product. The init block then uses this existing Option to link the Product back to it.
I hope it was the right approach! : D

src/main/kotlin/ecommerce/service/ProductService.kt
bojana-petroska: As we were working on this, this part particular was a bit blur for us and we were trying to understand the behavior. Probably we used save just to be sure : )
But, I just took another look and read more about it and believe it is actually redundant, because the `option` would be already saved from the `product repository`. Yes, it makes sense!

src/main/kotlin/ecommerce/model/Product.kt
bojana-petroska: We were initially not sure how to add the options to the product so we created this. But, we didn't remove it after initializing the `options` in the `init block` from the `product`. Again, we weren't exactly sure of the behavior, which i now start to understand : )

bojana-petroska:
Hi Suno!

I applied the changes and tried to answer the questions. I hope I did them correct : )

I was wondering about the display of the Products API, what exactly do you mean?

This is what I get as a response:

Screenshot 2025-08-05 at 16 17 11

src/main/kotlin/ecommerce/dto/CartItemResponse.kt
Wordbe: Even without using a `data class`, Kotlin provides getters by default for `val` properties!
A `data class` automatically generates `toString()`, `equals()`, `hashCode()`, and `copy()` methods—so if you don't need those, it's totally fine not to use it. 🙂 What do you think about that?

src/main/kotlin/ecommerce/model/Option.kt
Wordbe: OK, that makes sense. Thanks for the explanation! 🙂

src/main/kotlin/ecommerce/model/Product.kt
Wordbe: If it makes sense now, please remove it instead of just commenting it out! 🧹

Wordbe (changes requested):
Hello Bo!

Thank you for reflecting the review feedback so well.
When I mentioned that the product wasn’t visible on the screen, I meant that the product list isn’t displaying when accessing `http://localhost:8080/products` in the browser.
I’ll attach an image to show you what I mean. Please take a look and make the necessary corrections.


Screenshot 2025-08-06 at 10 26 56 PM

Thank you! 🙂

Wordbe (changes requested):
Re-requesting changes as the notification may not have gone through.

src/main/kotlin/ecommerce/dto/CartItemResponse.kt
bojana-petroska: yes, I see your point! And I tried it in both ways, it still works. I talked to my partner and we were discussing how until now having the data class helped us with debugging. And because I am still in a learning process, I was thinking of leaving them, but of course keeping in mind of their advantages if i really use it or not. I hope this makes sense to you too :S

Wordbe (approved):
Hello Bo!

I’ve confirmed that the product list screen, which wasn’t visible before, is now displaying correctly. Thanks for handling that well.
Also, I can see that your updates have led to a better overall structure. The domain objects now handle core business logic more appropriately, and your tests have become more thorough.
I’ll go ahead and approve and merge this step here.
Great work, and thank you! 🙂

PR #7, opened 2025-08-02
 [spring-ecommerce-order - Step-1] Submit arekiu's mission

arekiu:
Good day Pk! Nice to meet you!

### Overview
This pull request completes Step 1 of the spring-ecommerce-order mission. It includes:

- Migration from JdbcTemplate to Spring Data JPA

- Entity modeling and relationship mapping

- Pagination for product and wishlist endpoints

- Product option feature with validations and quantity update logic

- All changes are based on the previous mission codebase from my partner and follow the mission's process, functional, and programming requirements.

### Notes for reviewer

- We didn´t implement many tests for the repository and service since they rely on Spring Data JPA's built-in methods. But we implemented end-to-end tests.
- In CartHistory (Entity), we included @NamedNativeQuerys and @SqlResultSetMappings. I would like to confirm if this is a good practice, or would you recommend extracting these queries elsewhere.
- Some tests are depending on the data stored in data.sql. Is that a good practice?
- About pagination, the sorting logic is handled in the Controller. Is that the best place for it, or should we consider moving it to a different layer?

Thank you so much for your time to review our project! Have a nice weekend!


src/main/kotlin/ecommerce/controller/AdminProductController.kt
```diff
@@ -0,0 +1,88 @@
...
+    fun getAllProducts(
+        @AdminOnly member: MemberDto,
+        @RequestParam(defaultValue = "0") page: Int,
+        @RequestParam(defaultValue = "5") size: Int,
+        @RequestParam(defaultValue = "id") sortBy: String,
+        @RequestParam(defaultValue = "true") ascending: Boolean,
+    ): ResponseEntity<Page<ProductResponse>> {
+        val sort = if (ascending) Sort.by(sortBy).ascending() else Sort.by(sortBy).descending()
+        val pageable = PageRequest.of(page, size, sort)
+        return ResponseEntity.ok(productService.findAll(pageable))
```
pkeugine: **Q. About pagination, the sorting logic is handled in the Controller. Is that the best place for it, or should we consider moving it to a different layer?**

The easiest and most helpful question to ask ourselves here is:  
**"Would we still need this logic if our view or controller layer changes?"**  
(In other words, is this business logic?)

Imagine we added terminal support and brought back our old friend `OutputView` 😄. In that case, `OutputView` would handle input and output using `println()`, while `productService` would still be responsible for the important business logic that doesn't depend on how data is presented.

To me, sorting feels like something that belongs to the business layer. Even if we change the way users interact with the application, the sorting logic would likely stay the same. It also probably shouldn't vary between views. So in this case, I would keep the sorting logic inside the service layer. This way, developers working on the controller side don't have to keep writing:

```kotlin
val sort = if (ascending) Sort.by(sortBy).ascending() else Sort.by(sortBy).descending()
val pageable = PageRequest.of(page, size, sort)
```

every time they call `productService.findAll()`.

What do you think? :)

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.springframework.stereotype.Service
+
+@Service
+class ProductService(private val productJpaRepository: ProductJpaRepository) {
+    fun findById(id: Long): ProductResponse {
+        val product = productJpaRepository.getByIdOrThrow(id)
+        return product.toDto()
+    }
+
+    fun findAll(pageable: Pageable): Page<ProductResponse> {
```
pkeugine: That brings us to another, slightly higher-level discussion:

`Page` and `Pageable` are part of the package `org.springframework.data.domain`.  
This means our service layer is currently **tightly coupled to Spring Data JPA**, since the function signature directly uses its types.  
That raises the question:  
**"Is this a red flag in terms of flexibility and clean dependency layering?"**

What if we decide to change how we implement pagination later?  
(And believe me, there are many ways to do it, each with its own trade-offs.)  
Or what if we migrate from JPA to a different persistence library altogether?  
(Yes, there are plenty out there, from jOOQ to Exposed, and more.)

It might be worth considering whether this method could be made more tolerant to such changes, so that consumers of the service (like controllers) remain unaffected even if we drop Spring Data JPA.

What do you think? 🙂  
If this feels too abstract or tricky, feel free to talk with other crews or skip the discussion for now!

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    fun addOptionToProduct(
+        id: Long,
+        option: OptionDto,
+    ): ProductResponse {
+        val product = productJpaRepository.getByIdOrThrow(id)
+        product.addOption(option.toEntity())
+//        return productJpaRepository.save(product).toDto()
+        return product.toDto()
```
pkeugine: Let's remove commented out logic so it doesn't get merged into our final branch.

That said, this commented-out line is actually pretty interesting.

Even though we're not explicitly saving the `product`, the new option still ends up persisted in the DB... 🤯  
Why is that happening?

(I would like you to explain to me how JPA's persistence context works 😉 )

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,78 @@
...
+        productRequest: ProductPatchRequest,
+    ): ProductResponse {
+        val product = productJpaRepository.getByIdOrThrow(id)
+        val newProduct =
+            product.copy(
+                name = productRequest.name ?: product.name,
+                price = productRequest.price ?: product.price,
+                imageUrl = productRequest.imageUrl ?: product.imageUrl,
+                options = product.options,
+            )
```
pkeugine:
```suggestion
            product.copy(
                name = productRequest.name ?: product.name,
                price = productRequest.price ?: product.price,
                imageUrl = productRequest.imageUrl ?: product.imageUrl,
            )
```

How does `copy()` work?
To me, putting `product.options` here seems redundant 🙂

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,78 @@
...
+        try {
+            productJpaRepository.save(newProduct)
+            return newProduct.toDto()
+        } catch (e: Exception) {
+            throw ProductUpdateException("Failed to update product, id: $id")
+        }
```
pkeugine: As we work with more and more libraries, there will be plenty of situations where we need to **chain exceptions**, just like you did here. Nice job! 👍

Here's the fun part:

When chaining exceptions, I personally prefer including the **original cause** when throwing a custom exception, like this:

```kotlin
} catch (e: Exception) {
    throw ProductUpdateException("Failed to update product, id: $id", e)
}
```

This does mean we need to add another constructor to our custom exception class, but it gives us much more powerful options when it comes to exception handling.

Why do you think I prefer this pattern? 🙂
Would love to hear your thoughts!

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,38 @@
...
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+
+@Entity
+data class Product(
+    @Column(nullable = false, unique = true)
+    val name: String,
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST, CascadeType.REMOVE])
+    val options: MutableList<Option> = mutableListOf(),
```
pkeugine: Unlike the `Cart` ↔ `Member` relationship, the `Cart` ↔ `CartProduct` and `Product` ↔`Option` pairs seem quite similar in terms of lifecycle with `Cart` and `Product` being the central owners of their respective child entities.

Interestingly, I noticed [you've used `CascadeType.ALL` for `Cart` ↔ `CartProduct`](https://github.com/herotechcourse/spring-ecommerce-order/pull/2#discussion_r2249152091), but not for `Product` ↔ `Option`.

I'm curious about that design choice! Was there a specific reason for treating them differently?

sidenote : took me like 10 minutes to find `↔` symbol 😂.

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,38 @@
...
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+
+@Entity
+data class Product(
+    @Column(nullable = false, unique = true)
+    val name: String,
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST, CascadeType.REMOVE])
```
pkeugine: There's an option called `orphanRemoval` in `@OneToMany` annotation.
Let's look up what it does and see if we need it or not!

src/main/kotlin/ecommerce/exception/ProductAlreadyInDBException.kt
```diff
@@ -0,0 +1,3 @@
+package ecommerce.exception
+
+class ProductAlreadyInDBException(message: String) : RuntimeException(message)
```
pkeugine: Additional review about naming convention, following [this one](https://github.com/herotechcourse/spring-ecommerce-order/pull/2#discussion_r2249143123).

Class names should follow [the general camel case format, even when acronyms are included](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).
In this case, `ProductAlreadyInDBException` could be changed to `ProductAlreadyInDbException`. Why don't we change it? :)

I do understand that this part wasn't current mission's work (I checked git log),
but I just thought it would be nice to apply this as we check package names.

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,56 @@
...
+import ecommerce.repository.getByMemberId
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.PageImpl
+import org.springframework.data.domain.Pageable
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+
+@Transactional
+@Service
+class CartService(
```
pkeugine, line 21: `@Transactional` annotation! 😮 
Since you've added this annotation here, I would like you to tell me
* what this does
* what options it has
* and the difference between `org.springframework.transaction.annotation.Transactional` and `jakarta.transaction.Transactional`

This is a regular question in tech interviews, so we might as well go over it here :)

src/test/kotlin/ecommerce/endToEnd/AdminStatsControllerTest.kt
```diff
@@ -0,0 +1,68 @@
...
+import jakarta.transaction.Transactional
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+
+@Transactional
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+class AdminStatsControllerTest {
```
pkeugine: That brings us to the next topic : `@Transactional` annotation in tests :)
What does this do in tests? Does it differ from what it does in application logic?

(Again, although this part is from the previous mission, I think this topic is pretty important and is related to current mission as well)

src/test/kotlin/ecommerce/endToEnd/AdminStatsControllerTest.kt
```diff
@@ -0,0 +1,68 @@
+package ecommerce.endToEnd
+
+import ecommerce.dto.LoginRequest
+import io.restassured.RestAssured
+import io.restassured.http.ContentType
+import jakarta.transaction.Transactional
```
pkeugine: Which leads us to our final look on the `@Transactional` annotation.
Shouldn't this be `org.springframework.transaction.annotation.Transactional`? 🙂

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,38 @@
...
+    val price: Double,
+    val imageUrl: String,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    fun addOption(option: Option) {
+        require(options.none { it.name == option.name })
+        options.add(option)
+    }
```
pkeugine: Let's go back to the [topic about pros and cons of bidirectional mapping](https://github.com/herotechcourse/spring-ecommerce-order/pull/2#discussion_r2249151076).

I was going to keep this review for next time, but since we're having 2x review session (and a lot of higher level discussion here), might as well talk about this too.

I think there's a missing logic here, which is **CRUTIAL** to bidirectional mapping.
This is actually a perfect example of bidirectional mapping's downside!
I'm not saying you did a bad job, I'm just excited that we're experiencing it rather than just talking about it :)

What should get added here? 🤔

src/main/kotlin/ecommerce/service/ProductService.kt
pkeugine: Finale to our review, here to finalize what we've learned so far.
I say this because this review is related to a lot of reviews from above.

The big reveal : 
**Option is not getting persisted when we try to add it to a product.**
The api response says it did, the E2E test says it did,
but try checking the database after the api call or add another option to the same product right away. You'll notice that there isn't a new option in database, and api response won't show what you've added previously.

I would like you to find out the reason behind this and try solving it!
Like I said, the previous comments are here to help, so if you go over them slowly and tackle this issue, you'll be able to find out what the missing pieces are. (Yeah, there's more than one)

pkeugine (changes requested):
Hello Alex, nice work on this mission 🙂 

Before we start on the reviews in this PR,
I highly recommend checking out the [reviews in Vito's PR](https://github.com/herotechcourse/spring-ecommerce-order/pull/2).
This is because even though you two were pairs (with exact same code)
I'm the reviewer for both of you 😂 

And I assure you, I didn't just copy-and-paste,
I actually extended the review with more depth.
That means Vito's reviews are "easy - mid level"
And your's is "mid - high level, with reviews I was going to add next time"

This is the reason why I recommend checking out Vitto's first.
I would also like you to add comments about my questions in Vitto's PR here,
since it's also related to you too.
I'll react and answer to them.

If you have any questions or opinions, feel free to leave comments or DM me anytime!
Happy Hacking! 🔥 🔥 🔥

src/main/kotlin/ecommerce/controller/AdminProductController.kt
arekiu: Thank you for your detailed feedback!

 I realize now that the current approach was mainly based on patterns we found in some earlier examples, but your point about separation of concerns makes a lot of sense. I agree that handling sorting and pagination in the controller would generate unnecessary code duplication, especially if multiple controllers need similar logic.

I'll refactor the code to move the sorting and pagination logic into the productService. That way, controllers will simply pass the relevant parameters (such as page, size, sortBy, and ascending), and the service layer will take care of processing them.

src/main/kotlin/ecommerce/service/ProductService.kt
arekiu: It´s definitely an abstract topic and it took me some time to understand!

If I understand correctly, your point is that by using Spring specific classes like Page and Pageable everywhere in the code, we make our service layer tightly coupled to Spring Data JPA. This can make the code harder to maintain and could cause problems if we ever decide to switch to a different technology or library.

The better approach would be to create our own dedicated pagination class.  This abstraction would make it easier to change the implementation later without affecting other parts of the application.

Please let me know if I misunderstood anything or if there’s anything else I should consider!

src/main/kotlin/ecommerce/service/ProductService.kt
arekiu: I found this clear explanation in (https://www.baeldung.com/jpa-hibernate-persistence-context):

> "The persistence context is the first-level cache where all the entities are fetched from the database or saved to the database. It sits between our application and persistent storage.
> 
> Persistence context keeps track of any changes made into a managed entity. If anything changes during a transaction, then the entity is marked as dirty. When the transaction completes, these changes are flushed into persistent storage."

Previously, the new options were not being saved to the database, so I added the explicit line productJpaRepository.save(product) to persist the changes. However, I later realized that by adding the @Transactional annotation to the method, I can rely on Spring's transaction management to automatically detect and save changes made to the managed entity.

The @Transactional annotation ensures that the method runs within a database transaction. During this transaction, JPA tracks all changes made to entities it manages. When the transaction completes successfully, JPA automatically flushes these changes to the database.

src/main/kotlin/ecommerce/service/ProductService.kt
arekiu: You are right!
Because copy() copies all properties by default from the original object unless I explicitly override them, what makes the property "options" redundant and unnecessary.

src/main/kotlin/ecommerce/service/ProductService.kt
arekiu: Thanks for the suggestion!
I imagine that you prefer this pattern  because it makes possible to track original error details, what makes debugging much easier.

In this opportunity I included a couple of them to try it and definitely I would implement them in next projects.

src/main/kotlin/ecommerce/model/Product.kt
arekiu: We used CascadeType.ALL on Cart -> CartProduct because the cart fully owns the products inside, so all database actions should be cascaded automatically.

For Product -> Option, we originally only used PERSIST and REMOVE to keep some control on when the options get updated or detached.
But, since the lifecycle feels similar, it’s simpler and safer to use CascadeType.ALL for Product -> Option too.

I updated this now to keep it consistent and avoid confusion. Let me know if you think I should adjust it more!

Sidenote: I tried my luck to find that symbol and googled for the "come and go arrow symbol" xD

src/main/kotlin/ecommerce/model/Product.kt
arekiu: Thank you for the tip!

I initially thought that CascadeType.REMOVE handled all deletions of child entities, but now I understand it only deletes them when the parent entity itself is deleted, and not when a child is simply removed from the collection.

Using `orphanRemoval = true ` makes sure that when an option is removed from the product’s options list, it is also deleted from the database automatically, which fits the expected behavior better.

src/main/kotlin/ecommerce/exception/ProductAlreadyInDBException.kt
arekiu: I thought acronyms should be all capital letters, so this naming rule is really helpful.
I’ll keep this in mind and use it in my future code. Appreciate it!

src/main/kotlin/ecommerce/service/CartService.kt
arekiu: @Transactional makes sure the methods run inside a database transaction (the sequence of operations performed as a single unit of work). So if something goes wrong, all changes get rolled back automatically.

It´s possible to customize it with options like how transactions join, isolation levels, timeouts, and when to roll back.

About the two versions:

- Spring’s @Transactional (org.springframework.transaction.annotation.Transactional) is made for Spring and has more features.
- Jakarta’s (jakarta.transaction.Transactional) is a general standard used outside Spring too.

I use the Spring one because it fits better with the rest of the app.
I appreciate a lot the "questions from tech interviews" :)

src/test/kotlin/ecommerce/endToEnd/AdminStatsControllerTest.kt
arekiu: Thanks for the question. Tt doesn´t matter that is something from previous mission, because it´s always positive to learn something new!

In tests, @Transactional usually means each test runs in a transaction that gets rolled back afterward, so tests don’t leave data behind.

But in this case, because the test uses HTTP calls to a running server (@SpringBootTest with DEFINED_PORT), the rollback might not work like that since the server runs separately.
So data created during tests might stay in the database, and it might be necessary other ways to clean up.

src/test/kotlin/ecommerce/endToEnd/AdminStatsControllerTest.kt
arekiu: Absolutely! The jakarta was there as mistake!

src/main/kotlin/ecommerce/model/Product.kt
arekiu: Thanks for the great point!
I actually got a heads-up about this from my partner.
When adding an Option to the Product's options list, we also need to update the Option to know which Product it belongs to.
This is really important for bidirectional mapping so that both sides stay in sync and avoid inconsistencies.

I won´t forget this `option.product = this`

src/main/kotlin/ecommerce/service/ProductService.kt
arekiu: Oh, looks like my crew-team spoiled the surprise! I actually fixed this issue before! I guess that would’ve been quite a plot twist 😄

But honestly, before I read your comment, I included a call to fetch the updated product in my test and start fixing from there. That process really helped me appreciate how valuable tests are for catching things like this.

```
@Test
    fun addOption() {
        val optionRequest = OptionDto("newOptionTest", 10)

        val response =
            RestAssured.given().log().all()
                .auth().oauth2(token)
                .accept(ContentType.JSON)
                .contentType(ContentType.JSON)
                .body(optionRequest)
                .post("/api/admin/products/add/option/1")
                .then().log().all().extract()

        val **productResponse** =
            RestAssured.given().log().all()
                .auth().oauth2(token)
                .accept(ContentType.JSON)
                .contentType(ContentType.JSON)
                .get("/api/admin/products/1")
                .then().log().all().extract()

        assertThat(response.statusCode()).isEqualTo(HttpStatus.OK.value())

        val json = JSONObject(productResponse.asString())
        val array = json.getJSONArray("options")
        assertThat(array.length()).isEqualTo(2)
        assertThat(array.getJSONObject(1).getString("name")).isEqualTo("newOptionTest")
    }
```

To fix it, @Transactional in the method addOption, and the very important `option.product = this`  were crucial.

arekiu:
Thank you for the super detailed review! 

I enjoyed it a lot because it felt like a journey with a big plot twist!

Please let me know if you would like me to expand in my answers (sometimes I think that when I´m brief, it´s more clear). 
I´m looking forward to hear about your opinion!

Have a nice week :D

src/main/kotlin/ecommerce/service/ProductService.kt
pkeugine: Perfect 👍 
By creating our own classes related to pagination (such as DTOs) and by letting only few classes depend on Spring Data JPA related classes, we'll be able to manage migration or changes more easily, and also allow our logic to be reused in other places without too much hassle. I would also try to keep those classes that depend on Spring Data JPA to be the ones that are close to service layer or persistence layer, since it's related to database operations.

Nice answer! 💯

src/main/kotlin/ecommerce/service/ProductService.kt
pkeugine: Nice explanation with citation! 👍

src/main/kotlin/ecommerce/model/Product.kt
pkeugine: It seems alright to me 👍

pkeugine (approved):
Hello Alex, and it was really fun reading your answers to my reviews :)

I'm also glad that you managed to find the missing pieces in the project and made it possible for us to save entities that weren't getting persisted! I honestly thought it might not be so easy (at least I didn't think it was easy when I first learned JPA 😂 )

I hope you have a look at my reviews at [Vito's PR](https://github.com/herotechcourse/spring-ecommerce-order/pull/2) as well.
If you did, perfect, we are now ready to move onto our final step of this level 👍 

If you have any questions or thoughts that you'd like to discuss about,
feel free to send me DM anytime.

I'll see you soon!
Happy Hacking! 🔥 🔥 🔥

PR #9, opened 2025-08-02
[spring-ecommerce-order - Step-1] Submit san-ghun's mission

san-ghun:
# Overview

This PR is an implementation of a simple HTTP API that imitating some features of eCommerce API, which includes access to resources for product, member, cart item, cart statistics, and option of product. The PR is an extended implementation from previous project, [spring-ecommerce-product](https://github.com/herotechcourse/spring-ecommerce-product). 

In this post, I would like to discuss about the changes have made from previous PR to address how the extended implementation have been done. You can find further information about the previous implementation from the [previous PR](https://github.com/herotechcourse/spring-ecommerce-product/pull/31). 

The mission requires migration of codebase from the relational database based pattern using `JDBC` to the object relational mapping pattern using `JPA`. Therefore, the codebase include lots of change and refactoring including 'model', 'service', and 'repository' which was previously 'dao'.

## Requirements from the mission

1. Migrate from `JDBC` based codebase to `JPA` based codebase.
    - Application of ORM in `JPA` format including entities and repositories.
    - Refactoring domain model and service layer for the application of `JPA`.
    - Understanding of entity mapping.
2. Implement the pagination for `product` and `cartItem`.
    - Application of `Pageable` and `Page<T>`, using Spring Data.
    - Understanding and experimenting what and how can the `JPA`  provide support the application.
3. Implement `Option` model and introduce it to `Product` model.
    - Understanding of unidirectional and bidirectional relationships in ORM.
    - How the **Persistence Context** with JPA is operating with the application.
    - How to perceive the model which was an mapping for DB schema but now is an entity?
5. *(Optional) Provide a way to add options through an admin interface. --not included in this PR*

# Changes

## Edit application property to initialize data for testing purpose

- Apply `spring.jpa.defer-datasource-initialization=true` using `data.sql` script to initialize raw data.
- Edit `data.sql` script file and deactivate `schema.sql` script file.

## Alter previous models into entities

- Introduce `@Entity` to model objects to apply ORM using `JPA`.
- Introduce object relationships to each entities, using `@OneToOne`, `@OneToMany`, `@ManyToOne`.
- Introduce new model `Cart` to represent the cart to hold cart items.

## Edit service layer objects code for changes derived by application of `JPA`

- Inject `JPA` repository based repository objects into service layer objects to get and set data.
- Move some SQL query based function in previous repository (or DAO) into the service layer object as object based functions that manipulate required data. 

## Introduce the pagination feature

- Apply `Pageable` and `Page` provided by the Spring Data to request and return paginated objects.
- Refactor GET requests to view products and cart items to take input from URL parameters for pagination information.
- Refactor response for viewing products and cart items to provide data with DTO transformed from Model Entity, to prevent recursive output.

## Introduce `Option` for `Product`

- Introduce `Option` entity as an sub-information of `Product`.
- Add object validation for condition of object inside of model code for example, in `init{}`.

## Testing

- Introduce `@DataJpaTest` for repository object test.
- Move some business logic based test code from repository test file to service test file.

# Testable Endpoints

**Product**

- `POST /api/products` - create a product
- `GET /api/products` - read all products in `Page`
- `GET /api/products/{id}` - read a product with id
- `PUT /api/products/{id}` - update a product with id
- `DELETE /api/products/{id}` - delete a product with id

Web View

- `GET /` - provide view of list of all products
- `GET /add/product` - provide view of form to add new product
- `GET /update/product/{id}` - provide view of form to update a product

**Options for Product**
- `GET /api/products/{id}/options` - read all options for the product of the id

**Member**

- `POST /api/members/register` - register a member
- `POST /api/members/login` - login a member
- `GET /api/members/me` - get member data

**Cart Item**

- `GET /api/cart` - view user cart items in `Page`
- `POST /api/cart` - add item to user cart
- `DELETE /api/cart/{productId}` - remove an item from user cart
- `PUT /api/cart/{productId}` - update quantity of an item in user cart

**Cart Statistics**

- `GET /api/admin/cart-stats/top5-products` - get top 5 added products inn last 30 days
- `GET /api/admin/cart-stats/active-members` - get active members in last 7 days

# Database schema

```mermaid
erDiagram

    CARTS {
        bigint id PK
    }

    MEMBERS {
        bigint id PK
        varchar email
        varchar password
        varchar role
        bigint cart_id FK
    }

    CART_ITEMS {
        bigint id PK
        integer quantity
        bigint cart_id FK
        bigint member_id FK
        bigint product_id FK
        timestamp created_at
    }

    PRODUCTS {
        bigint id PK
        float price
        varchar image_url
        varchar name
    }

    OPTIONS {
        bigint id PK
        integer quantity
        bigint product_id FK
        varchar name
    }

    %% -- Relationships
    MEMBERS ||--o| CARTS : "cart_id"
    CART_ITEMS }o--|| MEMBERS : "member_id"
    CART_ITEMS }o--|| PRODUCTS : "product_id"
    CART_ITEMS }o--|| CARTS : "cart_id"
    OPTIONS }o--|| PRODUCTS : "product_id"

```

# Additional Note

Because of the migration from `JDBC` to `JPA` that includes application of different design pattern, the codebase feels like untidy and spaghetti. I am sorry for it and hope you understand. 

## Additional Concerns

- How to design and implement an appropriate testing strategy for the implemented features.
- How to separate code that is hard to unit test from code that is testable, and write unit tests for the testable parts.

## Reflection Points

- How to create robust test code that is less vulnerable to pattern changes.
- How to utilize IDE features and tools to react and refactor to pattern changes in development.
- Why developer need to adapt to the OOP with ORM, not RDB with Procedure.
- How to adapt into the mindset of OOP with ORM, not RDB with Procedure.

I will look forward to you review!
Thank you!


README.md
```diff
@@ -1 +1,37 @@
...
+   - Property
+     - [x] `id`
+     - [x] `name` - up to 50 characters
+     - [x] `quantity` - 1 to 100_000_000
+   - Method
+     - [x] decrease the quantity of the option by a specific amount
+     - [x] A `product` must always have at least one `option`
+   - DTO
+     - [x] implement `ProductResponse` to organize product response with pagination
+     - [x] implement `OptionResponse` to organize options response for a product
```
Hyeon9mak, line 37: nice habit 💯

build.gradle.kts
```diff
@@ -20,14 +21,25 @@ repositories {
 }
 
 dependencies {
+    // TODO: maybe remove jdbc later
     implementation("org.springframework.boot:spring-boot-starter-jdbc")
+
+    implementation("org.springframework.boot:spring-boot-starter-data-jpa")
```
Hyeon9mak: So, did u decide remove or left library? 😉 
ambiguous comment make ur fellow developer's confusion!

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice
+class GlobalExceptionHandler {
+    @ExceptionHandler(NotFoundException::class)
+    fun handleNotFoundException(e: NotFoundException): ResponseEntity<Void> {
+        println("NotFoundException occurred: " + e.message)
+        return ResponseEntity.notFound().build()
+    }
```
Hyeon9mak: What's different about between printing & logging?
What are the reasons for using a logger rather than print in production environments? 😉

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @ExceptionHandler(AuthorizationException::class)
+    fun handleAuthorizationException(e: AuthorizationException): ResponseEntity<Map<String, Any>> {
+        val error = mapOf("authorization" to e.message)
+        val errorBody = mapOf("errors" to error)
+        println("AuthorizationException occurred: $errorBody")
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build()
+    }
+}
```
Hyeon9mak: What would happen if an exception occurs that we didn't explicitly handle?
Hint: inheritance exception.

src/main/kotlin/ecommerce/auth/AuthorizationExtractor.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import jakarta.servlet.http.HttpServletRequest
+
+interface AuthorizationExtractor<T> {
+    fun extract(request: HttpServletRequest): T
+
+    companion object {
+        const val AUTHORIZATION = "Authorization"
+    }
+}
```
Hyeon9mak: samething in `org.springframework.http.HttpHeaders.AUTHORIZATION` 😉

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,71 @@
...
+    fun findMemberByEmail(email: String): Member? =
+        memberRepository.findByEmail(email) ?: throw AuthorizationException(MESSAGE_INVALID_EMAIL)
+
+    fun findMemberByToken(token: String): Member {
+        if (!jwtTokenProvider.validateToken(token)) {
+            throw AuthorizationException(MESSAGE_INVALID_TOKEN)
+        }
+        val email = jwtTokenProvider.getPayload(token)
+        val member = memberRepository.findByEmail(email) ?: throw AuthorizationException(MESSAGE_INVALID_EMAIL)
+        return member
```
Hyeon9mak: i think jwtTokenProvider can throw exception in itself! 😄 

```suggestion
    fun findMemberByToken(token: String): Member {
        jwtTokenProvider.validateToken(token)
        val email = jwtTokenProvider.getPayload(token)
        return memberRepository.findByEmail(email) ?: throw AuthorizationException(MESSAGE_INVALID_EMAIL)
    }
```

src/main/kotlin/ecommerce/controller/api/CartController.kt
```diff
@@ -0,0 +1,79 @@
...
+    @PutMapping("/{productId}")
+    fun updateQuantity(
+        @PathVariable productId: Long,
+        @RequestBody @Valid cartForm: CartUpdateQuantityForm,
+        @LoginMember member: Member,
+    ): ResponseEntity<String> {
+        val memberId = member.id
+        val message = cartItemService.updateQuantity(memberId, productId, cartForm.quantity)
+        return ResponseEntity.ok(message)
+    }
```
Hyeon9mak: What's different between HTTP method PUT & PATCH?

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
```diff
@@ -0,0 +1,35 @@
...
+    private val cartStatisticsService: CartStatisticsService,
+) {
+    @GetMapping("/top5-products")
+    fun getTop5Products(
+        @LoginMember member: Member,
+    ): ResponseEntity<List<TopProductStats>> {
+        member.id
+        val stats = cartStatisticsService.getTop5AddedProductsInLast30Days()
+        return ResponseEntity.ok(stats)
+    }
```
Hyeon9mak: Like a lost little child `member.id`. I wonder what brought it here!

src/main/kotlin/ecommerce/dto/ProductForm.kt
```diff
@@ -0,0 +1,29 @@
...
+    @field:Pattern(regexp = "^[a-zA-Z0-9()\\[\\]+\\-&/_ ]+$", message = "Contains unallowed character")
+    @field:Size(min = 1, max = 15, message = "Must be no more than 15 characters, including spaces")
+    var name: String = String(),
+    @field:Positive(message = "Product price must be greater than zero")
+    var price: Double = Double.NaN,
+    @field:NotBlank(message = "Product image URL is required")
+    @field:Pattern(regexp = "^https?://.*", message = "Must start with 'https://'.")
+    var imageUrl: String = String(),
+) {
+    companion object {
```
Hyeon9mak, line 21: (just a question. feel free 😄 )
What's different between dto & entity in ur opinion?

src/main/kotlin/ecommerce/ui/CheckAdminInterceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+        val token = authorizationExtractor.extract(request)
+        val member = authService.findMemberByToken(token)
+        if (member.role != "admin") {
+            response.status = 401
+            return false
+        }
+
+        return true
+    }
+}
```
Hyeon9mak: What's different between status code 401 & 403?

src/main/resources/application.properties
```diff
@@ -1 +1,14 @@
...
+jwt.secret-key=X1yk3dV+NQemyyJbg6WzioaRR1NArxZxjMvvroTZauk=
+jwt.expiration-ms=3600000
+
+spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.format_sql=true
+spring.jpa.hibernate.ddl-auto=create
+
+spring.jpa.defer-datasource-initialization=true
+
+logging.level.org.hibernate.orm.jdbc.bind=TRACE
```
Hyeon9mak: https://medium.com/@hyeon9mak/about-no-newline-at-end-of-file-warning-on-github-b42d14dccc93 😉

Hyeon9mak (changes requested):
Hi San! Sorry for late 🙏 
Ur reviewer here, and u did a good job implementing this mission. 👍

I left comments mainly focused on Spring MVC, DTO, and Entity.
Let's start by getting a good grasp of these concepts first, then we can take a deeper dive into JPA step by step. 😉

Feel free to ask anytime if you have any questions!

README.md
san-ghun: Thanks!

build.gradle.kts
san-ghun: My bad, sorry for giving you an confusion.
I handle to remove it.

src/main/kotlin/ecommerce/auth/AuthorizationExtractor.kt
san-ghun: Thanks, good to know!
I should have to dig a bit more into APIs provided by the Spring Framework.

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
san-ghun: My bad. Those are traces of previous projects. No excuse, those are my fault. Sorry about those. 
I will use the lost child to become a guardian of the application like this:

```kotlin
@GetMapping("/top5-products")
fun getTop5Products(
    @LoginMember member: Member,
): ResponseEntity<List<TopProductStats>> {
    memberRepository.findById(member.id).getOrNull()
        ?: throw AuthorizationException(String.format(MESSAGE_UNAUTHORIZED, member.id))
    val stats = cartStatisticsService.getTop5AddedProductsInLast30Days()
    return ResponseEntity.ok(stats)
}
```

Thanks for pointing out!

src/main/kotlin/ecommerce/dto/ProductForm.kt
san-ghun: A DTO (Data Transfer Object) is primarily used in the controller layer for handling both incoming requests and outgoing responses. It may also be used in the service layer to present data in a specific format, often to control or limit the exposure of internal data.

An Entity represents a database table or a record and is used to map application objects to the dedicated database schema using ORM.

By the way, the most confusing part for me was actually these;

- What is the difference between DAO and Repository?
    - My current understanding is that, those are design patterns to implement data access between application and datasource.
- What is the difference between Entity and Model?
    - Entity must have its `id` as it represents a record from table of database, but model is an abstraction of an actual concept that the application wants to actualize(?). Main difference I think from "Persistence"? Entity is yes, but Model is not.
- What is the Domain, is it Entity or Model?
    - I think Model is a representation of a Domain. The Domain is an actual concept that an application wants to actualize through usage of model, and entity if there is datasource.

src/main/resources/application.properties
san-ghun: Thanks for pointing out! 
Thanks to you, I now also set the setting up in IntelliJ. LOL

src/main/kotlin/ecommerce/ui/CheckAdminInterceptor.kt
san-ghun: oooohhhhhhh!!!

The HTTP `401 Unauthorized` response status code indicates that the client request has not been completed because it lacks of valid authentication credentials for the requested resource.

The HTTP `403 Forbidden` response status code indicates that the server understands the request but refuses to authorize it. The access to the resource is tied to the custom authorization logic.

You make me realize that it is better to response with status `403 Forbidden` in this case.

Thanks!

src/main/kotlin/ecommerce/service/AuthService.kt
san-ghun: Okay, that make sense. 
I will let 'jwtTokenProvider.validateToken()` to throw its exceptions, and introduce new exception handler in global exception handler.

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
san-ghun: Printing is an action writing output directly to the standard output (stdout), mostly purposed for quick debugging during development.

Logging is the process of recording structured text information about the status and activities of a program during its runtime.

Logging has levels to control category or amount of information to provide. Most of logging support async and optimized to language or platform to reduce overload on application. Logs  can be written in files and can be managed the outputs externally. 

I didn't know much about logging until you asked, but thanks to you, I learned about it.

san-ghun: I may use logging on next project.

Which logger do you like and can recommend to me?

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
san-ghun: During working on this project, what I noticed was that when I faced unhandled exceptions the applicatioin responded with `HTTP 500 Internal Server Error`.

I've read an article about 'Propagation' of exception, it may related with you hint 'inheritance exception'. What I understood from the article was that `RuntimeException` might going to handle most of the family of unhandled exception.

I am going to add these two new exception handlers in global exception handler;

```kotlin
    @ExceptionHandler(Exception::class)
    fun handleGeneric(e: Exception): ResponseEntity<Void> {
        val error = mapOf("error" to e.message)
        val errorBody = mapOf("errors" to error)
        println("Unexpected Exception occurred: $errorBody")
    }

    @ExceptionHandler(RuntimeException::class)
    fun handleRuntimeException(e: RuntimeException): ResponseEntity<Map<String, Any>> {
        val error = mapOf("error" to e.message)
        val errorBody = mapOf("errors" to error)
        println("RuntimeException occurred: $errorBody")
        return ResponseEntity.internalServerError().body(errorBody)
    }
```

src/main/kotlin/ecommerce/controller/api/CartController.kt
san-ghun: PUT replaces the entire resource, while PATCH updates part of the resource.
PUT requires entire fields of an object, while PATCH use only one or part of fields of an object.

I suppose Spring or Spring Boot support HTTP method PATCH as a protocol, however may not support the operation of the method automatically, which means that dev need to implement partial update logic manually. I may use PATCH on other project if it fit well with certain requirement I guess, but not in this project.

san-ghun:
Hey GuMak!

Hope you have a best summer!
Always thanks for your review with valuable questions.

I tried to apply your reviews.
I resolved conversations those I think don't need further conversation.
However I remained other conversations open, so you can find those in quick.
Feel free to resolve conversation when those are considered enough for you.

P.S. I thought it is okay to `resolve conversation` when I apply the review opinion, but Jason mentioned to us (students) don't have to. So I remain some conversation kept open.

## Main Changes

###  Add three new exception handlers in `GlobalExceptionHandler`

- introduce `handleGeneric()` to handle generic `Exception` and `HandleRuntimeException()` to handle `RuntimeException`, both returning `500 Internal Server Error`.
- introduce `handleJwtException()` to handle `JwtException` from JWT token provider and return `401 Unauthorized`.
- edit `handleIllegalArgumentException()` to return `500 Internal Server Error` without body.

### Apply annotation `Transactional` in service layer

- add `Transactional` annotation in service layer in class level to assure DB manipulation work under transaction and commit.

## Additional Note

Thanks to your review, I could improve readability of general codebase.
And, I really appreciate your thoughtful comments that make me learn and study even deeper.
Thank you!

### ref) Curated Questions from GuMak

- Difference between **Printing** and **Logging**?
	- What are the reasons for using logging, rather than printing in production environment?
- What happen if an exception **not caught** by code or handled via a global handler?
	- How can one prepare for such cases?
	- hint: inheritance exception
- Difference between HTTP method **PUT** and **PATCH**?
- Difference between DTO and Entity?
	- Difference between DAO and Repository?
	- Difference between Entity and Model?
	- What is the Domain, is it Entity or Model?
- Difference between HTTP Response status code 401 and 403?
- What are the reasons for a **newline** at the end of file (EOF)?

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,86 @@
...
+
+@ControllerAdvice
+class GlobalExceptionHandler {
+    @ExceptionHandler(Exception::class)
+    fun handleGeneric(e: Exception): ResponseEntity<Void> {
+        val error = mapOf("error" to e.message)
+        val errorBody = mapOf("errors" to error)
+        println("Unexpected Exception occurred: $errorBody")
+        return ResponseEntity.internalServerError().build()
+    }
```
Hyeon9mak, line 22: > Which logger do you like and can recommend to me?

If you haven’t decided which logger to use yet,
how about starting with the default one provided by Spring Boot and gradually exploring the differences as you go? 😉 
https://docs.spring.io/spring-boot/reference/features/logging.html

Hyeon9mak: > I am going to add these two new exception handlers in global exception handler;

nice 👍 

Taking it a step further, since these are unexpected errors, there's a risk that sensitive server-related information could be exposed to the user. That's why I prefer returning a general guidance message in the response body rather than detailed information. 😄

src/main/kotlin/ecommerce/controller/api/CartController.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    @PutMapping("/{productId}")
+    fun updateQuantity(
+        @PathVariable productId: Long,
+        @RequestBody @Valid cartForm: CartUpdateQuantityForm,
+        @LoginMember member: Member,
+    ): ResponseEntity<String> {
+        val message = cartItemService.updateQuantity(member.id, productId, cartForm.quantity)
+        return ResponseEntity.ok(message)
+    }
```
Hyeon9mak: > I suppose Spring or Spring Boot support HTTP method PATCH as a protocol, however may not support the operation of the method automatically, which means that dev need to implement partial update logic manually. I may use PATCH on other project if it fit well with certain requirement I guess, but not in this project.

Personally, I think PATCH is a better fit than PUT for an API that modifies quantity. However, it's true that PUT is more widely used in general. The most important thing is to understand the difference between PUT and PATCH and make an intentional choice when using them — and I think you've done that well. 😄

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
```diff
@@ -0,0 +1,45 @@
...
+) {
+    @GetMapping("/top5-products")
+    fun getTop5Products(
+        @LoginMember member: Member,
+    ): ResponseEntity<List<TopProductStats>> {
+        memberRepository.findById(member.id).getOrNull()
+            ?: throw AuthorizationException(String.format(MESSAGE_UNAUTHORIZED, member.id))
+        val stats = cartStatisticsService.getTop5AddedProductsInLast30Days()
+        return ResponseEntity.ok(stats)
+    }
```
Hyeon9mak: > I will use the lost child to become a guardian of the application like this:

🤣 

Taking it a step further, Controller layer represent HTTP protocol on Kotlin code.
So in my opinion, validate member is not Controller's responsibility.
Looks like our business logics. What do u think about this? 😉

src/main/kotlin/ecommerce/ui/CheckAdminInterceptor.kt
```diff
@@ -0,0 +1,29 @@
...
+        handler: Any,
+    ): Boolean {
+        val token = authorizationExtractor.extract(request)
+        val member = authService.findMemberByToken(token)
+        if (member.role != "admin") {
+            response.status = 403
+            return false
+        }
+
+        return true
```
Hyeon9mak: Check the `org.springframework.http.HttpStatus`!
it's makes reduce human error 😄 

```suggestion
        val token = authorizationExtractor.extract(request)
        val member = authService.findMemberByToken(token)
        if (member.role != "admin") {
            response.status = HttpStatus.FORBIDDEN.value()
            return false
        }

        return true
```

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,40 @@
...
+@Table(name = "cart_items")
+class CartItem(
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "member_id", nullable = false)
+    var member: Member,
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "product_id", nullable = false)
+    var product: Product,
+    @Column(nullable = false)
+    var quantity: Int = 1,
```
Hyeon9mak: What was your reasoning for choosing FetchType.LAZY?

For example, when displaying CartItem on a shopping page, it would be natural for the Product information to be shown together.
In that case, EAGER might actually be a better fit than LAZY.

How about considering fetch strategies from a business perspective like this? 😉

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,40 @@
...
+    @Column(nullable = false)
+    val createdAt: LocalDateTime = LocalDateTime.now(),
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
```
Hyeon9mak, line 29: Do you know the exact reason why setting id to 0L doesn't cause any issues? 😉 
check `org.springframework.data.repository.core.support.AbstractEntityInformation.class`!

In SpringDataJpa default settings, null or zero(0L) value represent "it's not yet granted ID"
image

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,51 @@
...
+class Member(
+    @Column(nullable = false, unique = true)
+    val email: String,
+    @Column(nullable = false)
+    val password: String,
+    val role: String? = null,
+    @OneToOne(cascade = [CascadeType.PERSIST])
+    @JoinColumn
+    var cart: Cart = Cart(),
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
```
Hyeon9mak: What's effect of `CascadeType.PERSIST`? 😉

src/main/kotlin/ecommerce/model/Option.kt
```diff
@@ -0,0 +1,33 @@
...
+    val id: Long = 0L,
+) {
+    @ManyToOne(optional = false)
+    var product: Product? = null
+
+    init {
+        require(name.length <= 50) { "Option name must have length up to 50" }
+        require(quantity in 1..100_000_000) { "Option quantity must be between 1 and 100,000,000" }
+        require(name.matches(Regex("^[a-zA-Z0-9()\\[\\]+\\-&/_ ]+$"))) { "Option name must contain valid characters" }
+    }
```
Hyeon9mak, line 28: looks like... very good class to generate test code 👀

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,51 @@
...
+    val id: Long = 0L,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST])
+    val options: MutableList<Option> = options.toMutableList()
+
+    init {
+        require(options.isNotEmpty()) { "Options must not be empty" }
+        require(options.size == options.map { it.name }.distinct().size) { "Option names must be distinct" }
+        options.forEach { it.product = this }
+    }
```
Hyeon9mak: Question event~~
What's different between `@OneToMany` annotation option `CascadeType.REMOVE` and `orphanremoval = true` ?

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.repository.PagingAndSortingRepository
+import org.springframework.stereotype.Repository
+import java.util.Optional
+
+@Repository
+interface CartItemRepository : JpaRepository<CartItem, Long>, PagingAndSortingRepository<CartItem, Long> {
+    fun findByMemberAndProduct(
+        member: Member,
+        product: Product,
```
Hyeon9mak: this interface really need `@Repository` annotation? 😉 
check the`JpaRepository` interface's implementation `SimpleJpaRepository.class`.

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,65 @@
...
+        return AuthResponse(accessToken)
+    }
+
+    fun findMemberByToken(token: String): Member {
+        jwtTokenProvider.validateToken(token)
+        val email = jwtTokenProvider.getPayload(token)
+        val member = memberRepository.findByEmail(email) ?: throw AuthorizationException(MESSAGE_INVALID_EMAIL)
+        return member
+    }
+
```
Hyeon9mak, line 47: what's different between `@Transactional(readOnly = false)`, `@Transactional(readOnly = true)` ?

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,20 @@
...
+
+@Entity
+@Table(name = "carts")
+class Cart(
+    @OneToMany(cascade = [CascadeType.PERSIST])
+    @JoinColumn(name = "cart_id", referencedColumnName = "id")
+    var cartItems: Set<CartItem> = emptySet(),
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+)
```
Hyeon9mak, line 20: I'd like to take a moment to redefine some of the commonly used terms: DTO, Entity, Domain, Model, DAO, and Repository.
This is based on my personal understanding, so feel free to disagree — but I hope this gives you a better sense of the nuance behind the feedback I leave in code reviews.

Hyeon9mak: DTO (Data Transfer Object)
As the name suggests, a DTO is an object used to transfer data across layers (e.g. between controller ↔ service, or even between persistence ↔ domain).
The main purpose of DTOs is to decouple layer boundaries, so that changes in one layer don’t ripple unnecessarily into others.

For example, if no DTOs were used and your controller accepted domain objects directly, changing a controller field could lead to unwanted changes in the domain and repository layers as well. DTOs prevent this by isolating the data structure at the boundary.

That’s why I believe business logic or validation rules should not be embedded in DTOs.
DTOs can be easily created and discarded as needed, and often many variations of them exist (e.g. ProductFullDTO, ProductPriceDTO, etc.).
If we embed too much logic in them, it becomes hard for developers to track down where certain business rules are defined.

Hyeon9mak: Entity
The concept of an Entity is not limited to RDBs — it's a broader concept that applies wherever we deal with identity-based equality.

For example, given the following two records:

- id = 1, name = "san"
- id = 1, name = "Hyeon9mak"

Should these be considered the same or different?

And what about:

- id = 1, name = "san"
- id = 99, name = "san"

Depending on how you define equality (identity-based vs. value-based), your entity's equals and hashCode should be overridden accordingly.

Hyeon9mak: Domain refers to the business layer itself — the core logic, rules, policies, and constraints that represent the problem space our application is trying to solve.

In a typical controller → domain → repository structure:

- The controller and repository are technical infrastructure.
- The domain is where business logic lives.

So when a new team member joins and wants to understand the business, they should be able to start with the domain layer.

Hyeon9mak: Model
The term "Model" is a bit abstract, but in JVM context, I treat it as a concrete class representing a business concept, often containing validation and rules beyond just data fields.

This is where confusion between Model and Entity arises.

You can:

- Treat your Entity as a Model (with business rules included),
- Or separate them: Entity purely for persistence, Model for domain logic (i.e. a "domain model").

They’re not mutually exclusive — you can layer both on the same class (like wearing sunglasses and a mask on the same face).

Hyeon9mak: DAO (Data Access Object)
A DAO is a concrete implementation used to directly interact with the persistence layer.

This often includes persistence-specific details, such as configurations for JdbcTemplate, Hibernate, MyBatis, DB connection info, etc.

Hyeon9mak: Repository
Repository is an abstraction similar to Java's Collection interface.
It defines operations like “save”, “find”, “delete”, but doesn’t care how they’re done.

A good way to implement this is:

- Define a Repository interface that declares the contract.
- Implement it using a DAO which handles the actual database logic.

Hyeon9mak (changes requested):
Hi San! nice work to reflect feedbacks!

In this time, I've left comments about JPA and ur question answer (dto, entity, etc ..)
Please take your time to review, and feel free to ask if you have any questions!

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
san-ghun: I am going to dig out more about the logging with **Logback**, and it seems like the Logback related with **SLF4J**. I might use these tools for logging.

As you mentioned, it would be better not to expose unexpected error message from server to user. I will remove response body for non-specific exception handlers.

src/main/kotlin/ecommerce/controller/api/CartController.kt
san-ghun: After your reply, I searched and read about the PUT and PATCh again. I realized that I am actually doing PATCH but using PUT method, which means I already have the partial update logic. 
Thanks to you, I am going to change the method into `PATCH` mapping.

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
san-ghun: Yes that make sense. 
These endpoints are 'admin' specific endpoints so those get caught by `CheckAdminInterceptor`, which means the access to the resource is already taken cared.
I am going to add some end-to-end test for these endpoints to make sure these are caught by the interceptor and I will remove our little guardians.
Thanks!

src/main/kotlin/ecommerce/ui/CheckAdminInterceptor.kt
san-ghun: Thanks! Certainly it is much better!

src/main/kotlin/ecommerce/model/CartItem.kt
san-ghun: I thought those as reversed way. lol 
Your right, my bad. Since `@ManyToOne`'s default fetch strategy is EAGER, I am going to remove the fetch assignments.

As you mentioned me to consider fetch strategy, I am going to update fetch strategy for `cart` in `Member` entity. Because it would be better to fetch cart of member in lazy, since member in general would display its email and other primitive data in usual cases, rather than its cart.

```kotlin
// from
    @OneToOne(cascade = [CascadeType.PERSIST])
    @JoinColumn
    var cart: Cart = Cart(),

// to
    @OneToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST])
    @JoinColumn
    var cart: Cart = Cart(),
```

src/main/kotlin/ecommerce/model/CartItem.kt
san-ghun: Oh great to know! Thanks for letting me know!
I used 0L because Jason used it in that way. If I didn't saw how he did I might used null.

src/main/kotlin/ecommerce/model/Member.kt
san-ghun: In this case, when the `Member` persist, its `cart` will also persist.
What I understood was, cascade persist for child creation, in here is `cart`

src/main/kotlin/ecommerce/model/Option.kt
san-ghun: Yes, thanks for pointing out! I forgot about that!

src/main/kotlin/ecommerce/model/Product.kt
san-ghun: Yeah I was also curious the difference.
The difference is based on which side is being removed and the effect from the action.

`CascadeType.REMOVE` is when I remove a `Product` then its children `options` also deleted.
`orphanRemoval = true` is when I remove a child 'option' from its parent 'Product' to remove relationship, then also delete the `option` from DB table.

I am going to add `orphanRemoval = true` for `options` in `Product`.

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
san-ghun: Oh, so `@Repository` annotation is only available on implemented **class**.
I am going to remove the annotation from repository interfaces.
I didn't read the documentation carefully.
Thanks for pointing out!

src/main/kotlin/ecommerce/service/AuthService.kt
san-ghun: Oh this is new one. I didn't even knew this option.

If `readOnly = true`, the transaction is optimized for read-only at runtime.
However, the doc say, this is just a hint for the transaction subsystem. So, when a transaction manager (I think it is mentioning about certain database program?) which do not support or interpret the hint will ignore and process the write operation.
The default value is `false`.

src/main/kotlin/ecommerce/model/Cart.kt
san-ghun: I really appreciate that you share your own definition and understanding about these concepts!
It may take some time for me to digest these, however it is enlightening and building my understanding more concrete. I will meditate on these precious sentences into my mind.

The terminologies Entity, Domain and Model were so confused (and still), but now those are much more clear!
Thank you!

san-ghun:
Hello GuMak!

Thanks for your precious review with valuable questions, again!

I've applied your review comments and answered you questions.

## Main Changes

### Replace HTTP method `PUT` to `PATCH` 

- Change HTTP method mapping for `updateQuantity()` in `CartController` from PUT to PATCH.

### Review and modify `CartItem` and `Product` entities

- Remove unnecessary fetch strategy `FetchType.LAZY` for `member` and `product` from `CartItem` which are `@ManyToOne`, since those are EAGER for the immediate usage from the CartItem object.
- Apply `orphanRemoval = true` to `options` in `Product`

## Additional Notes

I might have to take some time to apply logging with SLF4J using Logback. The document page you linked to me is helpful. Thanks!

Also, it may take a while for me to digest you understanding and make my own definitions for DTO, Entity, Domain, Model, DAO, Repository. My level of understanding and knowledge is not matured enough to analyze and judge you definitions. However, your definitions and understandings are giving me much inspiration to understand those topics deeper. 

I hope one day I can have an constructive discussion with you for the definitions and understandings related with those topics.

src/main/kotlin/ecommerce/controller/api/CartController.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    @PatchMapping("/{productId}")
+    fun updateQuantity(
+        @PathVariable productId: Long,
+        @RequestBody @Valid cartForm: CartUpdateQuantityForm,
+        @LoginMember member: Member,
+    ): ResponseEntity<String> {
+        val message = cartItemService.updateQuantity(member.id, productId, cartForm.quantity)
+        return ResponseEntity.ok(message)
```
Hyeon9mak, line 60: 😄 👍

src/main/kotlin/ecommerce/controller/api/CartStatsController.kt
```diff
@@ -0,0 +1,27 @@
...
+    @GetMapping("/top5-products")
+    fun getTop5Products(): ResponseEntity<List<TopProductStats>> {
+        val stats = cartStatisticsService.getTop5AddedProductsInLast30Days()
+        return ResponseEntity.ok(stats)
+    }
+
+    @GetMapping("/active-members")
+    fun getActiveMembers(): ResponseEntity<List<ActiveMemberInfo>> {
+        val members = cartStatisticsService.getActiveMembersInLast7Days()
+        return ResponseEntity.ok(members)
```
Hyeon9mak, line 25: > These endpoints are 'admin' specific endpoints so those get caught by CheckAdminInterceptor, which means the access to the resource is already taken cared.

💯 nice opinion.
That looks much cleaner and clearer now!

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,52 @@
...
+class Member(
+    @Column(nullable = false, unique = true)
+    val email: String,
+    @Column(nullable = false)
+    val password: String,
+    val role: String? = null,
+    @OneToOne(fetch = FetchType.LAZY, cascade = [CascadeType.PERSIST])
+    @JoinColumn
+    var cart: Cart = Cart(),
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
```
Hyeon9mak, line 28: nice! The `member` doesn't always need to access their cart items, so using `LAZY` loading is appropriate here.

Hyeon9mak: > In this case, when the Member persist, its cart will also persist.
What I understood was, cascade persist for child creation, in here is cart

I was wondering if this was an intentional choice. 😄 👍 💯

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,51 @@
...
+    var price: Double,
+    @Column(nullable = false)
+    var imageUrl: String,
+    options: List<Option> = emptyList(),
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST], orphanRemoval = true)
+    val options: MutableList<Option> = options.toMutableList()
+
```
Hyeon9mak, line 27: If a developer accidentally breaks a relationship and there is no orphanRemoval setting, leftover data might remain. This setting helps prevent such issues. 😄

src/main/kotlin/ecommerce/service/AuthService.kt
Hyeon9mak: Actually, annotations are just simple markers, and what really matters is how the framework reads and acts upon them. Hibernate, the JPA implementation, interprets the annotation and if readOnly = true, it does not perform the final commit. Therefore, CREATE/UPDATE/DELETE operations are not applied.

If readOnly = false, these operations are executed, and in case of an exception, it supports rollback back to the starting point.

There are many other convenience features provided as well, but I think it's better to learn those directly through experience. 😉

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,52 @@
...
+        ): Member {
+            return Member(member.email, member.password, id = id)
+        }
+
+        fun from(loginForm: LoginForm): Member {
+            return Member(email = loginForm.email, password = loginForm.password)
+        }
+
+        fun from(registerForm: RegisterForm): Member {
+            return Member(email = registerForm.email, password = registerForm.password)
```
Hyeon9mak, line 44: so, Entity class's equals and hashCode should be overridden accordingly for ID 😄

Hyeon9mak (approved):
Hi San! good work to reflect feedbacks. 👍 
I've left some simple comments.

These concepts can be challenging.
I believe you can gradually make them your own through various experiences. 👍 
Let's move on to the final stage. See you in the next step!

PR #10, opened 2025-08-04
[spring-ecommerce-order - Step-1] Submit saragrosser's mission

saragrosser:
This PR represents a foundational shift from simple JDBC to a more expressive domain model using JPA and Spring Data. From entity modeling to pagination and business rules around product options, each part helped me strengthen my understanding of core Spring Boot concepts. It helped me shift from basic entity definitions to more intentional, safe, and maintainable domain models. I revisited the design of Product and Option to remove validation logic from the entity, adopt DTO patterns consistently, and start writing more testable business logic.

Highlights:
	•	Introduced ProductRequest and OptionRequest DTOs to decouple input validation from entity logic.
	•	Ensured that business rules (like unique option names) are checked in the service layer, not inside entities.
	•	Updated ProductService to follow clean coding practices:
	•	Replaced inline logic with reusable toResponse() method.
	•	Made create() and update() more testable by keeping side effects explicit.
	•	Used "@Transactional" where needed to ensure data consistency.

⸻

Challenges Faced
	•	Got stuck on a infinite loop due to the bilateral relation created 
	•	Ensured all save() and saveAll() matchers used type-safe generics to avoid runtime issues.
	•	Rewrote and debugged multiple tests that were silently failing or passing without proper assertions.

⸻

This step helped me understand how to:
        •       Transitioning to JPA helped me better understand entity lifecycles and repositories.
	•	Write code that separates concerns (DTO ↔ Entity ↔ DB).
	•	Use Mockito properly in Kotlin, even with edge cases like save(any()).
	•	Improve my test structure to prepare for more robust flows in later steps.

Thank you so much for reviewing I welcome all suggestions and improvements


src/main/kotlin/ecommerce/entity/Cart.kt
```diff
@@ -0,0 +1,27 @@
...
+
+@Entity
+@Table(name = "cart")
+class Cart(
+    @ManyToOne
+    @JoinColumn(name = "member_id", nullable = false)
+    val member: Member,
+    @ManyToOne
+    @JoinColumn(name = "product_id", nullable = false)
+    val product: Product,
```
da-nyee: Entity mapping 👍

src/main/kotlin/ecommerce/entity/Cart.kt
```diff
@@ -0,0 +1,27 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import java.time.LocalDateTime
+
+@Entity
+@Table(name = "cart")
+class Cart(
+    @ManyToOne
```
da-nyee: Have you heard of `FetchType`?
It'd be better to set it to `LAZY`, and I'd recommend looking into the differences between the options.

src/main/kotlin/ecommerce/entity/Member.kt
```diff
@@ -0,0 +1,24 @@
...
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+
+@Entity
+@Table(name = "members")
+class Member(
```
da-nyee, line 12: What about making the no-args constructor private?

src/main/kotlin/ecommerce/entity/Product.kt
```diff
@@ -0,0 +1,38 @@
...
+    @Column(name = "price", nullable = false)
+    val price: Double,
+    @Column(name = "image_url")
+    val imageUrl: String,
+    options: List<Option>,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST, CascadeType.ALL])
```
da-nyee: Are both cascade types needed? I think one of them might be unnecessary 👀

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+@Entity
+@Table(name = "product_option")
+class Option(
+    @Column(name = "name", nullable = false, length = 50)
+    var name: String,
+    @Column(nullable = false)
+    var quantity: Int,
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "product_id", nullable = false)
+    @JsonIgnore
```
da-nyee: Is this necessary? 👀

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        require(quantity in 1 until 100_000_000) {
+            "Quantity must be between 1 and 99,999,999."
+        }
+    }
+
+    fun subtract(amount: Int) {
+        if (amount <= 0) throw IllegalArgumentException("Amount must be positive.")
+        if (quantity - amount < 0) throw IllegalStateException("Not enough quantity.")
```
da-nyee: It looks good, but what do you think about using `check` here? 👀

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,153 @@
...
+            )
+
+        val options =
+            request.options.map {
+                Option(
+                    name = it.name,
+                    quantity = it.quantity,
+                    product = product,
+                )
+            }
```
da-nyee: According to the requirement:

image

<br/>
<br/>

I think this might need to be fixed 🤔

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    init {
+        require(name.length <= 50) { "Option name must be 50 characters or fewer." }
+        require(name.matches(Regex("^[a-zA-Z0-9 ()\\[\\]+\\-&/_]*$"))) {
+            "Option name contains invalid characters."
+        }
+
+        require(quantity in 1 until 100_000_000) {
```
da-nyee, line 33: Underscore 👍

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,153 @@
...
+        request: ProductRequest,
+    ): ProductResponse {
+        val existingProduct =
+            productRepository.findById(id).orElseThrow {
+                NoSuchElementException("Product with ID $id not found.")
+            }
+
+        if (productRepository.existsByNameAndIdNot(request.name, id)) {
+            throw DuplicateProductNameException()
+        }
```
da-nyee, line 95: You've already checked for duplicate product names — good job 👍

But there's one more thing you might need to check — it's part of the requirements:

image

da-nyee (changes requested):
Hi Saragrosser! 🙌
Sorry for the late review 🥺

I've gone through the code based on the requirements.
It seems like you're doing great so far 👍

If there's anything you'd like me to take a closer look at, feel free to let me know 🙇‍♀️

src/main/kotlin/ecommerce/entity/Cart.kt
saragrosser: Hi Dani, thank you for pointing this out! I've updated both `@ManyToOne` relationships in the Cart entity to use `FetchType.LAZY` as recommended. I also reviewed both Eager and Lazy and how they’re accessed and confirmed that lazy loading is safe in this case.

src/main/kotlin/ecommerce/entity/Member.kt
saragrosser: Thank you for this, I am still learning how to implement JPA, I don't understand because I thought this was made automatically 

I read that, JPA requires some entity to have a no-args constructor, because it uses reflection to instantiate objects.

-Leaving it public allows other parts of the code to create a “blank” member (e.g. Member()), which might violate domain rules.
-Making it private or protected limits that misuse — while still allowing JPA to use it.

So at the end I refactored and added the private constructor for member entity :)

src/main/kotlin/ecommerce/entity/Option.kt
saragrosser: Hi thank you for this, this was a mistake, because first I was trying to use JsonIgnore to break my infinite loop, but then I learned about the DTO approach and I decided to implement ProductResponse +  OptionResponse to solve it and I forgot to delete the JsonIgnore, but now i have deleted it since it is not needed anymore! sorry! :)

src/main/kotlin/ecommerce/entity/Product.kt
saragrosser: Yes that was also a mistake to duplicate the cascades I have studies now that of I use CascadeType.ALL I do not need to specify the type and if I don't use it then I would have to specify, but not both! I have refactor this now :)

src/main/kotlin/ecommerce/entity/Option.kt
saragrosser: Thanks for the suggestion! 

I updated the method to use `require()` and `check()` for input and state validation, I read this is best practice for readability much cleaner and more idiomatic!

src/main/kotlin/ecommerce/service/ProductService.kt
saragrosser: Thanks for pointing this out! I've refactored the create method to ensure the product is saved with at least one option, as required. Now, options are created and attached to the product before it's persisted, avoiding the previous issue where the product was briefly saved with an empty option list.

src/main/kotlin/ecommerce/service/ProductService.kt
saragrosser: Thank you! I added a validation step in update to ensure that option names are unique within the same product. This prevents duplicates and aligns with the requierements, thank you!

src/main/kotlin/ecommerce/entity/Member.kt
da-nyee: Good job 👍 

I'd also recommend adding a private no-args constructor to the other entities as well 😄

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,43 @@
...
+            "Quantity must be between 1 and 99,999,999."
+        }
+    }
+
+    fun subtract(amount: Int) {
+        require(amount > 0) { "Amount must be positive." }
+        check(quantity - amount >= 0) { "Not enough quantity." }
```
da-nyee, line 40: 👍

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,161 @@
...
+                )
+            }
+
+        val productWithOptions =
+            Product(
+                name = request.name,
+                price = request.price,
+                imageUrl = request.imageUrl,
+                options = options,
+            )
```
da-nyee, line 78: 👍

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,161 @@
...
+        val duplicateOptionNames =
+            request.options
+                .groupingBy { it.name }
+                .eachCount()
+                .filter { it.value > 1 }
+                .keys
+
+        if (duplicateOptionNames.isNotEmpty()) {
+            throw IllegalArgumentException("Duplicate option names are not allowed: ${duplicateOptionNames.joinToString()}")
+        }
```
da-nyee: It's looking good, but what about checking for the duplicate using the repository?
You might also need to consider existing data in the database 👀

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/api/products")
+class ProductRestController(
+    private val productService: ProductService,
+) {
+    @GetMapping
+    fun getAll(pageable: Pageable): ResponseEntity<Page<ProductResponse>> {
```
da-nyee: Does this API work well?
I think it should accept parameters like `page` and `limit` 🤔

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,73 @@
...
+        val page = productService.getAllPaginated(pageable)
```
da-nyee: Have you heard of `no-offset` pagination?
How is it different from the current approach?

da-nyee (changes requested):
Hi Saragrosser! 🙌 

I've left a few more comments — please take another look when you have time.

If you have any questions, just let me know ☺️

src/main/kotlin/ecommerce/service/ProductService.kt
saragrosser: Thank you for the helpful suggestion! I updated the duplication check in the update() method to consider both existing and incoming option names using the repository, so we avoid name conflicts with already persisted data. :)

src/main/kotlin/ecommerce/controller/ProductRestController.kt
saragrosser: Hi thank yoou! I have updated the controller to explicitly support page, size, sortBy, parameters

src/main/kotlin/ecommerce/controller/ProductRestController.kt
saragrosser: I’ve updated the controller to support page, size, sortBy, explicitly — so pagination and sorting are now more flexible and clearer.

I’m familiar with the idea of no-offset (or keyset) pagination, I know it can be more efficient and stable for large datasets, especially when dealing with frequently changing data.

For now, the offset approach seems to fit the needs of this project , but I’d definitely be up for exploring keyset pagination if we run into performance issues or decide to support infinite scrolling in the future.

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,89 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/api/products")
+class ProductRestController(
+    private val productService: ProductService,
+) {
+    @GetMapping
+    fun getAll(
+
```
da-nyee, line 30: This newline seems a bit redundant 👀

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,89 @@
...
+        @RequestParam(defaultValue = "0") page: Int,
+        @RequestParam(defaultValue = "10") size: Int,
+        @RequestParam(defaultValue = "id") sortBy: String,
+        @RequestParam(defaultValue = "asc") direction: String,
```
da-nyee, line 34: 👍

src/main/kotlin/ecommerce/controller/ProductRestController.kt
da-nyee: You nailed it 💯

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,155 @@
...
+        if (productRepository.existsByNameAndIdNot(request.name, id)) {
+            throw DuplicateProductNameException()
+        }
+
+        val existingOptionNames = optionRepository.findByProductId(id).map { it.name }
+        val newOptionNames = request.options.map { it.name }
+
+        val combinedOptionNames = existingOptionNames + newOptionNames
+
+        val duplicateOptionNames =
```
da-nyee, line 102: 👍

da-nyee (approved):
Hi Sari! 🙌

You're doing great so far 👍
Everything's looking good — it's ready to be merged.

See you in the next step! Keep it up 💪

PR #11, opened 2025-08-04
[spring-ecommerce-order - Step-1] Submit Mariia's mission

mariyakulikova:
🧩 Step 1 – Transition to JPA & ORM

This PR marks a significant step forward in refactoring our application from basic JDBC-style logic to a more expressive, maintainable, and testable domain model using JPA and Spring Data JPA.

🔍 Key Improvements
	•	Introduced DTOs (ProductRequest, OptionRequest) to decouple validation and input handling from entity logic.
	•	Moved business rules (e.g. unique option names) from the entity layer to the service layer to keep entities pure and reusable.
	•	Refactored ProductService for clarity and testability:
	•	Extracted reusable toResponse() methods.
	•	Made side effects explicit in create() and update() methods.
	•	Applied @Transactional where necessary for consistency.
	•	Improved entity design, removing validation from constructors and using cleaner patterns.
	•	Pagination and filtering now handled in a way that better supports scalability.

⚠️ Challenges We Faced
	•	We initially ran into infinite recursion due to bidirectional relationships and solved it using @JsonIgnore and better DTO usage.
	•	Ensured type safety in save()/saveAll() calls to avoid runtime issues.
	•	Rewrote and cleaned up several tests that were either too implicit or silently passing.
	•	Figured out how to use Mockito properly in Kotlin, especially with any() matchers and just runs.

💡 What We Learned
	•	Better understanding of JPA entity lifecycle, persistence context, and repository behavior.
	•	How to write modular code with clear responsibility boundaries between DTOs, entities, services, and repositories.
	•	Improved our test structure and strategy to handle future growth and complex flows.
	•	Hands-on experience with clean code principles in the context of a growing Spring Boot application.

Thank you for reviewing this PR 🙏
I'm happy to hear your suggestions and open to improvements!


README.md
```diff
@@ -1 +1,28 @@
...
+## Step 1.3 - Product Option
+
+### Features to Implement step 1.3
+
+- [x] A product must always have at least one option.
+- [x] Option names can include up to 50 characters , including spaces.
+- [x] Allowed special characters in option names.
+- [x] Option quantity must be at least 1 and less than 100,000,000.
+- [x] Duplicate option names are not allowed within the same product to prevent confusion during purchase.
+- [x] Implement a method to decrease the quantity of a product option by a specified amount.
```
Hyeon9mak, line 28: nice habit 💯 keep it up 🏃‍♂️

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
```diff
@@ -0,0 +1,38 @@
...
+        val member =
+            try {
+                memberService.findByToken(token)
+            } catch (e: Exception) {
+                return unauthorized(response)
+            }
+
+        if (request.requestURI.startsWith("/admin") && member?.role != "ADMIN") {
+            return unauthorized(response)
+        }
```
Hyeon9mak: What's different between HTTP status code 401 & 403?

src/main/kotlin/ecommerce/auth/JWTProvider.kt
```diff
@@ -0,0 +1,3 @@
+package ecommerce.auth
+
+class JWTProvider
```
Hyeon9mak: 😮

src/main/kotlin/ecommerce/dto/MemberRequest.kt
```diff
@@ -0,0 +1,14 @@
...
+import jakarta.validation.constraints.Size
+
+data class MemberRequest(
+    @field:NotBlank(message = "Email must not be blank")
+    @field:Email(message = "Invalid email format")
+    val email: String,
+    @field:NotBlank(message = "Password must not be blank")
+    @field:Size(min = 6, message = "Password must be at least 6 characters")
+    val password: String,
+)
```
Hyeon9mak: Request/Response DTOs and Domain Objects are used in the Presentation Layer and Domain Layer, respectively.
In your opinion, which layer is more appropriate for handling validation?

Of course, there's no absolute right answer here.
Performing validation in both layers is the safest approach, but it can also lead to redundant checks.
I'm curious to hear your thoughts. 😄

src/main/kotlin/ecommerce/entity/Cart.kt
```diff
@@ -0,0 +1,27 @@
...
+    @JoinColumn(name = "member_id", nullable = false)
+    val member: Member,
+    @ManyToOne
+    @JoinColumn(name = "product_id", nullable = false)
+    val product: Product,
+    @Column(name = "created_at")
+    val createdAt: LocalDateTime = LocalDateTime.now(),
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
```
Hyeon9mak, line 26: Good! null-safe ID.
How did JPA(Hibernate) know `0L` value is new instance? 😄 
hint : `AbstractEntityInformation.class`

src/main/kotlin/ecommerce/entity/Member.kt
```diff
@@ -0,0 +1,24 @@
...
+    val id: Long = 0L,
+    @Column(name = "name")
+    val name: String = "",
+    @Column(name = "email", unique = true, nullable = false)
+    val email: String,
+    @Column(name = "password", nullable = false)
+    val password: String,
+    @Column(name = "role")
+    val role: String = "USER",
+)
```
Hyeon9mak: role is not a just String. i think it can convert to enum class!

src/main/kotlin/ecommerce/entity/Product.kt
```diff
@@ -0,0 +1,38 @@
...
+    val price: Double,
+    @Column(name = "image_url")
+    val imageUrl: String,
+    options: List<Option>,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.PERSIST, CascadeType.ALL])
+    val options: MutableList<Option> = options.toMutableList()
```
Hyeon9mak: What's different between Cascade PERSIST & ALL ?

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+    @Column(nullable = false)
+    var quantity: Int,
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "product_id", nullable = false)
+    @JsonIgnore
+    var product: Product? = null,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
```
Hyeon9mak: What was ur reasoning for choosing FetchType.LAZY?

For example, when displaying CartItem on a shopping page, it would be natural for the Product information to be shown together. In that case, EAGER might actually be a better fit than LAZY. 

How about considering fetch strategies from a business perspective like this? 😄 
(It's why Developers need to understand the Business, not just Tech.)

Hyeon9mak: What was ur reasoning for choosing `@JsonIgnore`? 👀

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,58 @@
...
+    @ExceptionHandler(DuplicateProductNameException::class)
+    fun handleDuplicateProductName(ex: DuplicateProductNameException): ResponseEntity<ErrorResponse> {
+        val error =
+            ErrorResponse(
+                message = "Validation failed",
+                errors = listOf(FieldError(ex.field, ex.message ?: "Invalid value")),
+            )
+        return ResponseEntity.status(HttpStatus.BAD_REQUEST).body(error)
+    }
+}
```
Hyeon9mak: What would happen if an exception occurs that we didn't explicitly handle?
Hint: inheritance exception.

src/main/kotlin/ecommerce/repository/ProductJpaRepository.kt
```diff
@@ -0,0 +1,19 @@
...
+interface ProductJpaRepository : JpaRepository<Product, Long> {
+    fun existsByName(name: String): Boolean
+
+    fun existsByNameAndIdNot(
+        name: String,
+        id: Long,
+    ): Boolean
+
+    override fun findAll(pageable: Pageable): Page<Product>
+}
```
Hyeon9mak: Do we really need `@Repository` annotation? 🤔 
Hint: `SimpleJpaRepository.class`

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun getCartItems(memberId: Long): List<Cart> {
+        return cartRepository.findByMemberId(memberId)
+    }
+
+    fun getCartItemsPaginated(
+        memberId: Long,
+        pageable: Pageable,
+    ): Page<Cart> {
+        return cartRepository.findByMemberId(memberId, pageable)
+    }
```
Hyeon9mak: What's different between `@Transactional` & `@Transactional(readOnly = true)`

src/main/resources/application.properties
```diff
@@ -1 +1,15 @@
...
+
+spring.sql.init.mode=always
+
+security.jwt.token.secret-key=hKXW4Bel2pw0AxqD30Q02cyOh+24kZbJ73IffoIJBOw=
+security.jwt.token.expire-length=3600000
+
+
+spring.jpa.hibernate.ddl-auto=create
+spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.format_sql=true
```
Hyeon9mak: https://medium.com/@hyeon9mak/about-no-newline-at-end-of-file-warning-on-github-b42d14dccc93 😉

Hyeon9mak (changes requested):
Hello Mariia, Ur spring-ecommerce-order mission reviewer here!
Did a good job this mission. 👍 

I've left some comments about validation layer, exception handle and JPA.
Let’s take a look at these first, and then we can talk more about testing in detail. 😄 

Feel free to ask anytime!

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
mariyakulikova: `401` is used when the user is not logged in or the token is invalid.
`403` is used when the user is logged in but doesn’t have permission (like not an admin).
I’ll check if we need to return different codes here.

src/main/resources/application.properties
mariyakulikova: Thanks for the article! I’ve set up my editor now, so it should add the newline automatically. 😊

src/main/kotlin/ecommerce/service/CartService.kt
mariyakulikova: `@Transactional` creates a transaction that can read and write.
It also checks if something changed and updates the database at the end (this is called dirty checking).
`@Transactional(readOnly = true)` is for read-only methods.
It doesn’t do dirty checking, so it’s faster and safer when we only need to read data.

src/main/kotlin/ecommerce/repository/ProductJpaRepository.kt
mariyakulikova: Thanks for the hint! 😊
Since this interface extends `JpaRepository`, Spring Data automatically detects it and registers it as a Spring bean using a proxy (like `SimpleJpaRepository`).
I’ll remove it to keep the code clean.

src/main/kotlin/ecommerce/entity/Member.kt
mariyakulikova: I replaced the role String with an enum class Role and used `@Enumerated(EnumType.STRING)` to store it as a string in the database.
This makes the code safer and avoids typos.

src/main/kotlin/ecommerce/entity/Cart.kt
mariyakulikova: Since the ID is a primitive Long, it’s never null, but Spring Data checks if the value is `0L` to treat it as a new entity.
I saw that `AbstractEntityInformation` (and related classes like `JpaMetamodelEntityInformation`) treat 0 as “not persisted”.
So `0L` works like a null-safe marker for a new entity.

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
mariyakulikova: `401` means the user is not authenticated (like missing or invalid token).
`403` means the user is authenticated but not allowed to access (like not an admin).
In my case, I plan to return 401 when token is invalid, and 403 when role is not admin.

src/main/kotlin/ecommerce/dto/MemberRequest.kt
mariyakulikova: I think simple validation like `@NotBlank` or `@Email` makes sense in the DTO, because it’s close to the input and gives quick feedback to the client.
But business rules (like “email must be unique”) should be checked in the domain layer or service.
So I prefer using both, but avoid duplicated logic.

src/main/kotlin/ecommerce/entity/Option.kt
mariyakulikova: I didn’t take the business requirements into account when choosing `FetchType.LAZY`.
In this case, since we usually need the Product data together with the Option, `EAGER` fetching makes more sense. I’ll update it accordingly.
As for the `@JsonIgnore`, we initially added it to avoid an infinite recursion issue during serialization. It’s no longer necessary, and I’ll remove it. 😊

Thanks for pointing this out!

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
mariyakulikova: Thanks for the hint! 😊
If an exception occurs that we don’t explicitly handle, the application might:
- crash with a 500 Internal Server Error, or
- return a raw or unclear error message that’s shown directly to the user.

I refactored the code by introducing a base `ApplicationException` class and made all custom exceptions extend it.

src/main/kotlin/ecommerce/entity/Product.kt
mariyakulikova: You’re right to point this out. There’s actually no need to include both `PERSIST` and `ALL` together.
`CascadeType.ALL` already includes `PERSIST`, `MERGE`, `REMOVE`, `REFRESH`, and `DETACH`.
So having both `PERSIST` and `ALL` is redundant.
I’ll update the annotation to use only `CascadeType.ALL` and refactor the service code to rely on cascading instead of calling `optionRepository.saveAll()` manually.

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
```diff
@@ -0,0 +1,41 @@
...
+                response.sendError(HttpServletResponse.SC_UNAUTHORIZED)
+                return false
+            }
+
+        if (request.requestURI.startsWith("/admin") && member?.role != Role.ADMIN) {
+            response.sendError(HttpServletResponse.SC_FORBIDDEN)
+            return false
+        }
+
+        return true
```
Hyeon9mak, line 34: 😉 👍

src/main/kotlin/ecommerce/dto/MemberRequest.kt
Hyeon9mak: niiiice thought 👍

src/main/kotlin/ecommerce/entity/Role.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.entity
+
+enum class Role {
+    USER,
+    ADMIN,
+}
```
Hyeon9mak, line 6: > used `@Enumerated(EnumType.STRING)` to store it as a string in the database.
This makes the code safer and avoids typos.

💯 👍

src/main/kotlin/ecommerce/entity/Product.kt
```diff
@@ -0,0 +1,38 @@
...
+    @Column(name = "image_url")
+    val imageUrl: String,
+    options: List<Option>,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.ALL])
+    val options: MutableList<Option> = options.toMutableList()
+
```
Hyeon9mak: Question Event~~
What's different between `CascadeType.REMOVE` and `orphanremoval = true` in `@OneToMany` annoation?

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,43 @@
...
+@Table(name = "product_option")
+class Option(
+    @Column(name = "name", nullable = false, length = 50)
+    var name: String,
+    @Column(nullable = false)
+    var quantity: Int,
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "product_id", nullable = false)
+    var product: Product? = null,
+    @Id
```
Hyeon9mak: `product_option` most likely only exists when there's an associated `product`.
So I think there no need null. What do you think?

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @ExceptionHandler(RuntimeException::class)
+    fun handleUnexpectedException(ex: RuntimeException): ResponseEntity<ErrorResponse> {
+        val error =
+            ErrorResponse(
+                message = "Internal server error",
+                errors = listOf(FieldError("unknown", ex.message ?: "Something went wrong")),
+            )
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(error)
+    }
```
Hyeon9mak: Taking it a step further, since these are unexpected errors, there's a risk that sensitive server-related information could be exposed to the user. That's why I prefer returning a general guidance message in the response body rather than detailed information. 😄

src/main/kotlin/ecommerce/controller/CartAdminController.kt
```diff
@@ -0,0 +1,33 @@
...
+class CartAdminController(
+    private val cartService: CartService,
+) {
+    @GetMapping("/top-products")
+    fun getTopProducts(
+        @LoginMember member: Member,
+    ): ResponseEntity<List<ProductStatResponse>> {
+        val result = cartService.getTop5MostAddedProducts()
+        return ResponseEntity.ok(result)
+    }
```
Hyeon9mak: Currently, the Member entity is also being used as the authenticated user model. This means that changes to the Member RDB table in the future could impact the authentication logic, and vice versa — changes in authentication could affect the Member entity.

This suggests that Member currently holds more than one responsibility.
In fact, after a user logs in, there are quite a few properties in Member that aren't really needed.

How about creating a separate object specifically for authentication purposes?

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,62 @@
...
+        return cartRepository.findByMemberId(memberId)
+    }
+
+    @Transactional(readOnly = true)
+    fun getCartItemsPaginated(
+        memberId: Long,
+        pageable: Pageable,
+    ): Page<Cart> {
+        return cartRepository.findByMemberId(memberId, pageable)
+    }
```
Hyeon9mak: not used! 👀

src/main/kotlin/ecommerce/service/MemberService.kt
```diff
@@ -0,0 +1,50 @@
...
+        val member =
+            memberRepository.findByEmail(request.email)
+                ?: throw IllegalArgumentException("Invalid email or password")
+
+        if (member.password != request.password) {
+            throw IllegalArgumentException("Invalid email or password")
+        }
+
+        val token = jwtTokenProvider.createToken(member.id)
+        return TokenMemberResponse(token)
```
Hyeon9mak: ask to object! then it makes more easier to generate test code.

```kotlin
// throw exception in itself.
member.validatePassword(password = password)
```

src/test/kotlin/ecommerce/entity/OptionTest.kt
```diff
@@ -0,0 +1,42 @@
...
+            imageUrl = "http://test.com",
+            options = emptyList(),
+        )
+
+    @Test
+    fun `throw error if name length is longer than 50 letters`() {
+        assertThrows<IllegalArgumentException> {
+            Option("a".repeat(51), 1, dummyProduct)
+        }
+    }
```
Hyeon9mak: niiiiiiiice 👍 
how do we more easier generate test code?
the answer is create `value class` 😉

Hyeon9mak (changes requested):
Hi Mariia! good work to reflect feedbacks.

I've left more deeper comments in this time,
Please take your time to review, and feel free to ask if you have any questions 😄

src/main/kotlin/ecommerce/entity/Option.kt
mariyakulikova: By adding `optional = false` and `nullable = false`, plus making the Kotlin property non-null via `lateinit`, we now guarantee at both the DB and code level that an Option always has a valid Product. Let me know if that looks good or if you’d prefer the constructor-injection style instead.

src/main/kotlin/ecommerce/entity/Product.kt
mariyakulikova: They actually control two different behaviors:
- **cascade = CascadeType.REMOVE**
When we call remove(product), JPA will automatically issue DELETE for all its options. It only fires on an explicit remove of the parent.
- **orphanRemoval = true**
When we remove an Option from the product.options list, JPA treats that detached child as an orphan and issues DELETE for it, even though you never called remove(product).

In our case, since Option can not exist without Product, I defined both behaviors.

src/test/kotlin/ecommerce/entity/OptionTest.kt
mariyakulikova: Thanks! 🙌 I added OptionName and ProductName as a value class, it really helped simplify the validation and made the test more focused. Let me know if you see any improvements I could make to the naming or structure!

src/main/kotlin/ecommerce/service/MemberService.kt
mariyakulikova: Thanks! Moved the password validation into a method in Member as suggested. Also I moved to Member other validation logic from DTO.

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
mariyakulikova: Thanks for pointing that out! I updated the handler to return a general error message instead of exposing `ex.message`, and I also added logging to make sure we still capture the exception details internally, since the next step is deployment.

src/main/kotlin/ecommerce/controller/CartAdminController.kt
mariyakulikova: That makes sense! I extracted a dedicated class for the authenticated user and updated the argument resolver to return that instead of the full Member.

src/main/kotlin/ecommerce/entity/Option.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    @ManyToOne(fetch = FetchType.EAGER, optional = false)
+    @JoinColumn(name = "product_id", nullable = false)
+    lateinit var product: Product
+
+    init {
+        require(quantity in 1 until 100_000_000) {
+            "Quantity must be between 1 and 99,999,999."
+        }
+    }
```
Hyeon9mak, line 34: > Let me know if that looks good or if you’d prefer the constructor-injection style instead.

This is just my personal take, but because a product is usually created first and then options are added afterward, I feel that using constructor injection might be a more natural fit. What do you think?

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    @ExceptionHandler(RuntimeException::class)
+    fun handleUnexpectedException(ex: RuntimeException): ResponseEntity<ErrorResponse> {
+        logger.error("Unexpected error occurred", ex)
+        val error =
+            ErrorResponse(
+                message = "Internal server error",
+                errors = listOf(FieldError("unknown","Something went wrong")),
+            )
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(error)
```
Hyeon9mak, line 52: >  I also added logging to make sure we still capture the exception details internally, since the next step is deployment.

💯 👍

src/main/kotlin/ecommerce/dto/AuthenticatedUser.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+import ecommerce.entity.Role
+
+class AuthenticatedUser(
+    val role: Role
+)
```
Hyeon9mak, line 7: 😄 👍

src/test/kotlin/ecommerce/entity/MemberTest.kt
```diff
@@ -0,0 +1,68 @@
...
+        val member =
+            Member(
+                email = "user@example.com",
+                password = "securePass123",
+            )
+
+        assertThrows<IllegalArgumentException> {
+            member.validatePassword("securePass")
+        }
+    }
```
Hyeon9mak, line 67: niiiice 💯

src/test/kotlin/ecommerce/entity/OptionNameTest.kt
```diff
@@ -0,0 +1,42 @@
...
+            OptionName("Cheese!")
+        }
+    }
+
+    @Test
+    fun `should be equal if values are the same`() {
+        val name1 = OptionName("Extra Cheese")
+        val name2 = OptionName("Extra Cheese")
+        assertEquals(name1, name2)
+    }
```
Hyeon9mak, line 41: Of course, adding more value classes would be even better, but I think what you've learned so far is already enough for this mission. 😄

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,145 @@
...
+                id = existingProduct.id,
+                name = request.name,
+                price = request.price,
+                imageUrl = request.imageUrl,
+                options = updatedOptions,
+            )
+
+        productRepository.save(updatedProduct)
+        return updatedProduct.toResponse()
+    }
```
Hyeon9mak, line 122: In JPA, when you modify fields of an `@Entity` (other than the ID) within a transaction, the changes are automatically updated (saved) when the transaction ends. This feature is called dirty checking, and it can be quite useful!
(That said, I personally prefer calling save() explicitly — I find it more clear and intentional. 😄 )

Hyeon9mak (approved):
Hi Mariia, great job quickly incorporating the feedback! 👍 
I've left some easssy comments.

You're now ready to move on to the final stage.
See you in the next step!

PR #13, opened 2025-08-04
[spring-ecommerce-order - Step-1] Submit knopgm's mission

knopgm:
# Spring-ecommerce-order Step-1
Implement Step 1 Features: JPA Integration, Pagination, Wishlist:(Cart), and Product Options

### 1. Step 1.1: JPA Integration
- Replaced JdbcTemplate with Spring Data JPA for database operations.
- Updated entities and services to use JPA.
- Added logging for JPA operations to improve traceability.
### 2. Step 1.2: Pagination and Wishlist
- Added pagination support for products.
- Implemented wishlist functionality.
- Added pagination support for wishlist items.
### 3. Step 1.3: Product Options
- Created the Option entity with the following constraints:
- Name limited to 50 characters.
- Quantity must be between 1 and 99,999,999.
- Prevented duplicate option names within the same product.
- Created subtract method in Option.

## Questions
At this project we had applied every new feature following the logic thread that was related to it and, at the end, we still had some Jdbc code mixed with our JPA implementation. 
The problem is that we tried to implement a Cart to hold the member id and a cart Item that holds the product id  and the cart id, and when we finally tried to migrate them as a JPA repository, we couldn't figured out how to join both information (member id and product id to add items at the cart and remove them too), so part of the project still with Jdbc library.
We have some suggestions from coaches and colleagues but, because of the time constrains, we decided to pull like it is and improve after your code analyses. 

I'm looking foward to your review! Thank you for that! 🙏 🙏


src/main/kotlin/ecommerce/entity/CartEntity.kt
```diff
@@ -0,0 +1,20 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.time.LocalDateTime
+
+@Entity
+@Table(name = "cart")
+open class CartEntity(
```
da-nyee:
```suggestion
open class Cart(
```

Just a tiny suggestion — `-Entity` suffix might be a bit redundant.

src/main/kotlin/ecommerce/entity/CartEntity.kt
```diff
@@ -0,0 +1,20 @@
...
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long? = null,
+    @Column(name = "member_id", nullable = false)
+    val memberId: Long,
+    @Column(name = "created_at", nullable = false)
+    val createdAt: LocalDateTime = LocalDateTime.now(),
```
da-nyee, line 19:
```suggestion
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    val id: Long? = null,
    
    @Column(name = "member_id", nullable = false)
    val memberId: Long,
    
    @Column(name = "created_at", nullable = false)
    val createdAt: LocalDateTime = LocalDateTime.now(),
```

Adding some newlines could help improve readability 👀

src/main/kotlin/ecommerce/entity/MemberEntity.kt
```diff
@@ -0,0 +1,38 @@
...
+    val email: String,
+    @Column(nullable = false)
+    val password: String,
+    @Column(nullable = false)
+    val role: String,
+    @Column(nullable = false)
+    var name: String,
+    @OneToMany(fetch = FetchType.LAZY)
+    @JoinColumn(name = "product_id", nullable = true)
+    val products: MutableList<ProductEntity> = mutableListOf(),
```
da-nyee: It might be better to make `Member` and `Product` independent. What do you think?

src/main/kotlin/ecommerce/entity/ProductEntity.kt
```diff
@@ -0,0 +1,48 @@
...
+    val id: Long? = null,
+    @Column(nullable = false, unique = true)
+    var name: String,
+    @Column(nullable = false)
+    val price: Double,
+    @Column(nullable = false, name = "image_url")
+    val imageUrl: String,
+    @OneToMany(
+        cascade = [CascadeType.MERGE, CascadeType.PERSIST],
+        orphanRemoval = true,
```
da-nyee, line 25: How does this option behave? I'd like to hear your take on it 👀

src/main/kotlin/ecommerce/entity/OptionEntity.kt
```diff
@@ -0,0 +1,38 @@
...
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+
+@Entity
+@Table(
+    name = "option",
+)
+open class OptionEntity(
```
da-nyee: Just curious — does an entity need to be an `open class`?

src/main/kotlin/ecommerce/entity/CartItemEntity.kt
```diff
@@ -0,0 +1,33 @@
...
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.OneToOne
+import jakarta.persistence.Table
+import jakarta.validation.constraints.Positive
+import java.time.LocalDateTime
+
+@Entity
+@Table(name = "cart_item")
+open class CartItemEntity(
```
da-nyee: What about adding a private no-args constructor?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,53 @@
...
+        return productRepositoryJpa.save(product)
+    }
+
+    fun getAllProducts(
+        page: Int,
+        size: Int,
+        sortBy: String = "name",
+        direction: Sort.Direction = Sort.Direction.ASC,
+    ): Page<ProductEntity> {
+        val pageable = PageRequest.of(page, size, Sort.by(direction, sortBy))
```
da-nyee, line 78: Pagination 👍

da-nyee: But there's a more efficient way to handle pagination.
Have you heard of `no-offset` pagination? If so, how is it different from this approach?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,53 @@
...
+) {
+    fun createProduct(productRequest: ProductRequest): ProductEntity {
+        if (productRepositoryJpa.existsByName(productRequest.name)) {
+            throw IllegalArgumentException("Product with name '${productRequest.name}' already exists.")
+        }
+        val options: MutableList<OptionEntity> =
+            mutableListOf(
+                OptionEntity(name = "Blue XL", quantity = 99),
+                OptionEntity(name = "Red Large", quantity = 42),
+            )
```
da-nyee: Would it be okay to hardcode the options?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.stereotype.Service
+
+@Service
+class ProductService(
+    private val productRepositoryJpa: ProductRepositoryJpa,
+) {
+    fun createProduct(productRequest: ProductRequest): ProductEntity {
+        if (productRepositoryJpa.existsByName(productRequest.name)) {
+            throw IllegalArgumentException("Product with name '${productRequest.name}' already exists.")
+        }
```
da-nyee, line 19: You did check for duplicate product names 👍

da-nyee: But according to the requirement:

image

<br/>
<br/>

It looks like you might need to check one more thing.

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,53 @@
...
+import ecommerce.entity.OptionEntity
+import ecommerce.entity.ProductEntity
+import ecommerce.repository.ProductRepositoryJpa
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.PageRequest
+import org.springframework.data.domain.Sort
+import org.springframework.stereotype.Service
+
+@Service
+class ProductService(
```
da-nyee, line 13: Maybe consider adding unit tests for the services 🤔

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,67 @@
...
+    private val cartRepository: CartRepository,
+    private val productRepositoryJpa: ProductRepositoryJpa,
+    private val cartItemRepositoryJpa: CartItemRepositoryJpa,
+) {
+    fun addToCart(
+        memberId: Long,
+        productId: Long,
+    ) {
+        productRepositoryJpa.findById(productId)
+            ?: throw NoSuchElementException("Product not found")
```
da-nyee: This one seems not used 👀

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,67 @@
...
+        cartRepository.add(memberId, productId)
+    }
```
da-nyee: > At this project we had applied every new feature following the logic thread that was related to it and, at the end, we still had some Jdbc code mixed with our JPA implementation.
The problem is that we tried to implement a Cart to hold the member id and a cart Item that holds the product id and the cart id, and when we finally tried to migrate them as a JPA repository, we couldn't figured out how to join both information (member id and product id to add items at the cart and remove them too), so part of the project still with Jdbc library.
We have some suggestions from coaches and colleagues but, because of the time constrains, we decided to pull like it is and improve after your code analyses.

Maybe you're referring to this part of the code, right?

Well, I sometimes use native queries to handle certain logic as well.
So using JDBC can be a good choice in some cases.

But in this case, I feel like if the `Cart` entity held the `productId` directly, it might be easier to manage 🤔

da-nyee (changes requested):
Hi Knopgm! 🙌
Sorry for the late review 🥺

I went through the code based on the requirements — so far, so good 👍
If you have any other questions, feel free to reach out 🙇‍♀️

src/main/kotlin/ecommerce/entity/CartEntity.kt
knopgm: Agreed! I would love to have line between some pieces of code, but our coach taught us to, from time to time, format the code using ktlinFormat (at the Gradle menu on the left side of IntelliJ), and this tool aim to clean the code even from empty lines 😐

src/main/kotlin/ecommerce/entity/ProductEntity.kt
knopgm: This piece of code gives us more control about the child elements related to their parent. 
`CascadeType.PERSIST` allows us to automatically save/update the child of `ProductEntity`, and the `CascadeType.Merge` is the same automation, but for merging action.
`orphanRemoval` is to ensure that by removing a child from the collection, it will be automatically deleted.

Is it ok to have this code all together? I just made a research and I think you are trying to tell me about the combo of the default Lazy-load behavior with the orphanRemoval. Am I right? If yes, which is the best approach instead of orphanRemoval? maybe Manually removing it? 👀

src/main/kotlin/ecommerce/entity/CartItemEntity.kt
knopgm: Interesting question! Thank you for pointing it to me! 🙏 
In my research I found out that, by installing JPA, we are automatically using Hibernate, which is a object-relational mapping that automatically makes entitys and their properties open.
I refactored all entity classes!

src/main/kotlin/ecommerce/entity/OptionEntity.kt
knopgm: Interesting question! Thank you for pointing it to me! 🙏 
In my research I found out that, by installing JPA, we are automatically using Hibernate, which is a object-relational mapping that automatically makes entitys and their properties open.
I refactored all entity classes!

src/main/kotlin/ecommerce/service/ProductService.kt
knopgm: Thank you for catching this one! 😅

src/main/kotlin/ecommerce/service/CartService.kt
knopgm: are you referring to the NoSuchElementException?
if yes, we are using it in the GlobalExceptionHandler.kt file.

src/main/kotlin/ecommerce/service/ProductService.kt
knopgm: Also interesting subject! The no-offset is how we manage the database content responses to be display in a UI as a infinite scroll.
This method has the advantage of:
- It is fast, because it doesn’t uses index efficiently
- it's performance is constant even with large datasets which allows scalability
- it is less affected by inserts/deletes between page loads
- is easy to hide internal data like row numbers
Can I leave this feature as a plus for the next mission? Or is it crucial to have a infinite scroll management for this mission?

src/main/kotlin/ecommerce/service/ProductService.kt
knopgm: Good Catch!
I implement the validation for unique option name and, by doing it, I realized that my ProductController wasn't separating concerns.
I also fix the separation of concerns, then. 😊

src/main/kotlin/ecommerce/entity/CartEntity.kt
knopgm: Since I had to create Member and Product Domains as their name+Entity (to be different from the Member and Product who still using Jdbc), I thought it should follow the pattern for the new Entity too. 
I'm planning to migrate all ecommerce application to JPA, but maybe it could take me too long to fix everything around this migration. 
Since you said it is not a problem to mix both (Jdbc and JPA) I'll leave this improvement for a future date.
Let me know if you think I need to change anything related to this migration now with the goal not to affect the next missions. Thank you so, so much! 🙏

src/main/kotlin/ecommerce/entity/CartItemEntity.kt
knopgm: Interesting! So JPA will be allowed to instantiate each entity by reflection. 
Another great lesson from this review! 😄

src/main/kotlin/ecommerce/service/CartService.kt
knopgm: hmm...
I'm not sure what to do here.
I really don't want to change all my code now and, as far as I could test it, everything is working fine with the Cart holding memberId and the CartItem holding the productId...
Do you mean that by adding the productId to the Cart it is going to be easier to implement the last step of the mission? 👀 👀

src/test/kotlin/ecommerce/service/ProductServiceTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.jupiter.api.Assertions.assertThrows
+import org.junit.jupiter.api.Assertions.assertTrue
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.transaction.annotation.Transactional
+
+@SpringBootTest
+@Transactional // Ensures DB is rolled back after each test
+class ProductServiceTest
```
da-nyee, line 18: 👍

src/main/kotlin/ecommerce/service/ProductService.kt
da-nyee: Oh, you don't need to implement the infinite scroll!
I just wanted you to know there's another type of pagination ☺️

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,99 @@
...
+                .orElseThrow { NoSuchElementException("Product with id $id not found") }
+
+        // Check for name uniqueness if name is being changed
+        if (product.name != request.name && productRepositoryJpa.existsByName(request.name)) {
+            throw IllegalArgumentException("Product with name '${request.name}' already exists.")
+        }
+
+        product.options.clear()
+        // this will execute the deletes before adding new ones
+        productRepositoryJpa.flush()
```
da-nyee, line 54: It might be better to validate duplicates rather than remove all the data and add new ones 👀

src/main/kotlin/ecommerce/entity/ProductEntity.kt
da-nyee: Well, `orphanRemoval` could lead to unexpected results.
I recommend not using it and instead writing the removal code explicitly 🤔

src/main/kotlin/ecommerce/service/CartService.kt
da-nyee: I thought there was no variable declared in that code, but now there's one!

src/main/kotlin/ecommerce/service/CartService.kt
da-nyee: Well, I reconsidered it, and I think it's okay to leave all the entities as they are for now.

da-nyee (approved):
Hi Gabi! 🙌

Thanks for applying the feedback and leaving some comments 🙇‍♀️
I've gone through all the changes and comments — looks good to merge!

Let's move on to the next step 🤸‍♀️

src/main/kotlin/ecommerce/service/ProductService.kt
knopgm: Agreed! It will take quite a long time to refactor everything to work like this.
I'll add this refactor improvement at the end of the mission, if the limited time allows. 🤞 😅

src/main/kotlin/ecommerce/entity/ProductEntity.kt
knopgm: Refactored!

PR #14, opened 2025-08-06
[Spring-Ecommerce-Order Step 1] Submit Petra's Mission

pebencze:
Hello PK!  

Hope you are doing well and thank you for reviewing my code! :)  

This is a small **e-commerce** HTTP API built using **Spring Boot**, where users can have _carts_ with _cart-items_. Cart items are _products_ with _options_ that an _admin_ can edit (add, delete or update). It is based on and extends this project: [spring-ecommerce-product](https://github.com/herotechcourse/spring-ecommerce-product).

---
Here is an overview of what [Mina](https://github.com/miooo0o/miooo0o) and I have implemented:

#### Step 1.1 - Product API
- refactor the existing codebase from **JdbcTemplate** -> **Spring Data JPA** 
- create Entities
- rerfactor Repositories using interfaces
- redesign relationships between Entities
    - Cart -> Member 
    - CartItem -> Product
    - CartItem <-> Cart
    - Option <-> Product

This was the most challenging part and we have used interfaces for ProductRepository and MemberRepository, but we had to completely refactor CartRepository and its Service and Controller layers.

#### Step 1.2 - Pagination
- introduce paginated responses using **Page<T>** using the **Pageable** object
- full pagination with total count, total pages, current page
- implemented for **ProductController** and **CartController**

#### Step 1.3 - Product Option
- add options to product information
- design and implement the feature using a **bidirectional** relationship between the Product and **Option** models

---

I would appreciate some feedback on the implementation. Difficult topics were/are:
- If I don't use Cascade.ALL, what can I use instead?
- What exactly is lazy and eager fetching and the Transactional annotation?

Thank you so much!

Best,

Petra


README.md
```diff
@@ -1 +1,94 @@
...
+
+## Considerations
+
+- [x] remove Boolean return type from all delete methods
+- [ ] change Double to BigDecimal inside the Entity (Product/price)
+- [x] decide on where and how to use Models
+    - Entity == Model
+- [ ] dont use Cascade.All but Persist, Merge etc.
+- [ ] move product-option mapping to the constructor
+- [ ] effective test: Create InMemory fake repos -> service uses the fake repos
```
pkeugine, line 94: Nice considerations!
I'm curious, are you going to implement these?

README.md
```diff
@@ -1 +1,94 @@
...
+#### Test
+- [x] test Option
+
+## Considerations
+
+- [x] remove Boolean return type from all delete methods
+- [ ] change Double to BigDecimal inside the Entity (Product/price)
+- [x] decide on where and how to use Models
+    - Entity == Model
+- [ ] dont use Cascade.All but Persist, Merge etc.
```
pkeugine, line 92: **Q. If I don't use Cascade.ALL, what can I use instead?**

I'm curious about why you think we shouldn't use `Cascade.ALL`!
Is it something you want to avoid in your entity relationship?

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,57 @@
...
+import ecommerce.dto.TokenResponse
+import ecommerce.exception.ConflictException
+import ecommerce.exception.ForbiddenException
+import ecommerce.exception.NotFoundException
+import ecommerce.exception.UnauthorizedException
+import ecommerce.infrastructure.JwtTokenProvider
+import ecommerce.model.Member
+import ecommerce.repository.MemberRepository
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
```
pkeugine, line 16: **Q. What exactly is lazy and eager fetching and the Transactional annotation?**

I think it would be great for you to explain it to me, and I'll tell you if the direction is wrong.
I say this because this topic is actually very important in our development environment,
and it's a regular question in tech interviews. Might as well study it now and prepare for those situations early!

pkeugine: My tip for studying `@Transactional` annotation :  
Look at it's implementation. You'll be able to see bunch of options what we can use.
Those options are all related to how a transaction gets created and how it's related to ACID, which is an acronym for transaction's property.

pkeugine: Also, if you have the time, it would be great to make a small api that uses `@Transactional` annotation, and do all sorts of things inside it. (ex: throw exception after saving an entity, update an entity after saving it and see the database connection logs... etc)

By trying out those experiments and debugging the process, you'll be able to experience `@Transactional` and how persistence context in JPA works. I remember my first time trying it out, and it really felt like some kind of magic..! (And it was also hard and confusing 😂 )

src/main/kotlin/ecommerce/config/DatabaseFixture.kt
```diff
@@ -0,0 +1,158 @@
+package ecommerce.config
+
+import ecommerce.dto.Role
+import ecommerce.model.Member
+import ecommerce.model.Option
+import ecommerce.model.Product
+
+object DatabaseFixture {
```
pkeugine: Since these fixtures are used in tests, why don't we manage these in test package? :)

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,71 @@
...
+        @LoginMember member: RegisteredMember,
+        @RequestParam(defaultValue = "0") page: Int,
+        @RequestParam(defaultValue = "10") size: Int,
+    ): ResponseEntity<Page<CartItemResponse>> {
+        val itemPages = cartService.getPages(member.id, page, size)
+        val headers =
+            HttpHeaders().apply {
+                add("X-Page-Number", itemPages.number.toString())
+                add("X-Page-Size", itemPages.size.toString())
+            }
```
pkeugine: It's interesting that you've put `X-Page-Number` and `X-Page-Size` headers here!
I'm curious about the thought process behind this, and if there's any reference for this way of paging!

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @DeleteMapping
+    fun deleteItem(
+        @LoginMember member: RegisteredMember,
+        @RequestBody request: CartItemRequest,
+    ): ResponseEntity<CartItemResponse?> {
+        cartService.deleteItem(member.id, request)
+        return ResponseEntity.noContent().build()
+    }
```
pkeugine:
```suggestion
    ): ResponseEntity<Void> {
        cartService.deleteItem(member.id, request)
        return ResponseEntity.noContent().build()
    }
```

Since this api doesn't add anything to it's response body, I think it would be better for us to reflect that in our return type.

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,71 @@
...
+        return ResponseEntity.status(HttpStatus.OK).body(body)
+    }
+
+    @GetMapping("/wishlist")
+    fun getAllItemsAsPages(
+        @LoginMember member: RegisteredMember,
+        @RequestParam(defaultValue = "0") page: Int,
+        @RequestParam(defaultValue = "10") size: Int,
+    ): ResponseEntity<Page<CartItemResponse>> {
+        val itemPages = cartService.getPages(member.id, page, size)
```
pkeugine: Since `Page` and `PageImpl` are both from Spring Data JPA, why don't we try our best to keep their usage scope be as small as possible?

For example, this `CartController` is now depending on `Page` interface, so if we wanted to change our persistence layer library, this controller also needs to be changed. By keeping the scope as small as possible, we'll be able to let this controller not care about which technology we us in persistence layer. How do you think?

src/main/kotlin/ecommerce/model/mapper/Mapper.kt
```diff
@@ -0,0 +1,39 @@
+package ecommerce.model.mapper
+
+import ecommerce.dto.CartItemResponse
+import ecommerce.dto.OptionResponse
+import ecommerce.dto.ProductResponse
+import ecommerce.model.CartItem
+import ecommerce.model.Option
+import ecommerce.model.Product
+
+object CartItemMapper {
```
pkeugine: Nice mapper class! 👍
I’m curious about why you chose to have this object depend on controller-related DTOs.

In a typical layered architecture, dependencies usually flow downward like this:
**controller / view / controller-related DTO** → **domain / service** → **repository**

In our project, the models act as domains. That means this mapper introduces a dependency from the domain layer up toward controller-related DTOs, which is the reverse of what we usually aim for.

This could make it harder to reuse these models in other contexts. For example, if we wanted to build a new project or module that uses these models but doesn’t have the same controller package.

What do you think about this?

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import java.time.LocalDateTime
+
+@Repository
+interface CartItemRepository : JpaRepository<CartItem, Long> {
+    fun findDistinctByCreatedAtAfter(dateThreshold: LocalDateTime): List<CartItem>
+
+    fun findDistinctByUpdatedAtAfter(dateThreshold: LocalDateTime): List<CartItem>
+}
```
pkeugine, line 13: When I first used Spring Data JPA, this really confused me 😂 
How does our `CartItemRepository` work when we don't even write codes about `findDistinctByCreatedAtAfter` method?!

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,57 @@
...
+
+@Transactional
```
pkeugine: If you've studied `@Transactional` annotation, you'll notice there's `readOnly` option. Let's see if we can use that option in our project, and find out what difference it makes!

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,80 @@
...
+        val product =
+            productRepository.findById(request.productId).orElseThrow { NotFoundException() }
+
+        val cart =
+            cartRepository.findCartByMemberId(memberId)
+                ?: cartRepository.save(Cart(member))
+
+        val item = cart.addItem(product, request.quantity)
+        // no need to save `cart` here because of Cascade
+        // cartRepository.save(cart)
```
pkeugine: Magic~ 🪄 
I still think this feature is really cool.
Let's remove this comment now that we experienced it! :)

This review applies to other comments related to JPA magic as well.

src/main/resources/application.properties
```diff
@@ -1 +1,23 @@
...
+security.jwt.token.secret-key=your-very-secret-key-super-super-long-special-key
+security.jwt.token.expire-length-ms=3600000
+
+spring.sql.init.mode=always
+
+# from JPA
+spring.jpa.defer-datasource-initialization=true
+spring.jpa.hibernate.ddl-auto=create-drop
+spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.format_sql=true
```
pkeugine, line 15: Could you explain to me what these options are, and why you chose these values for them? 😄

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,41 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.OneToMany
+import jakarta.persistence.Table
+
+@Entity
+@Table(name = "products")
+class Product(
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    var id: Long? = null,
```
pkeugine: In some entities, `id` field is at the lowest place while in others, it's the first.
Why don't we set a convention and keep them in the similar position?
These kinds of small conventions build up to make a clean, maintainable, easy-to-predict projects 😄

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,41 @@
...
+    @Column(name = "price", nullable = false)
+    var price: Double,
+    @Column(name = "image_url", nullable = false)
+    var imageUrl: String,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
+    var options: MutableList<Option> = mutableListOf()
+
+    override fun equals(other: Any?): Boolean {
+        if (this == other) return true
```
pkeugine:
```suggestion
        if (this === other) return true
```

Looks very similar, but acts in totally different way!
I would like you to tell me why! 😄

src/main/kotlin/ecommerce/model/Product.kt
pkeugine: I see you've used `Long?` type for id 👍 
There are some other people who used `Long` and gave `0L` as the value for 'non existing' id.
What do you think the differences are?
(To give you some more information, I also prefer `null` so you are not wrong 👍 )

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,41 @@
...
+    var id: Long? = null,
+    @Column(name = "name", nullable = false)
+    var name: String,
+    @Column(name = "price", nullable = false)
+    var price: Double,
+    @Column(name = "image_url", nullable = false)
+    var imageUrl: String,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
+    var options: MutableList<Option> = mutableListOf()
```
pkeugine: Like you've said, there are bidirectional relationship between some entities! 👍 
Could you tell me why you chose this way? I'm also curious about your experience with it.
Did you feel that this relationship gave you more pros than cons?

src/test/kotlin/ecommerce/model/OptionTest.kt
```diff
@@ -0,0 +1,34 @@
...
+    fun `names can include up to 50 characters`() {
+        assertDoesNotThrow {
+            Option(name = "a".repeat(49), product = createCanvas(), quantity = 3)
+        }
+    }
+
+    @Test
+    fun `names can not include more than 50 characters`() {
+        assertThrows<IllegalArgumentException> { Option(name = "a".repeat(51), product = createCanvas(), quantity = 3) }
+    }
```
pkeugine: Nice job keeping unit tests when needed 👍

pkeugine (changes requested):
Hello Petra, I told you I would finish the review by tomorrow,
but I had some leftover times so I decided to do it a bit early :)

You actually did a very nice job with this mission,
and I can see why it took you a long time.

Since some topics we cover in this mission are very important,
I encourage you to build up your own knowledge and answers to some questions.
Show me the results, and I'll help you with the directions and depth of the knowledge when I feel like it's necessary.

If you need any help, feel free to DM me or leave a comment in this PR anytime.

Happy Hacking! 🏃 🏃 🏃

README.md
pebencze: I would definitely like to implement Double->BigDecimal and experiment with using Cascade.MERGE, REMOVE, PERSIST. 
I will try moving Options inside product to the constructor. 
As for the tests, I haven't understood the benefit of in-memory repos yet. I might experiment with them. Have you ever used such tests?

README.md
pebencze: Even though it is not of importance in our small application, I thought of being more selective in order to learn what the different Types mean. In our application CartItem does not have meaning without a Cart and Option without a Product, so everything that is done to the parent should be also visible in the child, e.g. if the parent is removed, the child should be removed as well. 
Nonetheless, I think it is generally a better practice to only include what you need. For instance we only need to persist and remove the parent and modifications happen within the same transaction (not in detached state), so PERSIST and REMOVE are enough for our purposes. I have changed it accordingly.

I have not find a reason to use MERGE or DETACH in our application and even though I have read about the terms, I need to get more familiar with them. 

All in all, CasacdeType.ALL is no harm, but we have to be avare of our choices. What do you think? Do you see a usecas for detach in this project?

src/main/kotlin/ecommerce/service/AuthService.kt
pebencze: I would like to share what I have learnt so far... :)

- @Transactional follows the ACID rules:
- Atomic: either all or no operations succeed
- Consistent: the DB remains valid
- Isolated: transactions are isolated
- Durable: changes persist

How it works:
- persistence context (entity manager) is opened at the start of a method
- entities are tracked
- at the end of the transaction, flushes changes to the DB or rolls back, then closes the persistence context

I understand transactions as a logical bundle of operations. We usually use this annotation above Service class methods (or preferably above the Service class and then it is valid for all methods inside that class). If we only have a single operation, like saving a new member, then we don't really need Transactional. However, if we retrieve data from the DB and modify it and its related entities, we need Transactional: it will flush the changes after the end of the transaction -> no need to save manually except for brand new entities; and also make sure that if something goes wrong during a transaction, everything can be rolled back and the data does not persist in the database -> e.g. if we retrieve Cart and add a new Item to it, but that something goes wrong in that Item, Cart will not be modified.

=> we need transactional for multiple dependent CRUD operations / repo calls.

Without Transaction, there is no Lazy Loading. We would get a LazyInitializationexception if we tried to access a. non-initialized member outside of a transaction. Eager Fetching would still work, because it loads the related entity immediately. 

This works with EAGER, but not with LAZY if there is no Transaction:
```
val option = optionRepository.findById(id).get()
println(option.product.name)
```

Eager is recommended for: ManyToOne and OneToOne; Lazy is recommended for OneToMany and ManyToMany.

pebencze: > Also, if you have the time, it would be great to make a small api that uses `@Transactional` annotation, and do all sorts of things inside it. (ex: throw exception after saving an entity, update an entity after saving it and see the database connection logs... etc)
> 
> By trying out those experiments and debugging the process, you'll be able to experience `@Transactional` and how persistence context in JPA works. I remember my first time trying it out, and it really felt like some kind of magic..! (And it was also hard and confusing 😂 )

Thank you!! That is a good suggestion. :)

src/main/kotlin/ecommerce/config/DatabaseFixture.kt
pebencze: Fair. Forgot to refactor...

src/main/kotlin/ecommerce/controller/CartController.kt
pebencze: Oh, we were following an article, without paying attention to those details... oops. I will look into that.

pebencze: Now I have read that some APIs provide headers, some just the JSON body and some do both. I find it redundant to do both and I admit that it was not a concious decision to include those headers, but good to know that they exist.

src/main/kotlin/ecommerce/controller/CartController.kt
pebencze: Thank you! The idea behind was the following. Imagine we have a CartItem with a quantity of 3. 
- we delete all pieces -> return Void
- we delete only 1 piece -> return the updated Cart with 2 pieces remaining -> return CartItem

This would represent better if only an update of the quantity was made or if everything was deleted.
It seemed to be a bit complicated to manage, so we opted for simplifying it and only returning an empty body. What do you think about this?

src/main/kotlin/ecommerce/model/mapper/Mapper.kt
pebencze: I have moved the class to the service package. Am I on the right track? :)

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
pebencze: It deduces the query by certain supported query subject keywords from which we can build a method name by combining them in a certain order. There are also reserved methods like findAll() or deleteById() – we don't even have to declare those methods inside our interface, because they represent basic CRUD operations that have predefined functionality.
- find..By - subject keyword
- After - logical keyword
- CreatedAt - name of our column in the DB table

src/main/kotlin/ecommerce/service/AuthService.kt
pebencze: It tells Hibernate not to do dirty checking of the entities at the end of the transaction, which can lead to performance enhancements. Otherwise Hibernate would check if it has to UPDATE something - in read-only mode it only updates when save is called explicitly. Since in Authservice we don't really change the Member, - only save it at creation, - it is better to use read-only. If the Service gets extended, we might have to get rid of this annotation or remove it from the class-level.
We use MySQL and I have read that setting readOnly to true makes only a minor difference compared to e.g. PostgreSQL.

src/main/resources/application.properties
pebencze: spring.jpa.defer-datasource-initialization=true: 
- it tells the application that JPA should create tables before any sql scripts are run
- we were planning to use data.sql, but we are not, so it is not necessary to keep this property

spring.jpa.hibernate.ddl-auto=create-drop:
- it drops and creates a schema at startup and drops at shutdown
- should not be used in production!!
- apparently Spring Boot defaults to this with my current setup even without explicitly including this flag...

spring.jpa.show-sql=true:
- it logs every SQL statement – we found it useful for debugging and understanding tests

spring.jpa.properties.hibernate.format_sql=true
- it formats the SQL logs -> instead of one line they become several lines with indentation, which is more readable

src/main/kotlin/ecommerce/model/Product.kt
pebencze: Yeah! We actually forgot about that. Initially during migration we did not want to touch the schema... will change it!

src/main/kotlin/ecommerce/model/Product.kt
pebencze: ==: is equal to the other
===: is same as the other -> same place in memory

src/main/kotlin/ecommerce/model/Product.kt
pebencze: My teammate and I had different preferences :D

pebencze: I have used a nullable type, because DB indexes could also start at 0, so `null` makes it more clear that the entity either does not yet have an index or the index was not retrieved from the DB. This way, my entity can also act as a model and I don't have to pass around the ID every time.

pebencze: Of course, we have to use safe-call operators, elvis operators or not-null assertion operators to avoid NullPointerException. That made me consider switching to `0L`, but null was oftentimes more convenient for our purposes.

src/main/kotlin/ecommerce/model/Product.kt
pebencze: First of all, this seemed the most intuitive choice to me, because it mimics real-world relationships.
Secondly, I have heard that it can lead to complications, so I wanted to try it. I did not really encounter complications, like infinite loops. I have realized that it can lead to the N+1 problem sometimes, I could see the logs in the tests and that a lot of queries are made when retrieving product through options. 
Also, I did not see another way to validate duplicate Option names inside a Product than being able to iterate through a list of options inside Product `require(options.none { it.name == this.name }) { "duplicate name $name found" }`
I thought it would be convenient to be able to access options though product, because eventually there is no Product without an Option. A Product requires at least one Option.

However, it made sense to me that a Member does not know about their Cart, only the Cart knows about the Member. because I only need the Member's ID to be able to access the Cart. Not all Members have a Cart. Additonally, I did not have to change my already existing Entity and I think it is a good practice to extend, but not to modify existing things.

What do you think about my argumentation? I still feel like I don't have enough experience and overview yet to give a fully convinced opinion on this topic, but hope that e.g. the next step of this mission helps me understand ORM better and I am planning to reflect on my code more during the next weeks.

pebencze:
Hello PK,

Thank you for the intriguing questions — it was really nice to reflect on my code, and I appreciate this approach.

I’ve implemented your suggestions, addressed the comments, and also made a few minor additional refactorings.
If you could take a look and let me know whether there’s still room for improvement or if I’ve misunderstood any concepts, I’d be glad.

Best regards,
Petra

README.md
pkeugine: I don’t usually use in-memory repositories, to be honest. For tests, I typically rely on an in-memory H2 database, which keeps the database-related logic intact while still giving me fast, isolated test runs!

README.md
pkeugine: I think it's great as it is.
It's great to try out and learn, and like you said, it's also great to only include what we need.
That also means we'll get to use other types when we need it, so I wouldn't worry too much about it!

src/main/kotlin/ecommerce/service/AuthService.kt
pkeugine: What a great summarization of knowledge you've learned!
I think you are right on track. 💯 
Now you'll be able to understand the options that `@Transactional` annotation has, and use them in right situation 👍

src/main/kotlin/ecommerce/controller/CartController.kt
pkeugine: Great! 👍 
I wasn't trying to say that we should definitely remove it, I was honestly curious about the reason behind these headers 😂 

However, you managed to think about the reason again, and make it better 👍

src/main/kotlin/ecommerce/controller/CartController.kt
pkeugine: If that is how our application has to work, I think that is one of the ways for us to make our api 👍 
I would consider making an `update` api for it though, and keep `delete` api as an actual removing method since changing the quantity feels more like an update to me. If me make an `update` api, that would also allow clients to add quantities more easily!

src/main/kotlin/ecommerce/model/mapper/Mapper.kt
pkeugine: Great!
I was thinking about `dto` or `controller` layer, since `Mapper` would be used along with DTOs, which are tightly coupled to view layer.

src/main/resources/application.properties
pkeugine: `should not be used in production!!` part really emphasizes the importance of not using it 😂 
Perfect response! 💯

src/main/kotlin/ecommerce/model/Product.kt
pkeugine: Great!
I prefer `null` over `0L` because it keeps the meaning of “non-existing” consistent across all types in the project.
For some classes, defining a default placeholder value is unclear or misleading, so `null` is the clearest option. While using `null` does introduce the possibility of NPEs, Kotlin’s null-safety features make them manageable and easy to detect at compile time.

src/main/kotlin/ecommerce/model/Product.kt
pkeugine: Already thinking of N+1 problem, I think we are gaining some speed in learning here 😮 
Also, I think your argument is persuasive, and I understand the point of trying things out for learning purpose since like I said in the previous comment about `@Transactional` annotation, trying it out is the best way to learn 👍 

Just keep in mind that bidirectional relationships can add complexity over time, and sometimes a unidirectional relation with a query is enough. Actually, more than enough that a lot of teams consider it for their relationship mapping. (This doesn't mean bidirectional mapping is bad, since other teams choose bidirectional mapping)

You'll be able to build up your thoughts on this as you gain more experience mapping entities and maintaining those logic! 👍

pkeugine (approved):
Hello Petra, and nice work refactoring the project and answering the questions!
You are doing great so far, and I think we are ready to move onto the next step.

I added some responses to your answers with some opinions/thoughts that I'd like to share. There is a point where you might want to refactor again (or not, based on your decision), but since it's a minor part, I think it'll be great for us to apply it to the next PR.

If you have any questions or thoughts that you'd like to share, feel free to DM me anytime!

Happy Hacking! 🚗 🚗 🚗

PR #15, opened 2025-08-06
[Spring-Ecommerce-Order Step 1] Submit miooo0o's Mission

miooo0o:
Hello Dani, thank you for understanding the late submission.
I’m sorry that it ended up being delayed by another day as well. 🙇 

## migrate from ecommerce and complete JPA refactor

This PR initializes the `spring-ecommerce-order` project based on the existing `spring-ecommerce` project from my pair programming partner and completes the implementation of core order-related features along with a comprehensive refactor.

---

### Entity and Relationship Setup
- Created `Product`, `Member`, `Cart`, `CartItem`, and `Option` entities and defined their relationships
- Improved handling of `CartItem.createdAt` and `updatedAt`
### Repository Structure Changes
- Migrated all repositories (including `Product` and `Member`) to JPA
- Removed JDBC dependencies

**Key decision:** We first refactored the JDBC-based repositories to match JPA return types, then fully migrated them to JPA.  
**Key decision:** For `Cart` and `CartItem`, DTOs were being used as entities, and modifying tables to fit JDBC looked too complex. We decided to delete the old implementation and rewrite it from scratch instead, which also impacted the tests.

### Feature Implementation
- **Cart Controller & Service:** Added item add/remove functionality and pagination
- **Product Controller & Service:** Added pagination and option management (Entity + DTO)
- **Statistics:** Fully refactored to JPA and improved sorting conditions

## Step 1.3 – Product Option
**Key decision:** We chose not to handle `Option` separately but instead keep it as a property of `Product`. and also use `CascadeType.ALL`.

```kotlin
@Entity
@Table(name = "products")
class Product(
    ...
) {
    @OneToMany(mappedBy = "product", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
    var options: MutableList<Option> = mutableListOf()
}
```

- Reason: To minimize changes to the existing code base due to time constraints

---

## Test Improvements

- Added controller tests for `Product`, `Member`, and `Cart` using MockMvc
- Introduced a `Fixture` object for reusable test instances
- Refactored statistics tests to use JPA
- Added unit tests for `Cart` service and controller

**Key decision:** We encountered an issue where updates to a `Product`'s `Option` list were not properly persisted to the database, especially in tests that used lateinit. To resolve this, we used `@Transactional` and explicitly accessed the options collection to force initialization.
```kotlin
@Transactional
@Service
class ProductService(private val productRepository: ProductRepository) {
    ...
    fun read(): List<Product> {
        val products = productRepository.findAll()
        products.forEach { it.options.size } // call all options here
        return products
    }
}
```
- Reason: To minimize changes to the existing code base due to time constraints

---

Thank you so much for reviewing this.
Let me know if anything is unclear or needs improvement and again, thank you for your understanding!


src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,45 @@
...
+    var price: Double,
+    @Column(name = "image_url", nullable = false)
+    var imageUrl: String,
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
+    var options: MutableList<Option> = mutableListOf()
+
+//    init {
+//        require(options.isNotEmpty()) { "Product name must not be blank" }
+//    }
```
da-nyee: What about removing these comments if they're unnecessary?

src/main/kotlin/ecommerce/model/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+import jakarta.persistence.ManyToOne
+
+@Entity
+class Option(
+    @Column(nullable = false)
+    var name: String = "",
+    @Column(nullable = false)
+    var quantity: Int = 1,
+    @ManyToOne(fetch = FetchType.EAGER)
+    var product: Product,
```
da-nyee: Is there any particular reason you set this as `EAGER`?

src/main/kotlin/ecommerce/model/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    init {
+        require(quantity in 1 until 100_000_000) { "quantity must be in range 1..100_000_000" }
```
da-nyee: Underscore 👍

src/main/kotlin/ecommerce/model/Option.kt
```diff
@@ -0,0 +1,45 @@
...
+        require(name.length <= 50) { "name length must be <= 50 characters" }
+        require(name.all { it.isLetterOrDigit() || it in allowedSpecialChars })
+        if (product.options.isNotEmpty()) {
+            require(product.options.all { it.name != this.name }) { "duplicate name ${product.name} found" }
+        }
```
da-nyee: I'd recommend delegating this validation to the service layer,
since it seems like business logic, and fetching other options for a product within an option feels a bit awkward 🤔

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,78 @@
...
+    ): PageImpl<CartItemResponse> {
+        val cart = findCart(memberId)
+        val itemResponses = cart.items.map { CartItemMapper.toResponse(it) }
+        val pageRequest = PageRequest.of(page, size, Sort.by("productName"))
+        val start = pageRequest.offset.toInt()
+        val end = min(start + pageRequest.pageSize, itemResponses.size)
+
+        val pageContent = itemResponses.subList(start, end)
+        return PageImpl<CartItemResponse>(pageContent, pageRequest, itemResponses.size.toLong())
+    }
```
da-nyee: You did well, but it feels a bit complex.

How about this?

```suggestion
    ): Page<CartItemResponse> {
        val cart = findCart(memberId)
        val itemResponses = cart.items.map { CartItemMapper.toResponse(it) }
        val pageRequest = PageRequest.of(page, size, Sort.by("productName"))

        // You need to implement `toResponse()` though.
        return cartRepository.findAll(pageRequest).toResponse()
    }
```

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,78 @@
...
+                ?: cartRepository.save(Cart(member))
+
+        cart.removeItem(product, request.quantity)
+        cartRepository.save(cart)
+    }
+
+    fun getPages(
+        memberId: Long,
+        page: Int,
+        size: Int,
```
da-nyee, line 68: Have you heard of `no-offset` pagination?
How is it different from the current approach?

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,77 @@
...
+    fun removeItem(
+        product: Product,
+        quantity: Int,
+    ) {
+        require(quantity > 0) { "Item quantity must be greater than zero." }
+        val existingItem = items.find { it.product.id == product.id } ?: throw IllegalArgumentException("Item not found.")
+        var quantityToRemove = quantity
+        if (existingItem.quantity < quantity) quantityToRemove = existingItem.quantity
+
+        existingItem.quantity -= quantityToRemove
```
da-nyee: What do you think about delegating this responsibility to the `CartItem` rather than accessing it through a getter?

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,57 @@
...
+@Service
+class AuthService(
+    private val jwtTokenProvider: JwtTokenProvider,
+    private val memberRepository: MemberRepository,
+) {
+    fun findMember(payload: String): RegisteredMember {
+        val member = memberRepository.findByEmail(payload)
+        if (member == null) {
+            throw NotFoundException("Member not found")
+        }
```
da-nyee:
```suggestion
        val member = memberRepository.findByEmail(payload) ?: throw NotFoundException("Member not found")
```

What about using the Elvis operator to make it more idiomatic Kotlin?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        if (productRepository.existsByName(productRequest.name)) {
+            throw ConflictException("Product with name ${productRequest.name} already exists")
+        }
+        val product = productRepository.save(productRequest.toProduct())
+        return product.id
+            ?: throw NotFoundException("Product with name ${productRequest.name} not found")
+    }
+
+    fun read(): List<Product> {
```
da-nyee, line 35:
```suggestion
    @Transactional(readOnly = true)
    fun read(): List<Product> {
```

You could consider setting `readOnly` when you're only querying data 👀

src/main/kotlin/ecommerce/service/StatService.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val cartItemRepository: CartItemRepository,
+) {
+    fun getActiveMembersInThePast7Days(): List<MemberStatsResponse> {
+        val threshold = LocalDateTime.now().minusDays(7)
+
+        val cartItems = cartItemRepository.findDistinctByCreatedAtAfter(threshold)
+
+        val members =
+            cartItems
+                .map { it.cart.member }
```
da-nyee: Maybe the object could handle it instead of relying on a getter 👀

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Column(name = "email", nullable = false)
+    val email: String,
+    @Column(name = "name", nullable = false)
+    var name: String = "",
+    @Column(name = "password", nullable = false)
+    var password: String,
+    @Column(name = "role", nullable = false)
+    val role: String = Role.USER.name,
+) {
+    fun validatePassword(password: String) {
```
da-nyee, line 28: What do you think about writing unit tests for the models?

da-nyee (changes requested):
Hi Mina! 🙌

You're doing great so far 👍
There are a few things that could be improved, so I've left some comments — please take a look.

If anything's unclear, just let me know 🙇‍♀️

src/main/kotlin/ecommerce/model/Member.kt
miooo0o: Here, `Member.validatePassword` doesn’t actually validate anything. It’s not from my codebase, so if I don’t need to add more validation logic right now, I’d prefer not to touch it until I actually need it. 

I wonder what you think.

src/main/kotlin/ecommerce/service/StatService.kt
miooo0o: agree.

```kotlin
            cartItems
                .map { it.getMember() }
                .distinctBy { it.id }
```

Here’s my current approach. I’ll apply the remaining changes, and if I come up with a better idea, I’ll leave another comment.

src/main/kotlin/ecommerce/service/CartService.kt
miooo0o: Ah, I see.  I completely missed that your question was about `no-offset`!

From what I’ve studied, in my current pagination using `OFFSET` and `LIMIT`, `no-offset` works more like the familiar “Load More” approach.  
If `OFFSET` needs to fetch “10 rows after skipping 1,000,” it will actually read 1,010 rows in total, discard the first 1,000, and then return the remaining 10. The more rows it has to skip, the slower it becomes.

`no-offset`, on the other hand, finds the index of the starting point for the query and only reads as many rows as the page size (like the first page), which makes it much more efficient.

src/main/kotlin/ecommerce/model/Option.kt
miooo0o: While trying to fix this, I ended up having to change quite a few things, [here](https://github.com/herotechcourse/spring-ecommerce-order/pull/15/commits/7f38956b7bc4c51064deb9665acf4a904ae437bf)

It all started with the question of which service layer this logic should belong to. As you said, this is indeed business logic. I tried placing the functions in different spots within the `ProductService`, but I couldn’t find a good place for them. One big reason was that I had linked `Product` and `Option` bidirectionally(an intentional shortcut, but not the most elegant approach) which tightly coupled their creation and made the responsibility unclear. On top of that, I was using `productRepository.save(request.toProduct())`, and the `toProduct` process didn’t even properly create the `Option`. Just... all mixed up.

In the end, I decided to use a `ProductFactory`. It felt wrong for the `Option` domain to create itself from the request, and it also didn’t seem right for the `Product` to be directly responsible for creating options. 

BUT I’m still couldn't move name validation logic to service, , please share if you have any ideas!

I’d like to hear your thoughts.

src/main/kotlin/ecommerce/model/Option.kt
miooo0o: Maybe an error occurred when `product` was set to `LAZY`? I had been using `FetchType` without really understanding it, so I had no clear reason for the choice mostly just wanted to avoid some errors. 

After studying it today, I switched it to `LAZY`.

__

According to what I learned today, `EAGER` fetching carries the risk of loading large amounts of unnecessary data, which can hurt performance. In practice, `LAZY` fetching combined with `fetch joins` is more commonly used, and even `EAGER` fetching cannot fully prevent the `N+1 problem` in some cases.

Am I understanding this correctly?

src/main/kotlin/ecommerce/service/ProductService.kt
miooo0o: Thanks for letting me know about the feature. I applied it.

src/main/kotlin/ecommerce/service/CartService.kt
miooo0o: The actual implementation was handled a bit differently.  

Since I wanted to perform pagination directly at the database, I decided to use `CartItemRepository` instead of `CartRepository`.

```kotlin
 ): Page<CartItemResponse>  {
        val pageable = PageRequest.of(page, size, Sort.by("product.name"))
        return cartItemRepository
            .findByCartMemberId(memberId, pageable)
            .map(CartItemMapper::toResponse)
    }
```

applied [here](https://github.com/herotechcourse/spring-ecommerce-order/pull/15/commits/dd8d9e4da000d5d8ae832f2d115d8f346e187472)!

miooo0o:
## Step 1 Refactoring

Hi Dani, thank you for your review.

I tried to apply all your feedback, but as I mentioned in the above comment, I couldn’t really complete this one: "Move product validation to the service layer because the logic belongs to business logic."
### Repository & Service
-  Apply fetch join in `ProductService` and `ProductRepository` to eagerly load options and prevent N+1 queries
-  Replace manual pagination in `CartService` with repository-based pagination
###  Codebase Cleanup
-  Upsert : Return both `UpsertStatus` and `ProductResponse`
-  Change `var` → `val`
-  Remove bidirectional relation setup from `Option` constructor
-  Remove unnecessary commented code (`Product.kt`)
### Factory & Domain Logic
-  Introduce `ProductFactory` to create `Product` with pre-built `Option` entities
-  Move responsibility for binding `Option` to `Product.addOption(s)`
-  Add duplicate option name validation inside `Product` 
### Cart & CartItem
-  Encapsulate quantity adjustment inside `CartItem` instead of modifying it directly in `Cart`
### Test Fixture
-  Rename `DatabaseFixture` → `TestFixture`, Move `DatabaseFixture` to the `test` directory

### Other Refactoring
-  Use Elvis operator for null checks (`AuthService.kt`)
-  Mark read-only methods with `@Transactional(readOnly = true)` (`ProductService.read`)
-  Expose member access through a method in `StatService` instead of nested property access
-  Reconsider adding unit tests for `Member.validatePassword()`

---

### What was the issue during the refactor? 
I can say, I didn’t study fetch type when I wrote that code, and I didn’t consider where/why that logic should be or what it should be with. I checked a few Tecotalk videos and tried to adopt them, but I’m still not sure my understanding is enough or even correct. But it was very interesting.

If there’s a Tecotalk you think would be good for me to watch during this process, could you let me know?
I’ll look for more on my own, but if you have any tips, I’d appreciate it.

Thank you again :)

src/main/kotlin/ecommerce/service/CartService.kt
miooo0o: applied [here](https://github.com/herotechcourse/spring-ecommerce-order/pull/15/commits/dd8d9e4da000d5d8ae832f2d115d8f346e187472)

src/main/kotlin/ecommerce/model/Cart.kt
miooo0o: good idea, I guess we implement methods but didn't applied. Now I updated [here](https://github.com/herotechcourse/spring-ecommerce-order/pull/15/commits/b4287cea5844de51cec630acf782fd5625c4a0cf)

src/main/kotlin/ecommerce/model/Product.kt
miooo0o: yeah, I completely missed! I would double check this time.

src/main/kotlin/ecommerce/service/AuthService.kt
miooo0o: Oh, I applied this part in this [commit](https://github.com/herotechcourse/spring-ecommerce-order/pull/15/commits/6f0710336f138c1c2809c51c39da60ec41311e74).

src/main/kotlin/ecommerce/model/Option.kt
```diff
@@ -0,0 +1,48 @@
...
+class Option(
+    @Column(nullable = false)
+    val name: String = "",
+    @Column(nullable = false)
+    var quantity: Int = 1,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long = 0L,
+) {
+    @ManyToOne(fetch = FetchType.LAZY)
```
da-nyee, line 21: 👍

src/main/kotlin/ecommerce/model/Option.kt
da-nyee: Yes! You've got it 👍

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,78 @@
...
+        require(!request.options.isNullOrEmpty()) { "options must not be empty" }
+
+        if (productRepository.existsByName(request.name)) {
+            throw ConflictException("Product with name ${request.name} already exists")
+        }
+        val product = productFactory.from(request)
+        return productRepository.save(product).id
+    }
+
+    @Transactional(readOnly = true)
```
da-nyee, line 34: 👍

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,56 @@
...
+@Transactional
+@Service
+class AuthService(
+    private val jwtTokenProvider: JwtTokenProvider,
+    private val memberRepository: MemberRepository,
+) {
+    fun findMember(payload: String): RegisteredMember {
+        val member =
+            memberRepository.findByEmail(payload)
+                ?: throw NotFoundException("Member not found")
```
da-nyee, line 26: 👍

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,72 @@
...
+        items.add(newItem)
+        return newItem
+    }
+
+    private fun updateExistingItem(
+        existingItem: CartItem,
+        quantity: Int,
+    ): CartItem {
+        existingItem.changeQuantityTo(quantity)
+        existingItem.markAsUpdated()
```
da-nyee, line 62: 👍

src/main/kotlin/ecommerce/service/CartService.kt
da-nyee: You nailed it 💯

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,75 @@
...
+    ): Page<CartItemResponse> {
+        val pageable = PageRequest.of(page, size, Sort.by("product.name"))
+        return cartItemRepository
+            .findByCartMemberId(memberId, pageable)
+            .map(CartItemMapper::toResponse)
```
da-nyee, line 73: Looking good 👍

src/main/kotlin/ecommerce/model/Member.kt
da-nyee: Alright, that makes sense!
Let's just leave it as is 👀

src/main/kotlin/ecommerce/service/ProductFactory.kt
```diff
@@ -0,0 +1,30 @@
...
+import ecommerce.model.Option
+import ecommerce.model.Product
+import org.springframework.stereotype.Component
+
+interface ProductFactory {
+    fun from(request: ProductRequest): Product
+}
+
+@Component
+class DefaultProductFactory : ProductFactory {
```
da-nyee, line 14: Well, I think it's a bit over-engineered 😅

What about moving this logic to `ProductService` and injecting `OptionRepository` into it to check for option name duplication? Since you're already using the helper method `addOptions`, mapping both `Product` and `Option` doesn't seem like a big deal. You could also add a helper method in `Option` to bind the `Product`.

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,57 @@
...
+) {
+    @OneToMany(mappedBy = "product", cascade = [CascadeType.ALL], fetch = FetchType.LAZY)
+    val options: MutableList<Option> = mutableListOf()
+
+    fun addOptions(newOptions: List<Option>): Product {
+        newOptions.forEach { this.addOption(it) }
+        return this
+    }
+
+    fun addOption(newOption: Option): Product {
```
da-nyee, line 35: This could be `private` 👀

da-nyee (approved):
Hi Mina! 🙌

I've gone through it and replied to your comments — please check them out when you have time.

Everything's looking good, so let's wrap it up and move on to the next one! 💪

src/main/kotlin/ecommerce/model/Option.kt
da-nyee: @miooo0o I replied to it [here](https://github.com/herotechcourse/spring-ecommerce-order/pull/15#discussion_r2265325688)!

PR #16, opened 2025-08-06
[spring-ecommerce-order - Step-1] Submit codebrew42's mission

jieun-pak:
## Feature Summary                                                           
   I and my pair implemented comprehensive JDBC to JPA migration with entity relationships, validation, JWT authentication, shopping cart functionality, and administrative analytics for an e-commerce platform.                                                                                                                                                                 
                                                                                                                                                                                        
   ## Questions                                                                                                                                                                         
   I have 5 Questions while implementing this project, it would be super helpful if you can share your opinion on any of those questions.                                               
                                                                                                                                                                                        
   **Step 1.1 - Entity Migration:**                                                                                                                                                     
   What are the performance implications of JPA lazy vs eager loading for product-option relationships?                    
                                                                                                                                                                                        
   **Step 1.2 - Database Design:**                                                                                                                                                      
   Should cart items be automatically cleaned up when JWT tokens expire?
                                                                                                                                                                                        
   **Step 1.3 - Validation Strategy:**                                                                                                                                                  
   Is it better to handle validation at the entity level (@Valid) or service layer?                                                                          
                                                                                                                                                                                                                                                                                                                                                        
   **Fundamental Concern:**                                                                                                                                                          
   How do we handle concurrent cart modifications when multiple users add the same product simultaneously with limited stock?                                                           
                                                                                                                                                                                        
   ## Notes                                                                                                                                                                             
   - **Step 1:** Complete JDBC → JPA migration with entity relationships and validation                                                                                                 
   - **Entities:** Product, ProductOption, Member, Cart, CartItem with JPA annotations                                                                                           
   - **Analytics(Cart Statistics):** Custom queries for business intelligence (top products, active users)


README.md
```diff
@@ -1 +1,64 @@
...
+- [x] **Repository** - JpaRepository with cart operations
+- [x] **Service** - Cart management logic
+- [x] **Controller** - Cart operations
+- [x] **Tests** - Integration testing
+
+## CartItem
+- [x] **Entity** - Cart-product relationship, `modify()` method
+- [x] **Repository** - JpaRepository with `@Modifying` queries
+- [x] **Service** - Refactored `saveCartItem()` (add/update unified)
+- [x] **Controller** - CRUD operations
```
Hyeon9mak, line 44: nice habit 💯

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -0,0 +1,71 @@
...
+    fun handleNotFoundException(e: NotFoundException): ErrorResponse {
+        return ErrorResponse(
+            error = "NOT_FOUND",
+            message = e.message ?: "Resource not found",
+        )
+    }
+
+    @ExceptionHandler(ProductValidationException::class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    fun handleProductValidationException(e: ProductValidationException): ErrorResponse {
```
Hyeon9mak: What would happen if an exception occurs that we didn't explicitly handle?
Hint: inheritance exception.

src/main/kotlin/ecommerce/controller/CartItemController.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    @PutMapping("/{itemId}")
+    fun updateCartItemForIncrement(
+        @PathVariable itemId: Long,
+        @PathVariable cartId: Long,
+        @RequestBody request: AddToCartRequest,
+    ): ResponseEntity<CartItem> {
+        val updatedItem = cartItemService.saveCartItem(request, itemId, cartId)
+        return ResponseEntity.ok(updatedItem)
+    }
```
Hyeon9mak, line 37: What's different between HTTP method Patch & Put?

src/main/kotlin/ecommerce/controller/PublicController.kt
```diff
@@ -0,0 +1,49 @@
...
+    @GetMapping("/admin")
+    fun adminEndpoint(): ResponseEntity<Map<String, String>> {
+        return ResponseEntity.ok(mapOf("message" to "Admin access granted"))
+    }
+
+    @GetMapping("/api/cart-items")
+    fun cartItemsEndpoint(): ResponseEntity<Map<String, String>> {
+        return ResponseEntity.ok(mapOf("message" to "Cart access granted"))
+    }
+}
```
Hyeon9mak: If the test has been completed, please clean up the code! 😄 
It makes confusion to fellow developers.

src/main/kotlin/ecommerce/dto/ProductDtos.kt
```diff
@@ -0,0 +1,100 @@
...
+    @field:NotEmpty
+    @field:Size(min = 1, message = "At least one product option must exist")
+    val productOptions: List<ProductOptionRequest>,
+    @field:NotNull(message = "Image Link must not be null")
+    @field:Pattern(
+        regexp = "^(http://|https://).*",
+        message = "url must begin with http:// or https://",
+    )
+    val imageUrl: String,
+)
```
Hyeon9mak, line 36: Request/Response DTOs and Domain Objects are used in the Presentation Layer and Domain Layer, respectively.
In your opinion, which layer is more appropriate for handling validation?

Of course, there's no absolute right answer here.
Performing validation in both layers is the safest approach, but it can also lead to redundant checks.
I'm curious to hear your thoughts. 😄

src/main/kotlin/ecommerce/interceptor/AuthInterceptor.kt
```diff
@@ -0,0 +1,85 @@
...
+            extractUserId(claims)
+                ?: throw AuthenticationException("Invalid token payload")
+
+        storeAuthenticatedUser(request, claims, userId)
+
+        if (requiresAdminAccess(request) && !hasAdminRole(claims)) {
+            throw AuthorizationException("Admin access required")
+        }
+
+        return true
```
Hyeon9mak, line 38: Nice job distinguishing between authentication and authorization! 👍 👍

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,51 @@
...
+import jakarta.persistence.OneToMany
+import jakarta.persistence.OneToOne
+import jakarta.persistence.Table
+import java.time.LocalDateTime
+
+@Entity
+@Table(name = "carts")
+class Cart(
+    @OneToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "member_id", nullable = true)
```
Hyeon9mak: Oh wow! What was ur reasoning for choosing FetchType.EAGER?

How about considering fetch strategies from a business perspective? 😄

For example, when displaying CartItem on a shopping page, it would be natural for the Product information to be shown together. In that case, EAGER might actually be a better fit than LAZY.

(It's why Developers need to understand the Business, not just Tech.)

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,51 @@
...
+    val member: Member? = null,
+    @OneToMany(mappedBy = "cart", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+    @JsonIgnore
```
Hyeon9mak: What's different between CascadeType.REMOVE and orphanRemoval = true?

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,55 @@
...
+    @JoinColumn(name = "product_option_id", nullable = true)
+    var productOption: ProductOption,
+    @Column(name = "quantity", nullable = true)
+    var quantity: Int,
+    @Column(name = "updatedAt", nullable = false)
+    var itemAddedAt: LocalDateTime? = null,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long? = null,
+) {
```
Hyeon9mak, line 39: Entity Question!

- id = 1, name = "Jin"
- id = 1, name = "Hyeon9mak"

Should these be considered the same or different?

- id = 1, name = "Jin"
- id = 99, name = "Jin"

Should these be considered the same or different?

equals and hashCode should be overridden accordingly!

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,24 @@
...
+import ecommerce.model.CartItem
+import ecommerce.model.ProductOption
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
+import org.springframework.data.jpa.repository.Query
+import org.springframework.stereotype.Repository
+
+@Repository
+interface CartItemRepository : JpaRepository<CartItem, Long> {
+    fun findByCartAndProductOption(
```
Hyeon9mak: Do we really need `@Repository` annotation? 🤔
Hint: `SimpleJpaRepository.class`

src/main/kotlin/ecommerce/repository/CartRepository.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun findByIdAndMemberId(
+        cartId: Long,
+        memberId: Long,
+    ): Cart?
+
+    @Query("SELECT c FROM Cart c JOIN c.cartItem ci WHERE c.member.id = :memberId AND ci.productOption.id = :productOptionId")
+    fun findByMemberIdAndCartItemProductOptionId(
+        memberId: Long,
+        productOptionId: Long,
+    ): Cart?
```
Hyeon9mak: Can I ask reason you chose to write the JPQL explicitly instead of using JPA’s name-based query creation?

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,106 @@
...
+    private val cartRepository: CartRepository,
+    private val cartItemRepository: CartItemRepository,
+    private val productOptionRepository: ProductOptionRepository,
+    private val memberRepository: MemberRepository,
+) {
+    fun getCartByUserId(userId: Long): Cart {
+        return cartRepository.findByMemberId(userId)
+            ?: throw NotFoundException("Cart not found for user $userId")
+    }
+
```
Hyeon9mak: What's different between `@Transactional(readOnly = false)` and `@Transactional(readOnly = true)` ?

src/main/kotlin/ecommerce/service/PasswordService.kt
```diff
@@ -0,0 +1,42 @@
...
+        return Base64.getEncoder().encodeToString(salt) + ":" + Base64.getEncoder().encodeToString(hashedBytes)
+    }
+
+    fun verifyPassword(
+        password: String,
+        storedHash: String,
+    ): Boolean {
+        if (storedHash.startsWith("$2a$") || storedHash.startsWith("$2b$") || storedHash.startsWith("$2y$")) {
+            return password == "secret" && storedHash == "\$2a\$10\$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2uheWG/igi."
+        }
```
Hyeon9mak: I was wondering what led you to separate this class — could you share a bit more about your thought process? 😉

Hyeon9mak (changes requested):
Hello Jin! welcome to the last mission!
I’ve left a few light comments that connect to the previous mission for you to review.
Let’s discuss deeper aspects after addressing these first.

Below are my answers to your questions, but I chose not to provide answers for the parts I believe you’re capable of figuring out by yourself.
I believe that learning things on your own can have a much stronger impact than simply being told the answer. That’s why I try to give you space to explore and figure things out yourself. 👍 

1-1.
https://www.baeldung.com/hibernate-lazy-eager-loading

1-2.
https://www.jwt.io/introduction

1-3.
Take some time to weigh the pros and cons of each, and let me know what you think. 😉 

Fundamental Concern.
As u know, There are many ways to think about flow control in system design (for example, using an in-memory DB to implement something like a Token Bucket algorithm)!

But in this mission, I think it's okay not to worry too much about that yet!
For now, focusing on handling transactions properly in the RDB should be enough for the current scale.
How about focusing on the mission for now and exploring these ideas step by step later? 😉 

Feel free to ask anytime!

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
jieun-pak: I've realized that unhandled exceptions in the current program could lead to a poor user experience.

### Problem
- The application currently handles only specific exceptions such as `NotFoundException` and `ProductValidationException`.
- However, if an unexpected exception occurs (e.g., `NullPointerException`, `IllegalStateException`), Spring Boot falls back to its default error handler.
- This results in `500 Internal Server Error` response with minimal information, which might be a bad user experience.

### Solution: Exception Inheritance
- Introduced a catch-all exception handler for the base Exception class to handle any uncaught exceptions.
- This will also avoid exposing internal stack traces or implementation details.

**Reflected in this commit**:
[feat: add exception handling with inheritance hierarchy](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/b6ec91364b88db04d6c0cd31347e7048afba02bb)

src/main/kotlin/ecommerce/controller/CartItemController.kt
jieun-pak: ##  PUT vs PATCH: 
- In this project, I implemented both `PUT` and `PATCH` for updating resources.
    - `PUT` for replacing the entire resource state.
    - `PATCH` allows for partial updates, but can also handle full replacements when needed.

### CartItem (Used **PUT** only)
- A simple resource: `{ productOptionId, quantity }`
- Every update replaces the entire item state
- I thought partial updates aren't needed, and using `PATCH` would be misleading here

### Product (Used **PUT** + **PATCH**)
- A more complex resource: `name`, `price`, `imageUrl`, `description`, etc.
- Used `PUT` for full replacement
- Use d`PATCH` when updating only a few fields

jieun-pak: I learned about this in the previous project `ecommerce-product` while doing pair-programming.
And I thought, `PATCH` might be not necessary for `cart-item`. 
Do you recommend any update in terms of PUT and PATCH? @Hyeon9mak

src/main/kotlin/ecommerce/controller/PublicController.kt
jieun-pak: I completely agree. 
 > Leaving dead tests can lead to confusion for both reviewers and future developers. 

I've removed the unnecessary test to keep the codebase clean, and realize that some tests were using authentication paths, I fixed that too.

Thank you for pointing it out, it must be important for my future career. 

**Reflected in this commit**: 
[refactor: remove dead test and fix authentication paths](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/c9a207c24712033deafec4987c06639b202251aa)

src/main/kotlin/ecommerce/dto/ProductDtos.kt
jieun-pak: The trainees here have discussed about **self-describing code**. And I wanted to try that idea, so I added validation logic in **two places**:

### 1. DTO Validation (`CreateProductRequest`)
- I used `@Size` and `@Pattern` to validate things like name length and allowed characters.

### 2. Domain Entity Validation (`Product.kt`)
- I added similar checks in the domain model using `require(...)`.
- thinking it might be a cool idea about `self-description` of the `entity`
- But I also understand your concern about **duplicate logic problem**, and I agree it can make the code harder to maintain.
- For the next project, I think it might be better to **remove the redundant code** and choose just **one place** to put the validation, based on these guidelines:
---

### Better Approach: Split Responsibilities

* **DTO layer** → Input format & basic validation (e.g., size, format)
* **Service layer** → Business logic validation (e.g., "is name already used?")

```kotlin
data class CreateProductRequest(
    @field:NotBlank
    @field:Size(max = 15)
    val name: String,
    
    @field:Positive
    val price: Double
)

@Service
class ProductService {
    fun createProduct(request: CreateProductRequest): Product {
        validateBusinessRules(request)
        return productRepository.save(Product(request.name, request.price))
    }

    private fun validateBusinessRules(request: CreateProductRequest) {
        if (productRepository.existsByName(request.name)) {
            throw DuplicateProductNameException()
        }
    }
}
```

jieun-pak: I wanted to try out self-descriptive code, thinking that having the entity explain itself could be helpful, even if causes some duplicate validation.
But I’ve now updated the code to avoid duplication by moving the validation logic to the`DTO(*Request.kt)` and `Service` layers.
It looks much cleaner now. Thanks for pointing it out!

**Reflected in this commit**: 
[refactor: Move validations from Entity to DTO and Service](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/78b24baac7f5be34862f20677aa5a899a83ee7ce)

src/main/kotlin/ecommerce/model/Cart.kt
jieun-pak: # JPA Fetch Strategy Optimization: EAGER vs LAZY

  ## Update in My Code
  ### 1. Cart.member: EAGER → LAZY ✅
```
class Cart(
    @OneToOne(fetch = FetchType.LAZY) //Before: EAGER
    @JoinColumn(name = "member_id", nullable = true)
    val member: Member? = null,
    //Rest of the code
)
```
  **Reason:** Most cart operations already have user context from authentication. Services like
  `getCartByUserId()` already know the userId, making eager member loading redundant.


  ### 2. ProductOption.product: EAGER → LAZY ✅
```
class ProductOption(
    @ManyToOne(fetch = FetchType.LAZY) //Before: EAGER
    @JoinColumn(name = "product_id", nullable = false)
    var product: Product,
    //Rest of the code
)
```
  **Reason:** When working with product options, the program already have product context (from
   product pages, option management). Full Product entity isn't always needed.

  ### 3. CartItem.productOption: EAGER (Kept) ✅
```
class CartItem(
    @OneToOne(fetch = FetchType.EAGER) //No change
    @JoinColumn(name = "product_option_id", nullable = true)
    var productOption: ProductOption,
    //Rest of the code
)
```
  **Reason:** Shopping cart display ALWAYS needs product details (name, price, image). This is
  core business logic: cart items are meaningless without product information.

jieun-pak: I chose `EAGER` just in case it might be necessary, but ended up thinking that `LAZY` would be better  in 2 cases.
I'll keep in mind not to default to EAGER just because it's "safe". Thanks for your insight on it!

**Reflected in this commit:**
[refactor: optimize JPA fetch strategies by replacing EAGER with LAZY](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/6f06df1d9fa71d6b52680061c52caf579a5064eb)

src/main/kotlin/ecommerce/model/Cart.kt
jieun-pak: # CascadeType.REMOVE vs orphanRemoval = true
  ## CascadeType.REMOVE

  - Triggered when: Parent entity is explicitly deleted via entityManager.remove() or repository.delete()
  - What happens: All child entities are deleted when parent is deleted
  - Example: When you delete a Cart, all its CartItems are deleted
```
  cartRepository.delete(cart) // → All cartItems are deleted too
```
  ## orphanRemoval = true

  - Triggered when: Child entity is removed from parent's collection (becomes "orphaned")
  - What happens: Child entity is automatically deleted from database when unlinked from parent
  - Example: When you remove a CartItem from Cart's collection, that CartItem is deleted
```
  cart.cartItem.remove(someCartItem) // → someCartItem is deleted from database
  cartRepository.save(cart)
```

  ## Why Cart Uses Both (CascadeType.ALL + orphanRemoval = true):
1. `CascadeType.ALL` - When a cart is deleted, all cart items should be deleted (business rule:
no orphaned cart items)
2. `orphanRemoval = true` - When user removes an item from their cart UI, that item should be
deleted from database (not just unlinked)

  This combination ensures complete lifecycle management. Children are cleaned up whether you
  delete the parent OR just remove individual children from the collection.

jieun-pak: Thanks for asking me about this!

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
jieun-pak: ## why `@Repository` here is not needed?
- Spring Data JPA automatically creates proxy beans for interfaces extending `JpaRepository`
- no need for the @Repository annotation since `SimpleJpaRepository` already handles the implementation.

**Reflected in this commit:**
[fix: remove @Repository since inheritance ofJpaRepository enables it](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/d3dd3528277fef6213bbd5da2b41a28ebc537c9b)

src/main/kotlin/ecommerce/model/CartItem.kt
jieun-pak: ## Why Should We Override `equals()` and `hashCode()`?

### Problem Without Overriding

```kotlin
val item1 = CartItem(cart, productOption, quantity = 2)
val item2 = CartItem(cart, productOption, quantity = 3)

println(item1 == item2) // false
```

* Even though `item1` and `item2` have the same `cart` and `productOption`, they are treated as different objects.
* This happens because the default `==` in Kotlin/Java compares object references.
* However, from a business perspective, these should be considered the **same cart item**, only with a different quantity.

---

## Why Use `cart + productOption` as the Equality Key?

Business rule:

> A cart cannot contain duplicate items with the same product option.

### Example (like Amazon):

* If you add "iPhone 15 Pro 256GB" to your cart twice, it shouldn't appear as two separate items.
* Instead, the quantity of the existing item should increase.

---

## Understanding the `equals()` Logic

```kotlin
override fun equals(other: Any?): Boolean {
    if (this === other) return true
    if (other !is CartItem) return false

    if (id == null || other.id == null) {
        return cart == other.cart && productOption == other.productOption
    }

    return id == other.id
}
```

### Two comparison cases:

#### 1. New entities (not yet saved, `id == null`)

* We compare based on `cart` and `productOption`.
* This reflects the business identity of a cart item.

#### 2. Persisted entities (`id != null`)

* The database ID becomes the definitive identity.
* If IDs match, they are the same item.

---

## Why Two Comparison Strategies?

* Before saving to the database, there's no ID, so we use business logic to define identity.
* After saving, the database ID is the absolute truth.

---

## Understanding the `hashCode()` Logic

```kotlin
override fun hashCode(): Int {
    return Objects.hash(cart.id, productOption.id)
}
```

* The hash is based on `cart.id` and `productOption.id`, which are stable and represent business identity.
* `hashCode()` must always be consistent with `equals()`.

---

## Real Example

```kotlin
val cartItems = mutableSetOf<CartItem>()
cartItems.add(CartItem(cart, iPhone, quantity = 1))
cartItems.add(CartItem(cart, iPhone, quantity = 2))

println(cartItems.size) // Without equals/hashCode → 2 (wrong)
```

```kotlin
// With proper equals/hashCode
println(cartItems.size) // 1 (correct)
```

---

## Summary

* Entity identity should reflect business identity, not just memory reference.
* In e-commerce, a unique cart item is defined by the combination of `cart` and `productOption`.
* This business rule is enforced through proper implementation of `equals()` and `hashCode()`.

jieun-pak: This feedback is very important and basic one which I need to keep in mind.
Thanks a lot for explaining the necessity of `override functions`!

**Reflected in this commit:**
[fix: override function equals and hashCode for identifying CartItem properly](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/60f3098510f7ff6d469f0dde00ac351760e7ac5b)

src/main/kotlin/ecommerce/repository/CartRepository.kt
jieun-pak: I initially thought a custom query was necessary for `fun findByMemberIdAndCartItemProductOptionId`, and tried using custom queries even for simple cases thinking it can be clear to have them too.

However, I realized that for straightforward queries, JPA's name-based ones are more then enough. And most of all, it's not consistent and there was no need to do it.

I created two commits for this, and the code became simpler 😸 
Thanks for pointing it out! 

**Reflected in these two commits:**
[fix: replace JPQL queries with JPA's name-based queries](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/97b0ad642147f97929aa3a4a395af97c6f0b3aa3)

[fix: replace JPQL queries with JPA's name-based queries](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/ec79f5dbb05c44cd3d2550a5b434c3f2a2a13827)

src/main/kotlin/ecommerce/service/CartService.kt
jieun-pak: ## `readOnly = false` is default
- I was using `@Transactional` on methods like `addToCart`, `clearCart`, and `updateQuantity`, not specifying `readOnly = false` which is the default.

## where `readOnly = true` is needed
- Methods like `getCartByUserId`, `getCartItemsOfCartByCartId`, etc. perform only read operations but aren't marked with @Transactional(readOnly = true).

## Why it matters:
- `@Transactional(readOnly = true)` can
    - Optimizes database interactions.
    - Prevents unnecessary flush operations.
    - Hints that no changes will be made.

-  while`@Transactional(readOnly = false) (default)` allows entity updates.

jieun-pak: I added `@Transactional(readOnly = true)` to read-only methods to these functions, after thinking about your question.
```
@Transactional(readOnly = true)
fun getCartByUserId

@Transactional(readOnly = true)
 fun getCartByIdAndUserId

@Transactional(readOnly = true)
fun getCartItemsOfCartByCartId
```
       
Thank you so much!

**Reflected in this commit:**
[fix: add @Transactional(readOnly = true) for optimization](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/45c88e57cf23b76056c29ecdcacbc64092457876)

src/main/kotlin/ecommerce/service/PasswordService.kt
jieun-pak: I separated password-related logic (hashing, verification) into a single unit to separate concerns, making the service layer easy to maintain and test.
For example, I imagined that this structure might be easy to be modified like when switch hashing strategies in the future.

Do you recommend me any improvement here 😃 ? @Hyeon9mak

src/main/kotlin/ecommerce/model/Cart.kt
codebrew42: git add . && git commit -m "perf: fix N+1 query problem in cart operations                       │
│                                                                                                    │
│   Add optimizations to prevent N+1 queries when loading cart data:                                 │
│                                                                                                    │
│   - Cart.cartItem: Add @BatchSize(20) to batch-load cart items                                     │
│   - CartItemRepository.findByCartId: Add @EntityGraph to eagerly fetch productOption and product   │
│    in single query                                                                                 │
│                                                                                                    │
│   Performance improvement:                                                                         │
│   - Before: 1 + N queries (cart + N individual item queries)                                       │
│   - After: 1-2 queries total (cart + batched items with joined product data)                       │
│                                                                                                    │
│   Significant performance boost for cart operations with multiple items.

jieun-pak: #  Why `@BatchSize` and `@EntityGraph` Are Critical for Cart Performance

 ## The N+1 Query Problem

  Without Optimization:
```
  -- Loading a cart with 10 items triggers 11+ queries:
  SELECT * FROM carts WHERE id = 1;                    -- 1 query for cart
  SELECT * FROM cart_items WHERE cart_id = 1;          -- 1 query for all items
  SELECT * FROM product_options WHERE id = 1;          -- Query for item 1's product
  SELECT * FROM product_options WHERE id = 2;          -- Query for item 2's product
  -- ... 8 more individual queries for remaining items
```
  ## Solution

  1. @BatchSize(size = 20) on Cart.cartItem Collection

  - Problem: Loading cart items one-by-one when accessing the collection
  - Solution: Hibernate batches up to 20 cart item loads into a single WHERE cart_id IN 
  (1,2,3...) query
  - Result: Multiple cart operations become more efficient

  2. @EntityGraph on CartItemRepository.findByCartId()

  - Problem: Each CartItem triggers separate queries for ProductOption and Product
  - Solution: Single query with JOINs fetches all related data at once
  - SQL Generated:
```
  SELECT ci.*, po.*, p.*
  FROM cart_items ci
  LEFT JOIN product_options po ON ci.product_option_id = po.id
  LEFT JOIN products p ON po.product_id = p.id
  WHERE ci.cart_id = 1;
```
  ## Performance Impact

  - Before: 1 cart + 10 items + 10 product options = 21 queries
  - After: 1 optimized query with JOINs = 1 query

jieun-pak: I refined fetch strategies in entities (LAZY vs EAGER) first, then further optimized actual query execution with @BatchSize and @EntityGraph. For better design and runtime performance.

**Reflected in this commit:**
[fix: introduce @BatchSize and @EntityGraph for Cart Performance](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/27f88ec3be0874d9ef82492b06acaba0327f90ac)

src/main/kotlin/ecommerce/dto/ProductDtos.kt
jieun-pak: @woowahan-pjs
This is the self-descriptive part causing duplicate validation, which I asked about today. In the end, it’s going to be removed… 🤣

jieun-pak:
Hello, Gumak!

Your valuable resources, feedback, and insightful questions helped me a lot to grow.

Thinking about your questions let me think deeper. And I appreciate you being considerate by giving the small hints. I was able to study efficiently.

There were things I had overlooked but were really important, like `LAZY vs EAGER`, `overriding equals and hashCode` and `unnecessary JPQL queries`. I will keep your feedback in mind going forward.

I tried to carefully reflect on your feedback and fixed the critical issues accordingly.
Thank you again for taking your valuable time for this project.

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -0,0 +1,85 @@
...
+    @ExceptionHandler(RuntimeException::class)
+    @ResponseStatus(HttpStatus.INTERNAL_SERVER_ERROR)
+    fun handleGenericException(e: RuntimeException): ErrorResponse {
+        log.error("Unhandled exception occurred", e)
+
+        return ErrorResponse(
+            error = "INTERNAL_SERVER_ERROR",
+            message = "An unexpected error occurred. Please try again later.",
+        )
+    }
```
Hyeon9mak, line 29: 💯 👍

src/main/kotlin/ecommerce/controller/AdminController.kt
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @PutMapping("/{id}")
+    fun updateProduct(
+        @Valid @RequestBody productRequest: UpdateProductRequest,
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> {
+        val updated = productService.updateProduct(id, productRequest)
+        return ResponseEntity.ok(updated)
+    }
```
Hyeon9mak, line 56: As Jin explained, the choice between PUT and PATCH depends on whether you want to update the whole resource or just part of it. I think you applied this well, and I was simply curious if you are distinguishing between the two when using them. 😄 👍

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,59 @@
...
+)
+class Cart(
+    @OneToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "member_id", nullable = true)
+    val member: Member? = null,
+    @OneToMany(mappedBy = "cart", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+    @BatchSize(size = 20)
+    @JsonIgnore
+    val cartItem: MutableList<CartItem> = mutableListOf(),
+    @Column(name = "quantity", nullable = false)
```
Hyeon9mak: > I chose EAGER just in case it might be necessary, but ended up thinking that LAZY would be better in 2 cases.
I'll keep in mind not to default to EAGER just because it's "safe". Thanks for your insight on it!

Hmm, I was actually curious whether you chose between EAGER and LAZY intentionally, knowing their differences. For example, in the case of Cart and CartItem, since users will always need the CartItems when retrieving a Cart, choosing EAGER loading might be more reasonable!

src/main/kotlin/ecommerce/service/PasswordService.kt
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Base64
+
+@Service
+class PasswordService {
+    private val secureRandom = SecureRandom()
+
+    fun hashPassword(password: String): String {
+        return BCrypt.hashpw(password, BCrypt.gensalt())
+    }
+
```
Hyeon9mak, line 16: Since this class contains only pure logic without maintaining any state, it seems possible to delegate the responsibility to the Member and ask it whether the password is correct. I’m curious about the decision process behind separately naming it Service and adding annotations. 🤔

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,52 @@
...
+    @Column(name = "name", nullable = false)
+    val name: String,
+    @Column(name = "role", nullable = false)
+    @Enumerated(EnumType.STRING)
+    val role: Role = Role.USER,
+    @OneToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "cart_id", nullable = true)
+    var cart: Cart? = null,
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
```
Hyeon9mak: Since a Cart is always created after a Member, and currently there’s no logic where Member directly uses Cart, it might be possible to remove the Cart dependency from Member. What do you think about that? 😉

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,52 @@
...
+    val id: Long? = null,
+) {
+    constructor(
+        email: String,
+        name: String,
+    ) : this(
+        email = email,
+        password = "",
+        name = name,
```
Hyeon9mak, line 40: https://github.com/herotechcourse/spring-ecommerce-order/pull/16#discussion_r2260456943 😉

src/main/kotlin/ecommerce/repository/ProductOptionRepository.kt
```diff
@@ -0,0 +1,12 @@
+package ecommerce.repository
+
+import ecommerce.model.ProductOption
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+
+@Repository
+interface ProductOptionRepository : JpaRepository<ProductOption, Long> {
```
Hyeon9mak, line 8: https://github.com/herotechcourse/spring-ecommerce-order/pull/16#discussion_r2260458349 🤔

src/main/kotlin/ecommerce/service/CartItemService.kt
```diff
@@ -0,0 +1,160 @@
...
+    private fun validateQuantityIncrement(
+        request: AddToCartRequest,
+        productOption: ecommerce.model.ProductOption,
+    ) {
+        if (request.newProductOptionQuantity <= productOption.quantity) {
+            throw IllegalArgumentException(
+                "New product option quantity (${request.newProductOptionQuantity}) " +
+                    "must be greater than current quantity (${productOption.quantity})",
+            )
+        }
```
Hyeon9mak: ask to object. ask to `ProductOption`. it makes more super easier generate test code.

```kotlin
productOption.checkQuantity(...)
```

src/main/kotlin/ecommerce/service/CartItemService.kt
```diff
@@ -0,0 +1,160 @@
...
+
+        // Note: Stock validation removed to allow business flexibility
+        // This must be handled at checkout time LATER
+        // if (requestedQuantity > productOption.quantity) {
+        //     throw IllegalArgumentException(
+        //         "Requested quantity ($requestedQuantity) exceeds available stock (${productOption.quantity})"
+        //     )
+        // }
+    }
```
Hyeon9mak: 🤔

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,109 @@
...
+                    ?: throw NotFoundException("Member not found")
+            val newCart =
+                Cart(
+                    member = member,
+                    cartItem = mutableListOf(),
+                    quantity = request.newProductOptionQuantity,
+                    newItemAddedAt = LocalDateTime.now(),
+                )
+            cartRepository.save(newCart)
+        }
```
Hyeon9mak: In JPA, when you modify fields of an `@Entity` (other than the ID) within a transaction, the changes are automatically updated (saved) when the transaction ends. This feature is called dirty checking, and it can be quite useful!
(That said, I personally prefer calling save() explicitly — I find it more clear and intentional.)

Hyeon9mak (changes requested):
Hello Jin, good work to responding to the feedback comments so promptly. 👍 

However, I noticed that several parts haven’t been reflected in the actual code yet, so I’ve left some additional comments.
Please also take a moment to review the parts I didn’t specifically comment on, just to make sure nothing was missed. 💪 

It’d be great to take your time and gradually improve the overall completeness of the project.
Feel free to ask any questions anytime! 😄

src/main/kotlin/ecommerce/service/CartItemService.kt
jieun-pak: I removed this note from this branch.

src/main/kotlin/ecommerce/model/Cart.kt
jieun-pak: I understood the concepts and the difference itself before, but didn't have a full understanding about the better domain design. Now the structure is changed. Thanks.

[fix: change Cart → CartItems (LAZY → EAGER)](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/88b877774c509484898ffbee874b79cf198737b0)

src/main/kotlin/ecommerce/service/PasswordService.kt
jieun-pak: I had two reasons behind this design
1. including a hashPassword method for encryption to the Member entity may pollute the domain model. 
2. it's easy to mock/test Password independently to separate them.

I would appreciate hearing your suggestions again on it.

src/main/kotlin/ecommerce/model/Member.kt
jieun-pak: Yes, removing it is not only possible but seems like a cleaner design.

Thank you for your suggestion.

[fix: make member-cart relationship unidirectional](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/b1414c6366bf7d6170456156fd9f61213cd3601c)

src/main/kotlin/ecommerce/model/Member.kt
jieun-pak: [add: override function equals and hashCode for identifying entities properly](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/0e2e20a8e6e1d9df3af406702bb72d33d3fe6896)

src/main/kotlin/ecommerce/repository/ProductOptionRepository.kt
jieun-pak: Do you mean I need to implement it to implement additional methods here?


[add: override function equals and hashCode for identifying entities properly](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/0e2e20a8e6e1d9df3af406702bb72d33d3fe6896)

src/main/kotlin/ecommerce/service/CartItemService.kt
jieun-pak: I thought it could be also easy to understand and handle them by locating related methods in one file: CartItemService. But I understand that your way is easier. 

[refactor: locate quantity related validation to its entity](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/c623cc980d75ac9ab4eef5e4346382b694ef9449)

[remove: remove unnecessary parts and used quantity related validation of its entity](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/4f789631b3e910ac4719a28de23392ad13d6b4c0)

src/main/kotlin/ecommerce/controller/AdminController.kt
jieun-pak: Yes, as I mentioned, I learned about it during the previous projects. 😄 👍

src/main/kotlin/ecommerce/service/CartService.kt
jieun-pak: `Commit 1` refactored my code to a more domain-driven design, making the entity responsible for its own state changes. This is the previous stage for the second commit. 

`Commit 2` directly applied the dirty checking concept by removing the redundant `save()` and keeping `save()` which are still needed for new entities.

**Commit 1:**
[refactor: locate add/update/patch logic to domain entities](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/2c74eb39ae3da483b06f29a8a4add416eefa6227)

**Commit 2:**
[refactor: rely on JPA dirty checking and call save() only for new entities](https://github.com/herotechcourse/spring-ecommerce-order/pull/16/commits/47563c42df94fc69c47d2494ac6adba64f06484b)

jieun-pak:
Thank you for the detailed and insightful feedback. I really appreciate it.
Our conversation here and on slack made me reflect and think more deeply about my strategy and design.

I’ve applied your feedback, and I’d be grateful if you could take another look when you have time.  😄

src/main/kotlin/ecommerce/service/PasswordService.kt
Hyeon9mak: 1. What does it mean when we say a domain model is "polluted"? If it becomes polluted, what exact changes occur in the code or maintainability?
2. Does adding the `@Service` annotation make testing easier? If so, why does it make testing easier?

Jin, what is your understanding of the `@Service` annotation? It would be great if you could also share any additional thoughts on this!

src/main/kotlin/ecommerce/repository/ProductOptionRepository.kt
Hyeon9mak: Have you looked into the `SimpleJpaRepository.class` I mentioned? 👀 
If you look into it, you’ll be able to understand what the comment means.

src/main/kotlin/ecommerce/service/CartItemService.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    private fun createNewCartItem(
+        cart: ecommerce.model.Cart,
+        productOption: ecommerce.model.ProductOption,
+        request: AddToCartRequest,
+    ): CartItem {
+        CartItem.validateQuantity(request.newProductOptionQuantity)
+
+        return cartItemRepository.save(
```
Hyeon9mak, line 85: > I thought it could be also easy to understand and handle them by locating related methods in one file: CartItemService. But I understand that your way is easier.

I see.... From what I saw in CartItemServiceTest, the procedure to run the test seems much longer compared to when only using ProductOption. What are your thoughts on that?

Hyeon9mak (approved):
Hello Jin, great job incorporating the feedback.

There isn’t much time left for the mission,
and I believe there’s still plenty of room for improvement even while moving on to the next step. 💪 

Let’s keep up the momentum until the very end—see you in the next stage! 😄

PR #17, opened 2025-08-06
[Spring-Ecommerce-Order Step 1] Submit Apoo's Mission

Apoorva-S-Natesh:
Hi Seongho,
This pull request completes the migration of our ecommerce project's persistence layer from JDBC to JPA using Spring Data JPA in Kotlin. It introduces annotated JPA entities, Spring Data repositories, service layer refactoring, and comprehensive controller and integration tests supporting the new data access strategy.

**Migration Setup**

- Updated project dependencies (build.gradle.kts) with JPA, Hibernate, and related libraries.

- Configured application.properties for compatibility with H2 (test) and MySQL (production).

- Prepared data.sql script to initialize consistent test data throughout integration tests.

**Entities**

- Complete JPA entity mappings for:

- Cart, CartItem, CartStatistics,
- Member (with roles and cart relationship),
- Product, and ProductOption.
- Relationships annotated with appropriate cascade and fetch strategies.
- Business logic methods (subtract(), modify()) integrated within entities.

**Repositories**

- Replaced JDBC DAOs with Spring Data JPA repositories.

- Created custom JPQL queries annotated with @Query for complex retrieval.

- Utilized Kotlin idiomatic nullable types instead of Optional.

**Services**

- Refactored service layers for all entities:
- Handling CRUD with transaction management (@Transactional).
- Unified add/update operations (saveProductOption(), saveCartItem()).
- Exception handling with domain-specific exceptions (NotFoundException, DuplicateNameException).

**I would Like your suggestions and feedback on these:**

- Are the JPA mappings and cascade/fetch configurations optimal for our use case?
- Any recommendations on improving lazy-loading strategies to balance performance and convenience?
- Do the JPQL queries, especially complex ones in CartStatisticsRepository, adhere to best practices?
- Are @Transactional scopes properly applied and sufficient?
- How to design and implement an appropriate testing strategy for the implemented features.
-Why developer need to adapt to the OOP with ORM, not RDB with Procedure.

We tried entity mapping by hands and also used a tool to create one.
Untitled diagram _ Mermaid Chart-2025-08-06-125033

I look forward for your review! 
Thank you!


build.gradle.kts
```diff
@@ -19,16 +20,25 @@ repositories {
...
+    runtimeOnly("io.jsonwebtoken:jjwt-impl:0.12.5")
+    runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.12.5")
+    developmentOnly("org.springframework.boot:spring-boot-devtools")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     runtimeOnly("com.h2database:h2")
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
+    testImplementation("io.rest-assured:rest-assured:$assureVersion")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    developmentOnly("org.springframework.boot:spring-boot-devtools")
```
Wordbe: `devtools` is duplicated. You might want to clean that up. 🙂

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -0,0 +1,71 @@
...
+@RestControllerAdvice
+class GlobalControllerAdvice {
+    @ExceptionHandler(NotFoundException::class)
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    fun handleNotFoundException(e: NotFoundException): ErrorResponse {
+        return ErrorResponse(
+            error = "NOT_FOUND",
+            message = e.message ?: "Resource not found",
+        )
+    }
```
Wordbe: These messages could be managed within the exception classes. 💡

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    @ExceptionHandler(value = [IllegalArgumentException::class, InsufficientProductOptionsException::class])
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    fun handleIllegalArgumentException(e: Exception): ErrorResponse {
+        return ErrorResponse(
+            error = "BAD_REQUEST",
+            message = e.message ?: "Invalid request",
+        )
```
Wordbe: Same here.

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -1,11 +1,51 @@
...
-    val quantity: Int,
-    val addedAt: LocalDateTime? = null,
-)
+) {
+    constructor() : this(
+        member = null,
+        cartItem = mutableListOf(),
+        quantity = 0,
+        newItemAddedAt = LocalDateTime.now(),
+    )
```
Wordbe: This constructor doesn’t seem necessary! Take a look at the primary constructor above and think about why that might be. 🙂

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -1,11 +1,51 @@
...
+
+    constructor(member: Member) : this(
+        member = member,
+        cartItem = mutableListOf(),
+        quantity = 0,
+        newItemAddedAt = LocalDateTime.now(),
+    )
```
Wordbe: It looks like this isn’t being used anywhere. 🧐

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -1,11 +1,51 @@
...
+
+    override fun toString(): String {
+        return "Cart(id=$id, member=$member, cartItem=$cartItem, quantity=$quantity)"
+    }
```
Wordbe: Where is this being used? 🔍

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -1,62 +1,44 @@
...
-    @PostMapping("/cart-items")
-    fun addToCart(
-        @Valid @RequestBody addToCartRequest: AddToCartRequest,
-        user: AuthenticatedUser,
-    ): ResponseEntity<Cart> {
-        val cart = cartService.addToCart(user.userId, addToCartRequest)
-        return ResponseEntity.created(URI.create("/api/cart-items")).body(cart)
+    @GetMapping("/{userId}")
+    fun getCart(
+        @PathVariable userId: Long,
```
Wordbe: Since it looks like a `{cartId}` placeholder, I suggest using a query string instead to design the REST API!

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -1,62 +1,44 @@
...
+        return cartService.getCartByUserId(userId)
     }
 
-    @PutMapping("/cart-items/{productId}")
-    fun updateQuantity(
-        @PathVariable productId: Long,
-        @Valid @RequestBody updateRequest: UpdateQuantityRequest,
+    @GetMapping("/{cartId}/items")
+    fun getAllCartItemsOfCart(
+        @PathVariable cartId: Long,
```
Wordbe: It seems like it would be more appropriate to place this in `CartItemController`! 🙂

src/main/kotlin/ecommerce/controller/CartStatisticsController.kt
```diff
@@ -2,23 +2,25 @@ package ecommerce.controller
...
-    private val analyticsRepository: AnalyticsRepository,
+class CartStatisticsController(
+    //  private val cartStatisticsRepository: CartStatisticsRepository,
+    private val cartStatisticsService: CartStatisticsService,
 ) {
     @GetMapping("/top-products")
     fun getTopProducts(): List<TopProductAnalytics> {
-        return analyticsRepository.findTop5MostAddedProductsLast30Days()
+        return cartStatisticsService.findTop5MostAddedProductsLast30Days()
+        // return cartStatisticsRepository.findTop5MostAddedProductsLast30Days()
```
Wordbe: Please remove the comment. 🧽

src/main/kotlin/ecommerce/dto/ProductDtos.kt
```diff
@@ -0,0 +1,60 @@
+package ecommerce.dto
+
+import jakarta.validation.constraints.Max
+import jakarta.validation.constraints.Min
+import jakarta.validation.constraints.NotEmpty
+import jakarta.validation.constraints.NotNull
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Size
+
+data class ProductRequest(
```
Wordbe: Why did you use a `data class`? Is it really necessary? 🤔

src/main/kotlin/ecommerce/dto/ProductDtos.kt
```diff
@@ -0,0 +1,60 @@
...
+    val productOptions: List<ProductOptionRequest>,
+    @field:NotNull(message = "Image Link must not be null")
+    @field:Pattern(
+        regexp = "^(http://|https://).*",
+        message = "url must begin with http:// or https://",
+    )
+    val imageUrl: String,
+)
+
+data class ProductOptionRequest(
```
Wordbe: Same here.

src/main/kotlin/ecommerce/dto/ProductDtos.kt
```diff
@@ -0,0 +1,60 @@
...
+    val name: String,
+    @field:NotNull(message = "quantity must not be null")
+    @field:Min(1, message = "quantity must be greater than 0")
+    @field:Max(99999999, message = "quantity must be lesser than 100,000,000")
+    val quantity: Int,
+    @field:NotNull(message = "Product must not be blank")
+    var productId: Long,
+)
+
+data class ProductPatchRequest(
```
Wordbe: Same here.

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -1,11 +1,51 @@
...
-data class Cart(
+@Entity
+@Table(name = "carts")
+class Cart(
+    @OneToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "member_id", nullable = true)
+    val member: Member? = null,
+    @OneToMany(mappedBy = "cart", cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+    @JsonIgnore
+    val cartItem: MutableList<CartItem> = mutableListOf(),
```
Wordbe: It would be better to create and use a separate DTO like `CartResponse` 🙂

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,55 @@
...
+    ) {
+        if (cart != null) {
+            this.cart = cart
+        }
+        if (productOption != null) {
+            this.productOption = productOption
+        }
+        if (quantity != null) {
+            this.quantity = quantity
+        }
```
Wordbe: Since `Int` is always non-nullable, there’s no need for a null check here.

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,55 @@
...
+        if (itemAddedAt != null) {
+            this.itemAddedAt = itemAddedAt
+        }
+    }
+
+    override fun toString(): String {
+        return "Cart Item(id=$id, productOption=$productOption, quantity=$quantity)"
+    }
```
Wordbe: If it’s not needed, please delete it. Also, instead of exposing the `CartItem` entity directly from the controller, it would be better to convert it to a DTO. What are the benefits of using a DTO? 🤔

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -1,9 +1,46 @@
...
+    val id: Long? = null,
+) {
+    constructor(
+        email: String,
+        name: String,
+    ) : this(
+        email = email,
+        password = "",
+        name = name,
+    )
```
Wordbe: It is being not used!

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -1,9 +1,46 @@
...
+
+    override fun toString(): String {
+        return "Member(id=$id, email=$email, name=$name, role=$role, cart=$cart)"
+    }
```
Wordbe: Same here.

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -1,8 +1,39 @@
...
+    }
+
+    companion object {
+        private const val MAX_NAME_LENGTH = 15
+        private const val ALLOWED_SPECIAL_CHAR = "()[]+-&/_ "
+    }
+
+    override fun toString(): String {
+        return "Product(id=$id, name=$name, price=$price quantity=$quantity)"
+    }
```
Wordbe: Where this method is used? 🙂

src/main/kotlin/ecommerce/model/ProductOption.kt
```diff
@@ -0,0 +1,50 @@
...
+    companion object {
+        private const val MIN_QUANTITY = 1
+        private const val MAX_QUANTITY = 99999999
+        private const val MAX_OPTION_NAME_LENGTH = 50
+        private const val ALLOWED_SPECIAL_CHAR = "()[]+-&/_ "
+    }
+
+    override fun toString(): String {
+        return "Product Option(id=$id, name=$name, quantity=$quantity, product=$product)"
+    }
```
Wordbe: Same here.

src/main/kotlin/ecommerce/model/ProductOption.kt
```diff
@@ -0,0 +1,50 @@
...
+@Entity
+@Table(name = "product_options")
+class ProductOption(
+    @Column(name = "name", nullable = false)
+    var name: String,
+    @Column(name = "quantity", nullable = false)
+    var quantity: Int,
+    @ManyToOne(fetch = FetchType.EAGER)
+    @JoinColumn(name = "product_id", nullable = false)
+    var product: Product,
```
Wordbe, line 22: Is there a specific reason you chose `EAGER` over `LAZY`? With `EAGER`, related queries will always be triggered 🤔

src/main/kotlin/ecommerce/model/ProductOption.kt
```diff
@@ -0,0 +1,50 @@
...
+        require(name.length in 0..MAX_OPTION_NAME_LENGTH) { "Name should have length between 1 and 50" }
+        require(name.all { it.isLetterOrDigit() || it in ALLOWED_SPECIAL_CHAR }) { "Name contains invalid characters" }
+        require(quantity in MIN_QUANTITY..MAX_QUANTITY)
+    }
+
+    fun subtract(quantity: Int) {
+        if (quantity < 1) throw IllegalArgumentException("Quantity must be greater than 0")
+        if (quantity > this.quantity) throw IllegalArgumentException("Quantity can not be greater than stock of product quantity")
+        this.quantity -= quantity
+    }
```
Wordbe, line 38: Great validation 👍

src/main/kotlin/ecommerce/repository/CartRepository.kt
```diff
@@ -1,93 +1,29 @@
...
+
+    @Query("SELECT c FROM Cart c JOIN c.cartItem ci WHERE c.member.id = :memberId AND ci.productOption.id = :productOptionId")
+    fun findByMemberIdAndCartItemProductOptionId(
+        memberId: Long,
+        productOptionId: Long,
+    ): Cart?
+
+    @Modifying
+    @Query("DELETE FROM Cart c WHERE c.member.id = :memberId")
+    fun deleteByMemberId(memberId: Long)
```
Wordbe: If you're more often querying by `memberId` rather than by `Member`, it might be a good idea to also include a `memberId` field inside the `Cart` entity. 🙂

src/main/kotlin/ecommerce/service/CartItemService.kt
```diff
@@ -0,0 +1,134 @@
...
+        productOption.quantity = request.newProductOptionQuantity
+        productOptionRepository.save(productOption)
+    }
+
+    private fun updateCartQuantity(
+        cart: ecommerce.model.Cart,
+        request: AddToCartRequest,
+    ) {
+        cart.quantity += request.newProductOptionQuantity
+    }
```
Wordbe: If possible, try moving this entire part (private method) into the domain model to strengthen cohesion!

src/main/kotlin/ecommerce/service/MemberService.kt
```diff
@@ -1,29 +1,38 @@
...
-        val member =
-            Member(
-                email = request.email,
-                password = hashedPassword,
-                name = request.name,
-                role = request.role,
-            )
+        val member = request.toModel(hashedPassword)
+
+        member.cart = cartRepository.save(Cart())
```
Wordbe: Instead of using a setter, try creating an explicit method. 🙂

src/main/kotlin/ecommerce/service/PasswordService.kt
```diff
@@ -21,6 +21,13 @@ class PasswordService {
         password: String,
         storedHash: String,
     ): Boolean {
+        if (storedHash.startsWith("$2a$") || storedHash.startsWith("$2b$") || storedHash.startsWith("$2y$")) {
+            return password == "secret" && storedHash == "\$2a\$10\$92IXUNpkjO0rOQ5byMi.Ye4oKoEa3Ro9llC/.og/at2uheWG/igi."
+        }
```
Wordbe: This is hardcoded, so it looks like it needs to be fixed. ⚠️

src/main/kotlin/ecommerce/service/ProductOptionService.kt
```diff
@@ -0,0 +1,60 @@
...
+        productId: Long,
+        request: ProductOptionRequest,
+        productOptionId: Long? = null,
+    ): ProductOption {
+        if (!productRepository.existsById(productId)) {
+            throw NotFoundException("Product does not exist")
+        }
+
+        val product = productRepository.findById(productId).get()
+        val existingOption = productOptionId?.let { productOptionRepository.findById(it).orElse(null) }
```
Wordbe: Kotlin provides a function called `findByIdOrNull`. Try using it! 🙂

src/main/kotlin/ecommerce/util/Utils.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import ecommerce.dto.ProductRequest
+import ecommerce.dto.member.RegisterRequest
+import ecommerce.model.Member
+import ecommerce.model.Product
+import ecommerce.model.ProductOption
+
+fun ProductRequest.toModel(id: Long? = null) = Product(name, price, quantity, imageUrl, id)
+
+fun RegisterRequest.toModel(hashedPassword: String) = Member(email, hashedPassword, name, role)
```
Wordbe: The filename `Utils` would convey the meaning better if it specifically related to `ProductDTO`.
Also, rather than using too many extension functions, try to implement them as methods inside the class whenever possible. 🙂

src/main/kotlin/ecommerce/util/Utils.kt
```diff
@@ -0,0 +1,42 @@
...
+
+data class ProductResponse(
```
Wordbe: Is there a specific reason why a `data class` is needed here? 🤔

src/main/kotlin/ecommerce/util/Utils.kt
```diff
@@ -0,0 +1,42 @@
...
+    val id: Long?,
+    val name: String,
+    val price: Double,
+    val quantity: Int,
+    val imageUrl: String,
+    val productOptions: List<ProductOptionResponse> = emptyList(),
+)
+
+data class ProductOptionResponse(
```
Wordbe: Same here.

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.get
+import org.springframework.transaction.annotation.Transactional
+
+@SpringBootTest
+@AutoConfigureMockMvc
+@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
+@ActiveProfiles("test")
+@Transactional
+class CartStatisticsControllerTest() {
```
Wordbe: Could you explain the reason for using each annotation? 🤔

Wordbe (changes requested):
Hello Apoorva!

You did a great job completing the Spring ecommerce order mission.
Well done!
You introduced JPA and refined the existing service to work smoothly.
I’ve left some review comments. Please check them and apply any that you can.

Thank you! 😊

build.gradle.kts
Apoorva-S-Natesh: Great Catch! will clean that up

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
Apoorva-S-Natesh: Thank you for pointing that out! It was a good opportunity to clean the code and remove repeating code.  I've moved the error messages into the exception classes themselves with default values but can still be overridden with custom messages when needed
The global handler is now much cleaner and simply uses e.message!! since every exception is guaranteed to have a message.

src/main/kotlin/ecommerce/service/ProductOptionService.kt
Apoorva-S-Natesh: Thank you that leverages Kotlin's null safety features properly rather than java style.

src/main/kotlin/ecommerce/model/CartItem.kt
Apoorva-S-Natesh: Upon closer inspection, this method might be introducing unnecessary complexity. Since we're only updating two properties, a direct assignment or potentially using apply could be a simpler and more efficient alternative. Let me know if I'm on the right track with this.

src/main/kotlin/ecommerce/model/Member.kt
Apoorva-S-Natesh: All the `toString()` was added for debugging and printing while testing. Thank you for pointing it out, they have to be removed! Please suggest if there are better ways to debug these entities in the Tests

src/main/kotlin/ecommerce/model/Product.kt
Apoorva-S-Natesh: All the toString() was added for debugging and printing while testing. Thank you for pointing it out, they have to be removed!

src/main/kotlin/ecommerce/model/ProductOption.kt
Apoorva-S-Natesh: All the toString() was added for debugging and printing while testing. Thank you for pointing it out, they have to be removed!

src/main/kotlin/ecommerce/model/Cart.kt
Apoorva-S-Natesh: Thank you for pointing it out, this was not used but was required in MemberService to register a member. In the MemberService.register method, I was creating a new Cart and assigning it to member.cart, but was not setting the member reference in the Cart entity. This was leading inconsistencies in the relationship, potentially causing `org.springframework.dao.DataIntegrityViolationException caused by a Unique index or primary key violation`

Changes made:
- Create the Cart with the member reference set in the constructor.
- Save the Cart first to ensure it has a valid ID.
- Assign the saved Cart to member.cart before saving the Member.

src/main/kotlin/ecommerce/model/Cart.kt
Apoorva-S-Natesh: You are right! This constructor is not used is unnecessary as all fields in the primary constructor have default values or are nullable (member: Member? = null, cartItem: MutableList<CartItem> = mutableListOf(), quantity: Int = 0), newItemAddedAt: LocalDateTime = LocalDateTime.now())

I realized that a primary constructor with default values can act as a no-arg constructor if all parameters have default values

src/main/kotlin/ecommerce/model/Cart.kt
Apoorva-S-Natesh: All the toString() was added for debugging and printing while testing. Thank you for pointing it out, they have been removed! Please suggest if there are better ways to debug these entities in the Tests

src/main/kotlin/ecommerce/controller/CartController.kt
Apoorva-S-Natesh: Thank you for your feedback on the CartController endpoint. I agree that using` {userId}` as a path variable could be confusing, as it might imply a Cart ID rather than filtering by a user’s ID. To address this and align with REST API conventions, I’ve updated the endpoint from `GET /api/carts/{userId}` to `GET /api/carts?userId={userId}.` The userId is now passed as a query parameter, making it clear that we’re filtering the carts resource by the user’s ID.

src/main/kotlin/ecommerce/controller/CartController.kt
Apoorva-S-Natesh: I was thinking since the `CartId` was being accessed, it would be appropriate to be in `CartController` but now when I think about it, it is fetching all the items from `CartItem` Repository so it would be better to be in `CartItemController `. have moved it to CartItemContoller

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
Apoorva-S-Natesh: **@SpringBootTest** - used to bootstrap the entire Spring application context for integration testing. It loads all the beans, configurations, and components simulating the runtime environment.

Loads the CartStatisticsController, tokenService, and any services/repositories needed to handle the /admin/analytics/top-products request.

**@AutoConfigureMockMvc** - This configures and autowires a MockMvc instance, which is used to perform HTTP requests against Spring MVC controllers without starting a full HTTP server.

 Provides the mockMvc instance used to send the GET request.

@AutoConfigureTestDatabase(replace = Replace.NONE): This annotation configures the test database for the test that use a real database. The replace = Replace.NONE setting indicates that the test should use the application’s configured database.

@ActiveProfiles("test") - This activates a specific Spring profile for the test, allowing us to load profile-specific configuration files (e.g., application-test.properties or application-test.yml).

while testing we had the secret key stored in application-test.yml so used it. But it seems unnecessary.

@Transactional: Ensures any database changes when the endpoint modifies data are rolled back, keeping the database clean.

This test no more uses @ActiveProfiles and @AutoConfigureTestDatabase and can be removed.

src/main/kotlin/ecommerce/service/PasswordService.kt
Apoorva-S-Natesh: Implemented BCrypt to remove the hardcoded values

src/main/kotlin/ecommerce/model/Member.kt
Apoorva-S-Natesh: I agree that this is redundant. I have removed it. Thank you for catching that.

src/main/kotlin/ecommerce/model/ProductOption.kt
Apoorva-S-Natesh: Thank you!

src/main/kotlin/ecommerce/dto/ProductDtos.kt
Apoorva-S-Natesh: You're right to question this! we used data classes mainly for DTOs because they are primarily used to transfer data and are designed to be immutable which ensures data received from the client remains constant throughout and they don't contain any business logic or any complex behavior. They're easy to work with.

The DTOs use Jakarta Bean Validation annotations (e.g., @NotNull, @Size, @Min, @Pattern) to enforce constraints on fields like name, price, quantity, and imageUrl. Data classes work seamlessly with these annotations in Spring, as they are simple POJOs that Jackson can deserialize into. The generated equals() and toString() methods were used for debugging.

Regarding whether data class is necessary, I acknowledge that a regular class could be used since we’re not heavily relying on the generated copy() or component functions, we could refactor to regular classes.

- Is this more about code consistency across the project, or performance concerns?
- Are there any specific downsides you're concerned about with using data classes here?

src/main/kotlin/ecommerce/util/Utils.kt
Apoorva-S-Natesh: Good catch! Since we're not using data class features, we could refactor to regular class

src/main/kotlin/ecommerce/util/Utils.kt
Apoorva-S-Natesh: You are right. Since we're not using data class features, we could refactor to regular class

src/main/kotlin/ecommerce/dto/ProductDtos.kt
Apoorva-S-Natesh: I have given the reasons [ here ](https://github.com/herotechcourse/spring-ecommerce-order/pull/17#discussion_r2260974470)😊. Also have refactored to regular classes since we are currently not leveraging their main benefits

src/main/kotlin/ecommerce/dto/ProductDtos.kt
Apoorva-S-Natesh: I have given the reasons [ here ](https://github.com/herotechcourse/spring-ecommerce-order/pull/17#discussion_r2260974470)😊. Also have refactored to regular classes since we are currently not leveraging their main benefits

src/main/kotlin/ecommerce/model/Cart.kt
Apoorva-S-Natesh: Thank you for the suggestion 🙂! I noticed that other entities were also being exposed, so I went ahead and implemented DTOs for all entity responses to ensure consistency and better encapsulation.

src/main/kotlin/ecommerce/model/CartItem.kt
Apoorva-S-Natesh: I agree toString() method should be deleted. Thank you for the suggestion 🙂! I noticed that other entities were also being exposed, so I went ahead and implemented DTOs for all entity responses to ensure consistency and better encapsulation.

**Why DTOs Matter:**
1. Ensure incoming data meets requirements (Validation)
2. Control exactly what data is exposed (Security)
3. Send only necessary data over the network (Performance)
4. Change internal structure without breaking APIs (Maintainability)

I have learnt that DTO should be used in these places, please correct me
- API endpoints (always)
- Cross-service communication
- Frontend-backend data transfer
- Data validation requirements
- Complex data transformations

src/main/kotlin/ecommerce/model/ProductOption.kt
Apoorva-S-Natesh: Thank you for your question about choosing` FetchType.EAGER` for the `ProductOption.product` relationship. We opted for **EAGER** because a **ProductOption** (e.g., "Red" for a T-shirt) is meaningless without its associated Product, which provides critical context like the product name and price. In our application, ProductOption is typically used in scenarios where the Product data is required, such as displaying cart items or product details in the UI. Eager fetching ensures that the Product is always available, simplifying business logic and avoiding `LazyInitializationException`.
For the` Member.cart` relationship, we chose `FetchType.LAZY` because a `Member` can be loaded without its Cart in many cases, such as user profile views or authentication. This reduces unnecessary database queries, and we use MemberResponse with cartId to avoid serializing the Cart object, mitigating serialization issues.

I understand your concern that EAGER fetching for ProductOption.product triggers a join every time a ProductOption is queried, which could impact performance. To address this, we could switch to LAZY fetching.

Should we default to LAZY fetching for all relationships?

src/main/kotlin/ecommerce/repository/CartRepository.kt
Apoorva-S-Natesh: Some unnecessary JPQL queries are removed. Thank you for suggesting the addition of a memberId field to the Cart entity to optimize queries by memberId. After evaluating the suggestion, we believe the current @OneToOne relationship from Cart to Member (with member.id) is sufficient for our use case. The Cart.member mapping allows us to query carts by memberId using JPQL (e.g., SELECT c FROM Cart c WHERE c.member.id = :memberId), which leverages the existing relationship without introducing data redundancy. Adding a memberId field would duplicate data already stored in Member.id, potentially leading to consistency issues if Cart.member and Cart.memberId get out of sync, requiring additional logic to maintain them.

I am curious about your thoughts on this.

src/main/kotlin/ecommerce/service/MemberService.kt
Apoorva-S-Natesh: Initially:
```
val cart = Cart(member = member)
        member.cart = cartRepository.save(cart)
        val savedMember = memberRepository.save(member)
        return tokenService.generateToken(savedMember)
```

Change:
```
fun assignCart(cart: Cart) {
        if (cart.member != this) throw IllegalArgumentException("Cart must be associated with this member")
        this.cart = cart
    }
```

```
val savedCart = Cart(member = member)
        member.assignCart(savedCart)
        val savedMember = memberRepository.save(member)
```
To address your suggestion, I propose adding an explicit method to the Member entity, such as assignCart(cart: Cart), to handle cart assignment and enforce business rules (e.g., ensuring cart.member matches this member). This would make the operation more expressive and maintainable, aligning with our focus on a rich domain model.

src/main/kotlin/ecommerce/util/Utils.kt
Apoorva-S-Natesh: That’s a helpful observation, thank you! Just curious, are there any notable downsides to using extension functions in this context? I’d love to understand the trade-offs better.

src/main/kotlin/ecommerce/service/CartItemService.kt
Apoorva-S-Natesh: I have moved the private methods to the domain model (CartItem and ProductOption) while keeping CartItemService responsible for orchestration and persistence.

Apoorva-S-Natesh:
Thank you for your insightful and valuable feedback throughout the review process. Your suggestions have significantly improved the quality and maintainability of the codebase. Specifically, your recommendation to use a CartResponse DTO and move business logic (e.g., updateExistingCartItem, createNewCartItem) to the domain model enhanced cohesion in CartItem, Cart, and ProductOption, aligning with Domain-Driven Design principles. 
These changes have made the code more robust. Your feedback not only addressed immediate issues but also guided me toward better design practices. Please let me know if there are any adjustments or additional areas, I should focus on to meet the requirements. I truly appreciate your thorough guidance and look forward to applying these principles in future iterations.

src/main/kotlin/ecommerce/model/Cart.kt
Wordbe: I understand your intention. However, since the primary constructor already has a named argument, it can fully serve this purpose. Therefore, I think it can be removed. 🙂

src/main/kotlin/ecommerce/model/Cart.kt
Wordbe: Good Job!

src/main/kotlin/ecommerce/model/Cart.kt
```diff
@@ -0,0 +1,52 @@
...
+    )
+
+    fun toResponse() =
+        CartResponse(
+            id = id ?: throw IllegalStateException("Cart ID cannot be null"),
+            quantity = quantity,
+            newItemAddedAt = newItemAddedAt,
+            memberId = member?.id,
+            cartItem.map { it.toResponse() },
+        )
```
Wordbe: Make sure the domain model does not know about (or depend on) the DTO.
In a layered architecture, dependencies are usually one-directional to avoid circular references, but this seems to be in the opposite direction. 🙂

src/main/kotlin/ecommerce/controller/CartController.kt
Wordbe: Good. Thank you.

src/main/kotlin/ecommerce/dto/ProductDtos.kt
Wordbe: It’s both. I’d like to have a single rule so that the code remains consistent overall, and even though performance would be similar without using a `data class`, I don’t think we need to use something that isn’t necessary. 🙂

src/main/kotlin/ecommerce/model/CartItem.kt
Wordbe: Scope functions like `apply` are fine too. However, from an object-oriented perspective, I think it’s also good to give the domain model responsibilities through methods like this. 🙂

src/main/kotlin/ecommerce/model/CartItem.kt
Wordbe: Yes, as you mentioned, it is commonly used. You also identified the reason for using it well. Thank you. 🙂

src/main/kotlin/ecommerce/model/Member.kt
Wordbe: If you want to log all field values, implementing `toString()` can be convenient. Otherwise, you can extract and log only specific fields separately! 😊

src/main/kotlin/ecommerce/model/ProductOption.kt
Wordbe: Sounds good. If you’re aware that the query always runs only twice and it’s actually fine in practice, then I think it’s okay to use it as is. 
You could also use something like `EntityGraph` to join `Product` and `ProductOption`. However, in this case, running two queries doesn’t seem excessive, so it looks fine to leave it as is. 🙂

src/main/kotlin/ecommerce/repository/CartRepository.kt
Wordbe: Yes, you applied the approach without using `memberId`. I agree with the points you mentioned. 😊

src/main/kotlin/ecommerce/util/Utils.kt
Wordbe: Extension functions are mainly used to add convenient features to classes we can’t modify, like those in the standard library, without overusing them. They’re also useful when implementing Kotlin DSLs.
In the code above, whether you implement it as a method or an extension function doesn’t make much difference, but I see inconsistency across the whole codebase. When do you plan to use methods, and when will you use extension functions?
Also, if extension functions with the same receiver are scattered across different files, they can be hard to find. On the other hand, placing them as methods inside the object keeps state and behavior together, making maintenance easier. 🙂

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private fun createAdminToken(): String {
+        val adminMember = Member("admin@test.com", "password", "Admin User", Role.ADMIN, null, 1L)
+        return tokenService.generateToken(adminMember)
+    }
+
+    private fun createUserToken(): String {
+        val userMember = Member("user@test.com", "password", "Regular User", Role.USER, null, 2L)
+        return tokenService.generateToken(userMember)
+    }
```
Wordbe: This is not being used!

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.MediaType
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.get
+import org.springframework.transaction.annotation.Transactional
+
+@SpringBootTest
+@AutoConfigureMockMvc
+@Transactional
```
Wordbe: You are just using 'GET' method, so there is no data changed here. 🙂

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
Wordbe: Thank you for clearly explaining each annotation and removing the unnecessary ones! 😊

Wordbe (changes requested):
Hello Apoorva!

Thank you for carefully reading my review, responding thoroughly, and reflecting the feedback.
I’ve added some answers to your questions from the review along with a few additional comments.
You did such a great job that there might not be many changes needed.
Please take a look and apply any possible updates. Thank you! 🙂

src/main/kotlin/ecommerce/model/Cart.kt
Apoorva-S-Natesh: You are right! This constructor is also not used and is unnecessary as all fields in the primary constructor have default values. Have removed it

src/main/kotlin/ecommerce/dto/ProductDtos.kt
Apoorva-S-Natesh: I agree with you!

src/main/kotlin/ecommerce/model/Member.kt
Apoorva-S-Natesh: Got it, thank you 😊

src/main/kotlin/ecommerce/model/ProductOption.kt
Apoorva-S-Natesh: Thank you for suggesting` EntityGraph`, it's a good way to improve the runtime performance.

src/main/kotlin/ecommerce/util/Utils.kt
Apoorva-S-Natesh: Thank you! That is very clear explanation, and I understand when it should be used. I have removed the extension functions to increase code consistency as a function in class also does the same and these classes are not class we cannot modify.

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
Apoorva-S-Natesh: Good catch! This Test is testing `CartStatistics` and user token is not required at all. They have been removed

src/test/kotlin/ecommerce/controller/CartStatisticsControllerTest.kt
Apoorva-S-Natesh: It was an oversight! Since all the tests used `@Transactional` used it here as well without noticing that only `GET` is being used. I need to be more cautious on the annotations I use. Thank you, it has been removed.

src/main/kotlin/ecommerce/model/Cart.kt
Apoorva-S-Natesh: wow that's an interesting problem! It's something I hadn't thought about.
This Violates layered architecture, mixes responsibilities, also the core business logic (Cart) depends on the outer layer (DTO), creating tight coupling and risk of circular dependencies.

Fix:
I have moved the toResponse() logic out of the domain entity and into a mapper Object to ensure Clear separation of concerns

Apoorva-S-Natesh:
Thank you for your insightful and valuable feedback. The Layered architecture and circular dependency were really important to learn.  I also understand the importance of having consistency in code. I have carefully tried to implement the feedback. Thank you for your time.

Wordbe (approved):
Hello Apoorva!

Thank you for reflecting the review well.
Through this step of the mission, it seems you’ve gained a solid understanding of how to work with JPA in Spring, write domain logic, and create tests.
See you in the next step, and great work! 😊

PR #19, opened 2025-08-11
[spring-ecommerce-order - Step-2] Submit deniz-oezdemir's mission

deniz-oezdemir:
Hi there 👋

I hope you are doing well!

This pull request introduces the complete feature for placing an order and deploying the application to an AWS EC2 instance. This PR also resolves all open feedback from the previous [review](https://github.com/herotechcourse/spring-ecommerce-order/pull/8).

A Note on the deployment: I found the EC2 instance not to have enough resources to build the executable, so I've build it locally and transferred the JAR file instead. I added a [README for deployment](https://github.com/deniz-oezdemir/spring-ecommerce-order/blob/step2/README-Deployment.md) with a step-by-step explanation on how to do it.

I look forward to your feedback.

Thank you,
Deniz


Wordbe (changes requested):
The draft label is still on. May I start the review now? 🙂

deniz-oezdemir:
Hi @Wordbe, yes, I had forgotten to press the `Ready for review` button. Now it's ready for review, thanks so much for your time in advance! 🙏

deploy.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# Find the new JAR file in the build directory
+BUILD_PATH=$(ls ./build/libs/*.jar)
+JAR_NAME=$(basename $BUILD_PATH)
```
Wordbe: When you run `ls ./build/libs/*.jar`, multiple jar files appear. I think we should consider which one is actually needed and make sure only a single file is shown. Also, how about figuring out when `jar` and `plain.jar` are created and what they’re used for? 🙂

settings.gradle.kts
```diff
@@ -1 +1,3 @@
 rootProject.name = "spring-ecommerce"
+
+include("src:test:resources")
```
Wordbe: Why is this needed when it seems unnecessary? 🙂

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
```diff
@@ -139,6 +140,18 @@ class ApiErrorControllerAdvice {
         return buildResponse(apiError)
     }
 
+    @ExceptionHandler(PaymentFailedException::class)
+    fun handlePaymentFailedException(e: PaymentFailedException): ResponseEntity<ApiError> {
+        val errorMessage = e.message ?: "Payment processing failed."
+        log.warn("PaymentFailedException: $errorMessage", e)
```
Wordbe: How about defining the default value inside the exception constructor?
Also, I don’t think you need to explicitly print the error message when logging—if you pass `e` as a parameter, the exception message is already printed, so it’s being duplicated. 🙂

src/main/kotlin/ecommerce/config/StripeProperties.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties("stripe")
+data class StripeProperties(
```
Wordbe: Do we really need a data class? 🤔

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
```diff
@@ -23,4 +25,13 @@ class WebMvcConfiguration(
...
         resolvers.add(loginMemberArgumentResolver)
     }
+
+    override fun addCorsMappings(registry: CorsRegistry) {
+        registry.addMapping("/api/**")
+            .allowedOrigins("*") // TODO: replace with specific domain of my frontend for production
+            .allowedMethods("GET", "POST", "PUT", "DELETE", "OPTIONS")
+            .allowedHeaders("*")
+            .exposedHeaders(HttpHeaders.LOCATION)
+            .maxAge(3600)
```
Wordbe, line 35: Can you tell me why this was configured? 🙂

src/main/kotlin/ecommerce/config/interceptor/CheckLoginInterceptor.kt
```diff
@@ -17,6 +17,10 @@ class CheckLoginInterceptor(private val jwtTokenProvider: JwtTokenProvider) : Ha
         response: HttpServletResponse,
         handler: Any,
     ): Boolean {
+        if (request.method == "OPTIONS") {
+            return true
+        }
```
Wordbe: Try using the `HttpMethod` object.
Also, is there a reason why OPTIONS is not handled by the interceptor? 🙂

src/main/kotlin/ecommerce/entities/Payment.kt
```diff
@@ -0,0 +1,20 @@
...
+import jakarta.persistence.Table
+
+@Entity
+@Table(name = "payment")
+class Payment(
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long? = null,
+    @Column(nullable = false)
+    val amount: Long,
```
Wordbe, line 17: `Long` type has drawbacks like not being able to represent decimals and possible calculation errors. Shall we think it over? 🤔

src/main/kotlin/ecommerce/infrastructure/StripeClient.kt
```diff
@@ -0,0 +1,43 @@
...
+    private val stripeProperties: StripeProperties,
+) {
+    private val restClient = RestClient.create()
+
+    fun createPaymentIntent(req: StripePaymentRequest): StripePaymentIntentResponse {
+        val body =
+            listOf(
+                "amount=${req.amount}",
+                "currency=${req.currency}",
+                "payment_method=${req.paymentMethod}",
```
Wordbe, line 28: It’s better to validate required values!

src/main/kotlin/ecommerce/infrastructure/StripeClient.kt
```diff
@@ -0,0 +1,43 @@
...
+                    .header(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secretKey}")
+                    .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                    .body(body)
+                    .retrieve()
+                    .toEntity(StripePaymentIntentResponse::class.java)
+
+            return response.body!!
+        } catch (e: Exception) {
+            throw IllegalArgumentException("Stripe API error: ${e.message}")
+        }
```
Wordbe: There could be various reasons why the API request failed, so it seems difficult to simply conclude it’s an `IllegalArgumentException.` 🧐

src/main/kotlin/ecommerce/mappers/OrderMapper.kt
```diff
@@ -0,0 +1,24 @@
...
+        price = this.price,
+        quantity = this.quantity,
+    )
+
+fun Order.toDTO(): OrderResponseDTO =
+    OrderResponseDTO(
+        orderId = this.id!!,
+        orderDate = this.orderDate,
+        orderStatus = this.status,
+        totalAmount = this.payment.amount / 100.0,
```
Wordbe: Try converting the magic numbers into constants. 🔢

src/main/kotlin/ecommerce/model/StripePaymentIntentResponse.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.model
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+class StripePaymentIntentResponse(
+    @JsonProperty("id")
+    val id: String,
```
Wordbe: `@JsonProperty` seems unnecessary because the names are the same.

src/main/kotlin/ecommerce/repositories/OptionRepository.kt
```diff
@@ -1,6 +1,15 @@
...
+import org.springframework.data.jpa.repository.Lock
+import org.springframework.data.jpa.repository.Query
+import java.util.Optional
 
-interface OptionRepository : JpaRepository<Option, Long>
+interface OptionRepository : JpaRepository<Option, Long> {
+    // Find an option by its ID and apply a pessimistic write lock preventing other transactions from modifying it until this one is complete.
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @Query("SELECT o FROM Option o WHERE o.id = :id")
+    fun findByIdWithLock(id: Long): Optional<Option>
```
Wordbe, line 14: Wow great! you even considered concurrency. Shall we think about the trade-offs of doing it this way? ⚖️

src/main/kotlin/ecommerce/services/CartItemService.kt
```diff
@@ -1,21 +1,105 @@
...
     fun addOrUpdate(
         cartItemRequestDTO: CartItemRequestDTO,
         member: MemberDTO,
-    ): CartItemResponseDTO
+    ): CartItemResponseDTO {
+        validateProductExists(cartItemRequestDTO.productId)
 
-    fun findByMember(memberId: Long): List<CartItemResponseDTO>
+        val cartItem =
+            if (!cartItemRepository.existsByProductIdAndMemberId(cartItemRequestDTO.productId, member.id!!)) {
```
Wordbe: The downside of update is that two queries—exist and find—are executed, but I think using just find might be okay. 🙂

src/main/kotlin/ecommerce/services/CartItemService.kt
```diff
@@ -1,21 +1,105 @@
...
+        cartItemRequestDTO: CartItemRequestDTO,
+        member: MemberDTO,
+    ): CartItem {
+        val product = productRepository.findByIdOrNull(cartItemRequestDTO.productId)
+        if (product == null) throw OperationFailedException("Invalid Product Id ${cartItemRequestDTO.productId}")
+        return cartItemRepository.save(
+            CartItem(
+                member = member.toEntity(),
+                product = product,
+                quantity = if (cartItemRequestDTO.quantity == 0) 1 else cartItemRequestDTO.quantity,
```
Wordbe: I think setting the default value to 1 would avoid any misunderstandings.

src/main/kotlin/ecommerce/services/OrderService.kt
```diff
@@ -0,0 +1,94 @@
...
+                "eur",
+                req.paymentMethod,
+            )
+
+        val stripeResponse =
+            try {
+                stripeClient.createPaymentIntent(stripeRequest)
+            } catch (e: IllegalArgumentException) {
+                throw PaymentFailedException(e.message ?: "Payment failed due to an unknown error.")
+            }
```
Wordbe: It would be better to have a separate service for this.
Also, if error handling and validation that might occur here are inside it, it could be safer. 🙂
And it would be good to insert logs that can provide hints when a problem occurs.

src/main/kotlin/ecommerce/services/OrderService.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        val payment =
+            Payment(
+                amount = amountInCents,
+                stripePaymentId = stripeResponse?.id ?: "pi_error_id_not_found",
```
Wordbe: It would be good to define what kind of error this is. 🛑

src/test/kotlin/ecommerce/endtoend/OrderE2ETest.kt
```diff
@@ -0,0 +1,193 @@
...
+        assertThat(orderResponse.stripePaymentId).startsWith("pi_")
+
+        // verify database state
+        val updatedOption = optionRepository.findById(optionToPurchase.id!!).get()
+        assertThat(updatedOption.quantity).isEqualTo(initialStock - 1)
+
+        val orderInDb = orderRepository.findAll().firstOrNull()
+        assertThat(orderInDb).isNotNull
+        assertThat(orderInDb!!.member.id).isEqualTo(member.id)
+    }
```
Wordbe: Every time you test, an actual payment might be processed. How about using a stub to prevent that?
Also, my tests fail, but can we set it up so that tests can pass in some environments? 🫠🧐

start.sh
```diff
@@ -0,0 +1,19 @@
...
+then
+  echo "> No running application found."
+else
+  echo "> Stopping running application. PID: $CURRENT_PID"
+  kill -15 $CURRENT_PID
+  sleep 5
+fi
+
+echo "> Starting application..."
+nohup java -jar $DEPLOY_PATH$JAR_NAME > ./application.log 2>&1 &
```
Wordbe, line 19: What’s the difference from deploy.sh? Which one do you use for deployment? 🤔

Wordbe (changes requested):
Hello Deniz!
You created the code according to the mission requirements and submitted it well.
It must feel rewarding to have actually integrated an external API and completed the deployment.

I’ve reviewed some points worth considering,
so please check the review and apply the parts that are possible.
Thank you! 😊

deploy.sh
deniz-oezdemir: That's a great point, thank you. 🙏

I've looked into it and now have a clear understanding of the difference between the two JARs:
- `app.jar` is the executable "fat JAR". It contains the application code, all its dependencies, and the embedded web server. This is the self-contained file one runs with java -jar.
- `app-plain.jar` is a standard "thin JAR". It contains only the application's code and nothing else. It cannot be run by itself.

I've updated the script to use `grep -v 'plain'` to filter out the standard JAR, so it now reliably selects the correct executable fat JAR for deployment. This makes the script more robust.

settings.gradle.kts
deniz-oezdemir: You're right, that line was unnecessary. I've removed it to clean up the build configuration. Thank you for catching that. 👍

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
deniz-oezdemir: That was a great suggestion for making the error handling cleaner.

I've moved the default message into the `PaymentFailedException` constructor and simplified the log statement as you recommended. I found this pattern made the code much more consistent, so I applied the same logic to all the other custom exceptions and their handlers as well.

Thank you! 🙏

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
deniz-oezdemir: Of course. We configured these settings to meet the specific needs of our API:

- `.allowedMethods(...)`: This lists the HTTP methods that a frontend client is allowed to use when calling your API. We need GET (to list products), POST (to create orders), PUT/PATCH (to update), and DELETE. OPTIONS is required for the browser's preflight requests.
- `.allowedHeaders("*")`: This allows the client to send any HTTP headers in their request. This is important for headers like Authorization which contains the user's token.
- `.exposedHeaders(HttpHeaders.LOCATION)`: By default, browsers hide most response headers from JavaScript. The Location header is sent after creating a resource to provide its new URL. This line explicitly tells the browser to make that header visible to the client.
- `.maxAge(3600)`: This is a performance optimization. It tells the browser to cache the permissions from a preflight OPTIONS request for 3600 seconds (1 hour), so it doesn't need to send a preflight check before every single API call.

This setup ensures a web client can correctly interact with all features of our API.

While a frontend is not part of this specific mission, we might build one in our next team project. We are keeping this configuration in place to ensure the API is ready for that future requirement.

src/main/kotlin/ecommerce/config/StripeProperties.kt
deniz-oezdemir: You've raised a good point about consistency. Since we are not using the extra methods from a data class for this component yet, I've changed it to a regular class to keep it minimal for now. We can easily switch it back if we add specific tests or logging for it in the future. 👍

src/main/kotlin/ecommerce/config/interceptor/CheckLoginInterceptor.kt
deniz-oezdemir: Thanks for your suggestion! 🙏 

I've updated the code to use `HttpMethod.OPTIONS.name()` for a type-safe comparison.

And to answer your question: yes, we intentionally let OPTIONS requests pass through this interceptor. This is necessary to support CORS preflight requests from browsers, which are sent without an Authorization header and must be allowed for the actual API call to succeed.

src/main/kotlin/ecommerce/entities/Payment.kt
deniz-oezdemir: That's a great point about the importance of handling monetary values correctly.

I chose to use a `Long` to store the amount in the smallest currency unit (cents). This is a common pattern that avoids floating-point precision errors and is also the format that the Stripe API uses for amounts.

We handle the conversion back to euros and cents only in the `OrderMapper`, to ensure all internal calculations are done with safe, exact integer math. I believe this is a robust approach, but I am open to discussing `BigDecimal` if you think it's a better fit? 👀

src/main/kotlin/ecommerce/infrastructure/StripeClient.kt
deniz-oezdemir: That's a good idea for making the client more robust!

I've added require checks to the top of the createPaymentIntent method to validate the inputs before the network call. I also added a few new unit tests to confirm the validation is working as expected.

Thanks for the suggestion! 🙏

src/main/kotlin/ecommerce/infrastructure/StripeClient.kt
deniz-oezdemir: Yes, the generic exception was not very helpful.

I updated the StripeClient to catch the specific `RestClientException` and extract the detailed error message from Stripe's response body. I also added several new tests to verify that we correctly handle different card decline reasons. 😄

src/main/kotlin/ecommerce/mappers/OrderMapper.kt
deniz-oezdemir: Good catch, thank you! You're right, a named constant is much clearer.

I've defined CENTS_PER_EUR in the Payment entity and updated the mapper to use it. It reads a lot better now. Thanks again! 🙏

src/main/kotlin/ecommerce/model/StripePaymentIntentResponse.kt
deniz-oezdemir: You're right, that annotation wasn't needed. I've removed it to clean up the code. Thanks for spotting that! 👍

src/main/kotlin/ecommerce/repositories/OptionRepository.kt
deniz-oezdemir: Yes, I thought that we somehow need to make sure that there is enough stock to fulfill the users order (even if some other user is buying the same product simultaneously).

Your question about about the trade-offs is great!

We chose **pessimistic locking** here because it is the safest way to guarantee stock integrity and prevent overselling, which is critical during checkout.

The main **trade-off is performance**. We hold a database lock while waiting for the Stripe API to respond, which could block other users trying to buy the same item.

The alternative would be **optimistic locking**. This would offer better performance but would require us to build more complex logic to handle the cases where a conflict occurs and we need to inform the user that the item sold out while they were checking out.

For this critical feature, I felt the safety and simplicity of pessimistic locking was the right trade-off to start with.

src/main/kotlin/ecommerce/services/CartItemService.kt
deniz-oezdemir: That's a great performance tip, thank you! 🙏 

I've refactored the `addOrUpdate` method to use a single find query now and pass on the `existingItem` to the `handleUpdate` method, which removes the extra database call. 😄

src/main/kotlin/ecommerce/services/CartItemService.kt
deniz-oezdemir: I've updated the validation on the `CartItemRequestDTO` to require a positive number. This makes the business rule explicit and simplifies the service code. I appreciate the tip! 👍

src/main/kotlin/ecommerce/services/OrderService.kt
deniz-oezdemir: This is a great suggestion, thank you!

I've refactored the code and created a separate `PaymentService` now. It handles all the Stripe logic, errors, and logging. This really cleaned up the `OrderService` and makes the responsibilities much clearer.

I appreciate the guidance! 🙏

src/main/kotlin/ecommerce/services/OrderService.kt
deniz-oezdemir: I agree, silently saving an error is definitely not the way to go.

So I've updated the `PaymentService` to throw an exception if the Stripe response is empty. This is much safer and ensures our data stays clean. 🙏

src/test/kotlin/ecommerce/endtoend/OrderE2ETest.kt
deniz-oezdemir: Thank you for pointing this out! 🙏 

I've tried different things and settled on the following solution:

- Mocked Stripe Client: I've refactored `OrderE2ETest` and `StripeClientTest` to use a _mocked_ `StripeClient`. This removes the dependency on the live Stripe API. Now, ests now run much faster without network latency, we're no longer subject to network failures and we can now deterministically simulate both successful and failed payment scenarios.

- Test Suite Stabilization: To resolve the intermittent `PortInUseException` and data leakage issues, I've added `@DirtiesContext(classMode = ClassMode.AFTER_CLASS)` to our E2E test classes. This ensures each class runs in a completely isolated environment, guaranteeing consistent results. I'm aware of the performance impact and can track improving this as future technical debt.

- Setup Required: @Wordbe to run the tests, you'll need to create a local secrets file. The application requires the stripe.api.key property to be present on startup.
You can just copy the example file:
`cp src/test/resources/application-secrets.properties.example src/test/resources/application-secrets.properties`
An arbitrary string for the key is sufficient (e.g., stripe.api.key=test_key), as it won't be used by the mocked client.

start.sh
deniz-oezdemir: Thanks for the question! It highlighted that the deployment steps weren't clear. 👍

I've investigated the build process on the server and found that the EC2 instance has insufficient memory to compile the project, causing the build to fail.

To fix this, I've shifted to a "local build, remote deploy" workflow. The executable JAR is now built on a local machine and then securely transferred to the server. I've documented these steps in a new guide: [README for deployment](https://github.com/deniz-oezdemir/spring-ecommerce-order/blob/step2/README-Deployment.md).

Since the build process no longer happens on the server, the deploy.sh script (which was designed to handle a remote build) is now obsolete. The server only needs the start.sh script to run the application. To avoid confusion, I've removed deploy.sh from the repository.

deniz-oezdemir:
Hi @Wordbe ,

Thanks so much for your feedback. 🙏

I’ve incorporated your suggestions and explained my reasoning in each comment feed. I’m looking forward to the team project where we will build on this codebase.

This PR is ready for another review.

Thank you for your time,
Deniz

deploy.sh
Wordbe: You distinguished well between the fat jar and the plain jar, and also filtered them properly. Thank you! 🙏

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
Wordbe: Sounds good!

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
Wordbe: You explained it well, thank you! 🙌

src/main/kotlin/ecommerce/config/interceptor/CheckLoginInterceptor.kt
Wordbe: Good job!

src/main/kotlin/ecommerce/entities/Payment.kt
Wordbe: You used cents as the minimum unit—that should be fine. Thanks for the explanation. 🙂

src/main/kotlin/ecommerce/services/OrderService.kt
Wordbe: Great work. Thank you.

start.sh
Wordbe: Ah, thank you for the explanation. Since you deleted the file, we won’t get confused anymore. 🙂

src/test/kotlin/ecommerce/endtoend/OrderE2ETest.kt
Wordbe: Thank you for changing the test structure and showing me how to test it. Thanks to that, unnecessary real API calls are no longer made. That’s great! 🎉

src/main/kotlin/ecommerce/repositories/OptionRepository.kt
Wordbe: You explained well why you introduced pessimistic locking. Thank you. 🙏

Wordbe (approved):
Hello Deniz!
Great job working on the spring-ecommerce-order project. Based on the code you’ve written and the improvements you’ve made through the reviews, I think it’s ready to approve and merge.

I especially appreciate how you thoughtfully considered the trade-off questions I raised in the reviews and gave solid answers. If you continue approaching service development this way, I believe it will help you design excellent backend systems.

Well done on all your hard work, and congratulations on completing the mission. Thank you! 🎉

PR #20, opened 2025-08-11
[spring-ecommerce-order - Step-2] Submit AnnLvu's mission

AnnLvu:
# Implement "Place Order" and deploy application to AWS EC2

## Description
This PR implements the **"Place Order"** feature using the **Stripe Payment API** and deploys the application to **AWS EC2**.

## Changes include:
- ✅ Implemented order placement with Stripe Checkout Session creation.  
- ✅ Decreased stock of the selected product option after order placement.  
- ✅ Removed the ordered product from the user’s cart.  
- ✅ Handled payment approval errors with user-friendly messages (e.g., expired session, invalid payment method, insufficient balance).  
- ✅ Displayed order details, including:
  - Order date/time  
  - Status  
  - Purchased items  
  - Checkout session ID  
  - Payment amount  
- ✅ Built and deployed the Spring Boot application to AWS EC2 with public access.  

## Deployment details
- **Application URL:** [http://13.124.243.67/api/products](http://13.124.243.67/api/products)  

## Screenshots
Screenshot 2025-08-11 155902
Screenshot 2025-08-11 160029
 

## ❗️Important:
To test the project, you need to rename a .env_deploy file -> .env and add you value


.env_deploy
```diff
@@ -0,0 +1,4 @@
+JWT_SECRET=
+JWT_TIME=
+STRIPE_SECRET_KEY=
+APP_CORS_ALLOWED_ORIGINS=*
```
pkeugine, line 4: I see some new keywords in this mission!
What is `CORS`, and what does this `*` value mean here?

deploy.sh
```diff
@@ -0,0 +1,22 @@
...
+then
+  echo "> No running application found."
+else
+  echo "> Stopping running application. PID: $CURRENT_PID"
+  kill -15 $CURRENT_PID
+  sleep 5
+fi
+
+echo "> Starting application..."
+nohup java -jar $DEPLOY_PATH$JAR_NAME > ./application.log 2>&1 &
```
pkeugine, line 22: I see you've made a deploy script 😮 
I would like you to tell me what `nohup`, `2>&1` and the last `&` means here.
This is because these knowledge are important when it comes to deploying the application,
and they come out often in tech interviews :)

src/main/kotlin/ecommerce/config/CorsConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+                val tokens = allowedOriginsProp.split(",").map { it.trim() }.filter { it.isNotEmpty() }
+                if (tokens.size == 1 && tokens[0] == "*") {
+                    addAllowedOriginPattern("*")
+                } else {
+                    allowedOrigins = tokens
+                }
+
+                allowedMethods = listOf("GET", "POST", "PUT", "DELETE", "OPTIONS")
+                allowedHeaders = listOf("*")
+                exposedHeaders = listOf(HttpHeaders.LOCATION) // тест ждёт этот заголовок
```
pkeugine: Looks like you brought in a comment that is unnecessary as you brought in this logic.
Let's remove it, and from now on, make sure we match the code convention and remove unnecessary parts when we bring a logic from elsewhere.

For example, I think `cfg` can be written as `config` or `configuration` since we try to avoid shortening names.

I wouldn't say we should never copy and paste logic from outside, but maintaining it so that it feels like ours is important.

src/main/kotlin/ecommerce/extensions/OrderToDto.kt
```diff
@@ -0,0 +1,27 @@
...
+    )
+
+fun OrderItem.toDto(): OrderItemResponseDto =
+    OrderItemResponseDto(
+        productOption.id,
+        productOption.name,
+        quantity,
+        productOption.price,
+        productOption.price * quantity,
+    )
```
pkeugine: What are the differences of `OrderToDto` and `~Mapper`?
I want to know the reason behind the name difference!

src/main/kotlin/ecommerce/extensions/StripeErrorMessages.kt
```diff
@@ -0,0 +1,20 @@
...
+        "processing_error" -> "A processing error occurred. Please try again later."
+        "card_velocity_exceeded" -> "The card exceeded its transaction limit."
+
+        "requires_payment_method" -> "Please choose a different payment method."
+        "requires_action" -> "Additional authentication is required to complete the payment."
+        "canceled" -> "The payment session was canceled or expired."
+
+        null, "" -> "The payment was declined. Please try a different payment method."
+        else -> "The payment was declined ($this)."
+    }
```
pkeugine: It's interesting that you've made this `toUserFriendlyMessage` function! 👍 
However, I think having this as String's extension function and keeping it in `extensions` might not be so good in a long term maintenance point of view.

This is because
* When we try to manipulate String, this method will always show up as our option even though it's only for Stripe message handling
* Rather than having this in `extensions` package, it might be better to handle `stripe` related things in `stripe` related package.

This is my personal opinion, so it would be great to hear your's as well! 😄

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,141 @@
...
+    private fun loadUserById(userId: Long) =
+        userRepository.findById(userId)
+            .orElseThrow { IllegalArgumentException("Invalid user ID: $userId") }
+
+    private fun loadCartProductsForUser(user: User): List<CartProduct> {
+        val cart = user.cart ?: throw IllegalArgumentException("Cart not found for user ${user.id}")
+        val cartProducts = cartProductRepository.findByCart(cart)
+        if (cartProducts.isEmpty()) {
+            throw IllegalArgumentException("Cart is empty for user ${user.id}")
+        }
```
pkeugine: Let's try `require` where we can use it.
Check other places in the project as well!

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,141 @@
...
+            throw PaymentException(msg)
+        }
+
+        deductStockAndClearCart(user, optionQuantityList)
+
+        order = OrderMapper.applyPaid(order, payment.id, optionQuantityList)
+        order = orderRepository.save(order)
+
+        return PlaceOrderResponseDto(order.id, order.stripeSessionId)
+    }
```
pkeugine: Have you tried failing scenario tests here?
By test, I mean `checking database to see if the order information got persisted as we expected`.
These are the kinds of things we can't really check with the current integration tests, since they don't validate what's in the database.

src/main/resources/application.properties
```diff
@@ -10,3 +10,4 @@ security.jwt.token.expire-length=${JWT_TIME}
 spring.jpa.hibernate.ddl-auto=create
 spring.jpa.show-sql=true
 spring.jpa.properties.hibernate.format_sql=true
+stripe.secret-key=${STRIPE_SECRET}
```
pkeugine: Screenshot 2025-08-12 at 10 52 24 PM

I see a red sign that can be removed!
It would be great to automate this option in intelliJ.

.env_deploy
```diff
@@ -0,0 +1,4 @@
+JWT_SECRET=
+JWT_TIME=
+STRIPE_SECRET_KEY=
```
pkeugine, line 3: It would be great to have a Stripe secret key that you've created.
I understand that it's important not to expose it, but sending it via DM might be good enough. :)

pkeugine (changes requested):
Hello Ann, and welcome to the final step with a reviewer :)

You've done great,
and since we're already starting a team mission,
I tried to keep the review simple.
This way, it won't be too stressful to check and respond to the comments.

I kept it as `request changes` since there might be some points you'd like to change,
but I think we're good to go, and I'll try to merge this PR in the second review request.

Happy Hacking! 😄  😄  😄

.env_deploy
AnnLvu: `CORS` is a browser security mechanism that controls which external domains are allowed to make requests to an API.
`* `value means “allow all origins.” It is useful for local development, but in production it’s insecure, because it lets any website send requests to the API.
For now, since this is just a small project without production deployment, I used `*` for simplicity.

deploy.sh
AnnLvu: Thank you for your questions!

- `nohup`  - keeps the process running after the terminal is closed
- `2>&1` -  redirects error output (stderr) to normal output (stdout)
-  `&` - runs the process in the background

src/main/kotlin/ecommerce/config/CorsConfig.kt
AnnLvu: Sorry about that - I was still trying to understand this part of the code since it was new for me, so I left the comment in my language. I’ve now removed it and updated the naming to match our conventions. Thanks a lot for pointing it out!

src/main/kotlin/ecommerce/extensions/OrderToDto.kt
AnnLvu: You’re right - thanks for pointing this out. I’ve now renamed it  in OrderMapper, and state transitions are in OrderStateMapper for clarity.

Good question! The difference is mostly in style.

  - `Order.toDto()` is written as a Kotlin extension function.
  - `OrderMapper `is the more classical “mapper” object, common in Java/Spring projects

src/main/kotlin/ecommerce/extensions/StripeErrorMessages.kt
AnnLvu: I moved the function into the stripe package, so now it’s scoped only for Stripe-related logic.

src/main/kotlin/ecommerce/service/OrderService.kt
AnnLvu: Thanks for the suggestion! I replaced the if/throw with require and check in OrderService.

src/main/resources/application.properties
AnnLvu: Removed it!

src/main/kotlin/ecommerce/service/OrderService.kt
AnnLvu: Thanks for the suggestion - I added failing-scenario integration tests that assert the DB state. 
**Scenarios:** 

- `Stripe` returns a non-succeeded status - we expect `FAILED`, user-friendly failureReason, stripeSessionId set, and stock unchanged. 

- `StripeClient` throws `StripePaymentException` - we expect `FAIFLED`, user-friendly failureReason, empty stripeSessionId, and stock unchanged.

 **How:** 
`@SpringBootTest` + test-scoped `@TestConfiguration` with a `@Primary` mocked `StripeClient` (`mockito-kotlin`). After calling` placeOrder`, I read entities via repositories and assert persisted values.

.env_deploy
AnnLvu: Got it - I’ll send you the Stripe secret key via DM.

pkeugine (approved):
Hello Ann, and well done checking out the reviews!
It must've been challenging focusing on the team project and taking care of this mission at the same time. Nice work! 👍 

I think we can let this mission go now,
and I hope you do as well in your next project.
I wish you all the luck!

Happy Hacking! 🔥 🔥 🔥

PR #21, opened 2025-08-11
[spring-ecommerce-order - Step-2] Submit SebasNadu's mission

SebasNadu:
## PR Summary

This PR implements a full **Order & Payment** workflow with **Stripe integration**, adds **PostgreSQL** with **Liquibase**, sets up **Docker** configurations for both development and production environments, and finalizes deployment to **AWS EC2**.

This was my first time using Liquibase, so I’m curious to hear your thoughts on my implementation.
I used it as part of planning for the final project, along with Docker for the same reason.

For deployment, I ran into some challenges and had to deploy mroe "manually".
However, I’ve already created a GitHub Action inside example.github that’s ready for use with EC2 once we have the necessary permissions.
Right now, we don’t have permission to run it.

I’d like to know your opinion on the best way to deploy to EC2, as I want to start preparing for the final project.

---

#### **1. Stripe Integration**
- Created `StripeClient` to interact with Stripe's Payment Intent API using `RestClient`.
- Added `StripeProperties` with `@ConfigurationProperties` for managing Stripe credentials.
- Implemented `StripeService` to handle payment processing, checkout session creation, and error handling.
- Added configuration for `stripe.secret-key` in application properties.
- Integrated test card scenarios based on Stripe documentation to validate payment flows.

#### **2. Order Management**
- Implemented **Order** and **OrderItem** entities.
- Designed database relationships to link:
  - **Order** → multiple **OrderItems**
  - **Order** → associated **Payment** record
- Orders contain:
  - Order date and time
  - Order status
  - Purchased items list
  - Stripe Checkout Session ID
  - Payment amount
- Implemented service layer to:
  - Create orders upon successful payment intent
  - Store Stripe checkout session data in the database
  - Update order status based on payment result

#### **3. Payment Handling**
- Created `Payment` entity to store Stripe-related data separately from `Order`.
- Established **one-to-one relationship** between `Order` and `Payment`.
- Payment table stores:
  - Payment amount
  - Currency
  - Payment method
  - Stripe Payment Intent ID
  - Transaction status

#### **4. PostgreSQL Integration with Liquibase**
- Migrated from H2 to PostgreSQL for production.
- Added **Liquibase** for database change management:
  - Created `changelog-master.xml`
  - Added entity-specific changelog scripts for Orders, OrderItems, and Payments.
- Configured profiles for:
  - **dev** → local PostgreSQL container
  - **prod** → EC2-hosted PostgreSQL

#### **5. Docker Configuration**
- Created **Dockerfile** for production build:
  - Based on `eclipse-temurin:21-jdk-jammy`
  - Runs app as non-root `appuser`
- Created **docker-compose.dev.yml**:
  - Runs Spring app + PostgreSQL
  - Maps `.env` variables for local development
- Created **docker-compose.prod.yml**:
  - Runs Spring app + PostgreSQL + Nginx reverse proxy
  - Configured `ecommerce-network` bridge network
  - SSL support via mounted certs

#### **6. Deployment to AWS EC2**
- Built and saved the production Docker image locally.
- Transferred `.tar` file to EC2 instance.
- Loaded image on EC2 and ran using `docker-compose.prod.yml`.
- Nginx reverse proxy configured to route traffic to Spring Boot container.

---

### **Instructions to Access the App**
- **URL**: [https://3.35.139.129/](https://3.35.139.129/)  
  *(Self-signed certificate — your browser will show a security warning)*

---

### **Running the App Locally**
1. Copy `example.env` to `.env` and fill in your values.
2. To run with PostgreSQL locally:
   ```bash
   docker-compose -f docker-compose.dev.yml up -d
3. Access the application at:
 http://localhost:8080
 4. Or setup intellij to run the docker-compose but only the db container, change prod to the environment and you can use the db, in local.


.gitignore
```diff
@@ -38,3 +38,6 @@ out/
 
 ### Mac OS ###
 .DS_Store
+
+### Secrets ###
+.env
```
Hyeon9mak, line 43: 👍

certs/fullchain.pem
```diff
@@ -0,0 +1,20 @@
...
+QH3gfyTzW5wQSlAIaVVQD2IDWXGzEUOuH4IEIDa0X0Eso/1z6t74R5B/XKDtvgRZ
+0/x2FLrwAK/xneuH3KdcMhdoLCUFh+6SdQ2xKPtmBHwcqAx9LfLdtcUzt4oHNbN3
+fNZkCuaj/MoeMEGp534JMWkCAwEAATANBgkqhkiG9w0BAQsFAAOCAQEAvDmjmuCz
+s3T5Ur52Gl66Hx1yMI3Yvmbi2zc2k0plCx1n27cdysoSW9pNevOeoC+wXD6YFY6O
+saxaICLGZ1+ny0JOJVFjNGW2mTocxnr3qvvs4XgADxFRJjrsGZ2+IxD7QhhvwHsu
+6JHE8KbNuQyl2uWvg7++GMVouYpCdVaq8gBQAOKTbZn0YtPIxZFQsvUgSABlgzI+
+ZIPjNomA0I9tGJQk0xE9JHPF2cQQmlqz3yf4jt3z4b3+o1lmlzmQD8qFwjsgR1pj
+P2fIbjmu0NOZiUzjjGYkXHZ/NXbvTxQyEhpAg3Yxz8ErVtZJpCa2PR1X08uyfe2c
+OdfGonwOyQd8TA==
+-----END CERTIFICATE-----
```
Hyeon9mak: How about adding security-sensitive items to .gitignore and sending them to me separately? GitHub PRs or repositories can be easily compromised by outsiders.

docker-compose.prod.yml
```diff
@@ -0,0 +1,55 @@
...
+    environment:
+      SPRING_PROFILES_ACTIVE: prod
+      DB_HOST: db
+      DB_PORT: ${DB_PORT}
+      DB_NAME: ${DB_NAME}
+      DB_USERNAME: ${DB_USERNAME}
+      DB_PASSWORD: ${DB_PASSWORD}
+      JWT_SECRET_KEY: ${JWT_SECRET_KEY}
+      JWT_EXPIRATION_TIME: ${JWT_EXPIRATION_TIME}
+      STRIPE_SECRET_KEY: ${STRIPE_SECRET_KEY}
```
Hyeon9mak, line 15: If you send it to me via DM, I can run the tests locally!

example.github/workflows/deploy.yml
```diff
@@ -0,0 +1,91 @@
...
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
```
Hyeon9mak, line 17: auto deployment script 👍

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
```diff
@@ -23,4 +25,13 @@ class WebMvcConfiguration(
...
     }
+
+    override fun addCorsMappings(registry: CorsRegistry) {
+        registry.addMapping("/api/**")
+            .allowedOrigins("*")
+            .allowedMethods("GET", "POST", "PUT", "DELETE", "OPTIONS")
+            .allowedHeaders("*")
+            .exposedHeaders(HttpHeaders.LOCATION)
+            .maxAge(3600)
+    }
```
Hyeon9mak, line 36: 😄 👍

src/main/kotlin/ecommerce/config/security/JwtProperties.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.validation.annotation.Validated
+
+@Validated
+@ConfigurationProperties("security.jwt.token")
+class JwtProperties(
+    @field:NotBlank
+    val secretKey: String,
+    @field:Min(1000)
+    val expireLength: Long = 3600000,
+)
```
Hyeon9mak: second? nano? hours?
How about indicating the time unit through a comment or changing the variable name itself?

src/main/kotlin/ecommerce/config/stripe/StripeClientConfiguration.kt
```diff
@@ -0,0 +1,26 @@
...
+                .baseUrl("https://api.stripe.com")
+                .defaultHeader(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secretKey}")
+                .build()
+
+        val adapter = RestClientAdapter.create(restClient)
+        val factory = HttpServiceProxyFactory.builderFor(adapter).build()
+
+        return factory.createClient(StripeClient::class.java)
+    }
+}
```
Hyeon9mak, line 26: What would happen if the client side doesn’t provide a separate timeout?
And how can we prepare for that?

src/test/kotlin/ecommerce/integration/StripeServiceTest.kt
```diff
@@ -0,0 +1,75 @@
...
+                paymentMethod = "pm_card_visa",
+            )
+
+        val response = stripeService.createPaymentIntent(paymentRequest)
+
+        println(response)
+
+        assertNotNull(response)
+        assertEquals(1000.0, response.amountDecimal)
+        assertEquals("usd", response.currency?.lowercase())
```
Hyeon9mak: What problems might arise if we call the actual external client every time the test runs?
And in such cases, what kind of testing approach can we use to prevent this?

Furthermore, it’s important to consider that the key focus should be testing our internal logic,
not the behavior of the external client! 😄

Hyeon9mak (changes requested):
Hello Sebas, welcome to the final stage!
You’ve prepared the deployment process well.

If you send me the secret key values via DM, I’ll run the tests locally myself! 😄 
I've left few comments.

Free to ask anytime!

docker-compose.prod.yml
SebasNadu: Yes, sorry for that. I forgot to send you the keys!

certs/fullchain.pem
SebasNadu: Thanks for pointing this out. I’ve removed the certificate files from the repository and added the *.pem files to .gitignore so nothing inside will be tracked going forward. I’ll send you the required certs through slack.

src/main/kotlin/ecommerce/config/security/JwtProperties.kt
SebasNadu: Good point. The value is in milliseconds. I’ll rename the property to expireLengthMs so the unit is clear and avoid confusion.

src/main/kotlin/ecommerce/config/stripe/StripeClientConfiguration.kt
SebasNadu: I had several issues while trying to find a solution. I first tried using HttpClient5, but ran into many problems. Then I considered using the WebFlux client, but I didn’t want to add a whole library just for this feature. Finally, I found an alternative approach using only the config and created a test in the end-to-end admin suite for the /slow endpoint. Please let me know if this is the best approach or if there are better options.
```
spring.http.client.connect-timeout=5000
spring.http.client.read-timeout=10000
```

src/test/kotlin/ecommerce/integration/StripeServiceTest.kt
SebasNadu: Good point! I’ll replace the real Stripe API calls in this test with a mocked StripeService so we only verify our logic. We can keep a separate integration test profile for actual Stripe calls if needed.

src/main/kotlin/ecommerce/config/stripe/StripeClientConfiguration.kt
```diff
@@ -0,0 +1,26 @@
...
+@Configuration
+class StripeClientConfiguration {
+    @Bean
+    fun stripeClient(stripeProperties: StripeProperties): StripeClient {
+        val restClient =
+            RestClient.builder()
+                .baseUrl("https://api.stripe.com")
+                .defaultHeader(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secretKey}")
+                .build()
+
```
Hyeon9mak, line 20:
```suggestion
@Configuration
class StripeClientConfiguration {
    @Bean
     fun stripeClient(stripeProperties: StripeProperties): StripeClient {
        val requestFactory = SimpleClientHttpRequestFactory().apply {
            setConnectTimeout(Duration.ofMillis(500))
            setReadTimeout(Duration.ofSeconds(10))
        }

        val restClient =
            RestClient.builder()
                .requestFactory(requestFactory)
                .baseUrl("https://api.stripe.com")
                .defaultHeader(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secretKey}")
                .build()
```

also u can set timeout values in kotlin code!

Hyeon9mak: > I found an alternative approach using only the config and created a test in the end-to-end admin suite for the /slow endpoint. Please let me know if this is the best approach or if there are better options.

Testing timeouts is never easy. A common approach is to set up a separate API and run a server just for testing. You did a great job.

If the controller is only used for testing, another option is to place it under the test package instead of main. 😉

Hyeon9mak (approved):
Hi Sebas, Nice work to reflect feedbacks!

I’ve confirmed that it runs fine locally 👍 
Of course, there are many more ways we could improve it,
but as you know, Sebas, there are countless possibilities, and it’s hard to cover them all in this short time. 😉 

What really matters is gaining the experience of writing scripts and paying attention to security for a team project. 
Great job! 😄 👍 

I think we’re all set now. Have a wonderful weekend, and I hope you have fun with your team project!
Really appreciate all the effort you’ve put in!

PR #22, opened 2025-08-11
[spring-ecommerce-order - Step-2] Submit san-ghun's mission

san-ghun:
# Overview

This PR is an implementation of a simple HTTP API that imitating some features of eCommerce API, which include access to resources for product, member, cart item, cart statistics, option of product, and order. The application uses Spring Boot + JPA using Kotlin.

With step2 requirements, the implementation provide a feature to place an order using the "Stripe Payment Create API" with handling inventory of the ordered product, handling of third-party communication with Stripe API. Additionally, the requirement request to deploy the application in a Cloud instance properly.

## Requirements from the mission

- Implement the "Place Order" feature using the [Stripe Payment Create API](https://docs.stripe.com/api/payment_intents/create)
    - place an order with a **selected product option and quantity**:
        - The stock of the selected product option should be **decreased** accordingly.
    - If the ordered product exists in the user's **cart**, it should be **removed** from the cart.
    - If the payment approval API call fails, you must handle the error safely.
        - Ensure that the user is **clearly informed** of the failure reason.
        - In the case of a **order failure due to payment rejection**, display an appropriate error message indicating **why the payment failed** (e.g., expired session, invalid payment method, insufficient balance, etc.).

- Implement the "Orders" feature.
    - The following information must be visible in the "Orders" request:
        - Order date and time
        - Order status
        - Purchased items
        - Checkout session Id (issued by stripe)
        - Payment amount

- You must deploy your existing service and ensure it can interact with the client.
    - Write a deployment script to automate the deployment process.
    - Handle security issues when interacting with the client API.
        - For example, resolve issues caused when the server and client have different `Origin` values.
    - HTTPS is optional, but you should discuss with your team whether to apply it.

# Changes

## Implement "Place Order" feature using the third-party API

- Introduce a `RestClient` to communicate with the third-party API, **Stripe Payment Create API**
- Introduce a `ConfigurationProperties` for Stripe
- Introduce DTOs for the payment; `PaymentRequest` and `PaymentResponse`
- Introduce `.env` file to handle secret key --*\*additional comments are at the bottom*
- Introduce `OrderService` and `OrderPaymentService` to "Place Order"
- Add new exception to catch payment failure
- Add new exceptions to catch empty cart exception and insufficient stock to handle the order exception

## Implement "Orders" feature

- Introduce `Order` and `OrderItem` entities to handle order
- Introduce enum class `OrderStatus` to represent status of order
- Introduce DTOs to handle order request and response

## Prepare for deployment of the application

- Add `start.sh` file to auto start the application on deployment server
- Add `_deploy.sh` file to auto deploy from local machine to deployment server --*\*additional comments are at the bottom*
- Update `application.properties` file to configure port setting
- Add `application.properties` file in test directory to separate setting for testing
- Update configuration in `WebMvcConfiguration` to handle **CORS** 

## Testing

- To run test with environment variable for the stripe secret key, edited configuration for test in IDE manually.

1. Where to find editing configuration
Screenshot 2025-08-13 at 19 17 41

2. Where to put secret key as environment variable
Screenshot 2025-08-13 at 19 01 43

# Deployment

- Deployment has been conducted on AWS EC2 instance. 
- The original strategy to auto-deploy was like this,
    - git-clone the repository for targeted branch and compile the project to run the application.
- **However**, the capability of instance is unable to handle compilation of the application, so I managed to write `_deploy.sh` file to execute compilation and secure copy to deployment server instance and execute `start.sh` file in remote.
    - To make this possible, I added security credential information in `_deploy.sh` file. --*\*additional comments are at the bottom*

# Endpoints

**Order** - newly added

- `GET /api/orders` - get all orders of the login member
- `GET /api/orders/{orderId}` - get a order of the order id
- `POST /api/orders/place` - place an order

**Product**

- `POST /api/products` - create a product
- `GET /api/products` - read all products in `Page`
- `GET /api/products/{id}` - read a product with id
- `PUT /api/products/{id}` - update a product with id
- `DELETE /api/products/{id}` - delete a product with id

Web View

- `GET /` - provide view of list of all products
- `GET /add/product` - provide view of form to add new product
- `GET /update/product/{id}` - provide view of form to update a product

**Options for Product**
- `GET /api/products/{id}/options` - read all options for the product of the id

**Member**

- `POST /api/members/register` - register a member
- `POST /api/members/login` - login a member
- `GET /api/members/me` - get member data

**Cart Item**

- `GET /api/cart` - view user cart items in `Page`
- `POST /api/cart` - add item to user cart
- `DELETE /api/cart/{cartItemId}` - remove an item from user cart
- `PUT /api/cart/{cartItemId}` - update quantity of an item in user cart

**Cart Statistics**

- `GET /api/admin/cart-stats/top5-products` - get top 5 added products inn last 30 days
- `GET /api/admin/cart-stats/active-members` - get active members in last 7 days

# ER Diagram

```mermaid
erDiagram
    CART_ITEMS {
        bigint id PK
        integer quantity
        bigint cart_id FK
        timestamp created_at
        bigint member_id FK
        bigint option_id FK
        bigint product_id FK
    }
    
    CARTS {
        bigint id PK
    }

    MEMBERS {
        bigint id PK
        varchar email
        varchar password
        varchar role
        bigint cart_id FK
    }

    ORDER_ITEMS {
        bigint id PK
        integer quantity
        timestamp created_at
        bigint member_id FK
        bigint option_id FK
        bigint order_id FK
        bigint product_id FK
        timestamp updated_at
    }

    ORDERS {
        bigint id PK
        float payment_amount
        tinyint status
        timestamp created_at
        timestamp updated_at
        varchar checkout_session_id
        varchar payment_method
        bigint member_id FK
    }

    OPTIONS {
        bigint id PK
        integer quantity
        bigint product_id FK
        varchar name
    }

    PRODUCTS {
        bigint id PK
        float price
        varchar image_url
        varchar name
    }

    MEMBERS_ORDERS {
        bigint member_id FK
        bigint orders_id FK
    }

    CART_ITEMS }|--|| MEMBERS : "FKmyrsw3rg5yxntcqv45s8q6msh"
    CART_ITEMS }|--|| OPTIONS : "FK6puq1wt57hfl3i0prtunno0xt"
    CART_ITEMS }|--|| PRODUCTS : "FK1re40cjegsfvw58xrkdp6bac6"
    CART_ITEMS }|--|| CARTS : "FKpcttvuq4mxppo8sxggjtn5i2c"
    
    MEMBERS ||--|| CARTS : "FK9ir559ww1gdqpn7ey52h6h810"

    OPTIONS }|--|| PRODUCTS : "FK2m37jcoh1yt06dsivrphak718"

    ORDER_ITEMS }|--|| MEMBERS : "FKavyo7fsm7grgah9to6uu965s9"
    ORDER_ITEMS }|--|| OPTIONS : "FKei5ne8fxxpihl6vo3fee0j90w"
    ORDER_ITEMS }|--|| PRODUCTS : "FKocimc7dtr037rh4ls4l95nlfi"
    ORDER_ITEMS }|--|| ORDERS : "FKbioxgbv59vetrxe0ejfubep1w"

    ORDERS }|--|| MEMBERS : "FK2vq7lo4gkknrmghj3rqpqqg6s"

    MEMBERS_ORDERS ||--|| MEMBERS : "member_id"
    MEMBERS_ORDERS ||--|| ORDERS : "orders_id"

```

# Additional Note

For the security concern, I would like to send following information to you(reviewer) through DM in our Slack.
- Accessible URL
- `_deploy.sh` file

For the Stripe secret key, if you need my secret key please request me in DM or in comment. I will provide you through DM.
- `.env` file for Stripe secret key
    - Format is like this, `STRIPE_SECRET_KEY={key}`

I am looking forward to you review!
Thank you!


.gitignore
```diff
@@ -4,6 +4,8 @@ build/
 !gradle/wrapper/gradle-wrapper.jar
 !**/src/main/**/build/
 !**/src/test/**/build/
+.env
+_deploy.sh
```
Hyeon9mak, line 9: 👍

README.md
```diff
@@ -2,36 +2,91 @@
...
+            - [x] Allow all origins
+            - [x] Allow "simple" methods `GET`, `HEAD`, and `POST`
+            - [x] Allow all headers
+            - [x] set max age to 1800 seconds (30 minutes)
+        - [ ] login user
+        - [ ] login session 
+        - [ ] admin user
+        - [ ] admin control
+    - [ ] (optional) set up HTTPS
+
```
Hyeon9mak: They're done?

src/main/kotlin/ecommerce/WebMvcConfiguration.kt
```diff
@@ -21,4 +23,14 @@ class WebMvcConfiguration(
...
+
+    // TODO: assign 'allowedOrigins' to frontend address
+    override fun addCorsMappings(registry: CorsRegistry) {
+        registry.addMapping("/api/**")
+            .allowedOrigins("*")
+            .allowedMethods("HEAD", "GET", "POST", "PUT", "DELETE")
+            .allowedHeaders("*")
+            .exposedHeaders(HttpHeaders.LOCATION)
+            .maxAge(1800)
+    }
```
Hyeon9mak: TODO done? 😄

src/main/resources/data.sql
```diff
@@ -35,19 +35,23 @@ INSERT INTO members (email, password, role, cart_id) VALUES ( 'dan@htc.com', 'da
...
+INSERT INTO cart_items (member_id, product_id, option_id, quantity, created_at, cart_id) VALUES (1, 5, 1, 1, CURRENT_TIMESTAMP - INTERVAL '9' DAY, 1);
+INSERT INTO cart_items (member_id, product_id, option_id, quantity, created_at, cart_id) VALUES (1, 6, 1, 1, CURRENT_TIMESTAMP - INTERVAL '6' DAY, 1);
+INSERT INTO cart_items (member_id, product_id, option_id, quantity, created_at, cart_id) VALUES (2, 6, 1, 1, CURRENT_TIMESTAMP - INTERVAL '7' DAY, 2);
+INSERT INTO cart_items (member_id, product_id, option_id, quantity, created_at, cart_id) VALUES (3, 3, 2, 1, CURRENT_TIMESTAMP - INTERVAL '8' DAY, 3);
+INSERT INTO cart_items (member_id, product_id, option_id, quantity, created_at, cart_id) VALUES (1, 1, 1, 100, CURRENT_TIMESTAMP - INTERVAL '9' DAY, 1);
 ;
+-- INSERT INTO orders () VALUES ();
+;
+-- INSERT INTO order_items (member_id, product_id, option_id, quantity, created_at, order_id) VALUES (1, 1, 1, 1, CURRENT_TIMESTAMP - INTERVAL '4' DAY, 1);
+-- INSERT INTO order_items (member_id, product_id, option_id, quantity, created_at, order_id) VALUES (1, 2, 1, 1, CURRENT_TIMESTAMP - INTERVAL '2' DAY, 1);
```
Hyeon9mak: If testing is complete, please remove all comments, as they may cause confusion for fellow developers.

src/test/resources/application.properties
```diff
@@ -0,0 +1,22 @@
...
+spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.format_sql=true
+spring.jpa.hibernate.ddl-auto=create
+
+spring.jpa.defer-datasource-initialization=true
+
+logging.level.org.hibernate.orm.jdbc.bind=TRACE
+logging.level.org.springframework.beans.factory=debug
+
+stripe.secret-key=${STRIPE_SECRET_KEY}
```
Hyeon9mak, line 20: Now you can also hide the JWT secret key. 😄

src/test/kotlin/ecommerce/client/StripeClientTest.kt
```diff
@@ -0,0 +1,113 @@
...
+
+        val expected = "succeeded"
+        assertThat(actual).isNotNull()
+        assertThat(actual?.amount).isEqualTo(amount)
+        assertThat(actual?.`object`).isEqualTo("payment_intent")
+        assertThat(actual?.status).isEqualTo(expected)
+    }
+
+    @Test
+    fun `payment status failed with invalid payment method, card declined with Generic decline`() {
```
Hyeon9mak: What problems might arise if we call the actual external client every time the test runs?
And in such cases, what kind of testing approach can we use to prevent this?

Furthermore, it’s important to consider that the key focus should be testing our internal logic,
not the behavior of the external client! 😄

src/main/kotlin/ecommerce/controller/api/OrderController.kt
```diff
@@ -0,0 +1,79 @@
...
+        return ResponseEntity.badRequest().body(errorBody)
+    }
+
+    @ExceptionHandler(PaymentFailedException::class)
+    fun handlePaymentFailedException(e: Exception): ResponseEntity<Map<String, Any>> {
+        val error = mapOf("payment" to e.message)
+        val errorBody = mapOf("errors" to error)
+        println("PaymentFailedException occurred: $errorBody")
+        return ResponseEntity.status(HttpStatus.PAYMENT_REQUIRED).body(errorBody)
+    }
```
Hyeon9mak: What’s the difference between using a print statement and using a log? 🤔

src/main/kotlin/ecommerce/service/OrderPaymentService.kt
```diff
@@ -0,0 +1,27 @@
...
+    ): PaymentResponse {
+        val paymentRequest =
+            PaymentRequest(
+                amount = order.paymentAmount.toInt(),
+                currency = "usd",
+                paymentMethod = paymentMethod,
+            )
+        return stripeClient.createCheckoutSession(paymentRequest)
+            ?: throw PaymentFailedException("Payment failed with null")
+    }
```
Hyeon9mak: I’m curious why the payment method and currency unit were placed in the service layer instead of being treated as business objects. 👀

Hyeon9mak (changes requested):
Welcome to the last stage, San!

If you send me the secret key values via DM, I’ll run the tests locally myself! 😄
I've left some comments about ask ur idea. 😉 

Free to ask anytime!

README.md
san-ghun: Thanks for asking!
I am still searching for kinds of security issues I have to handle.
For now, I am going to update the list into this,

```markdown
    - [ ] handle security issue when interacting with the client API
        - [x] resolve CORS (Cross-Origin Resource Sharing)
            - [x] Allow all origins
            - [x] Allow "simple" methods `GET`, `HEAD`, and `POST`
            - [x] Allow all headers
            - [x] set max age to 1800 seconds (30 minutes)
        - [x] Authentication
            - [x] use JWT for stateless auth
            - [x] set validation of signature
            - [x] set expiration
            - [ ] ~~set rotating secret keys~~
        - [x] Authorization
            - [x] set validation for role-based access control
            - [x] set validation for members only available to access the resource they reference
```

Do you have your own list of security issues those must-dos?

src/main/kotlin/ecommerce/WebMvcConfiguration.kt
san-ghun: I think yes.
Since we don't have any domain address to assign, I think the "TODO" is not a matter anymore.
I am going to remove the comment.

src/main/resources/data.sql
san-ghun: Thanks for pointing out!
I will keep that in mind!

src/test/resources/application.properties
san-ghun: I am moving the JWT key into `.env` file.

src/main/kotlin/ecommerce/service/OrderPaymentService.kt
san-ghun: I put the payment method and currency unit into service layer, because I do not know well about the payment domain(?). 

After your comment, I have searched and considered about how to deal with the order and payment domain. I am placing new property `currency` into `Order` entity, and changing the `initiatePayment()` method into this,

```kotlin
@Service
class OrderPaymentService(
    private val stripeClient: StripeClient,
) {
    fun initiatePayment(order: Order): PaymentResponse {
        val paymentMethod = order.paymentMethod ?: throw PaymentFailedException(MESSAGE_PAYMENT_METHOD_REQUIRED)
        val paymentRequest =
            PaymentRequest(
                amount = order.paymentAmount.toInt(),
                currency = order.currency,
                paymentMethod = paymentMethod,
            )
        return stripeClient.createCheckoutSession(paymentRequest)
            ?: throw PaymentFailedException(MESSAGE_PAYMENT_FAILED_WITH_NULL)
    }

    companion object {
        const val MESSAGE_PAYMENT_METHOD_REQUIRED = "Payment method required"
        const val MESSAGE_PAYMENT_FAILED_WITH_NULL = "Payment failed with null"
    }
}
```

Thank you for your comment!

src/test/kotlin/ecommerce/client/StripeClientTest.kt
san-ghun: > What problems might arise if we call the actual external client every time the test runs?
And in such cases, what kind of testing approach can we use to prevent this?

Well, the external API requests might require some cost?
So, do you think it would be better to apply mocked test?

> Furthermore, it’s important to consider that the key focus should be testing our internal logic,
not the behavior of the external client! 😄

Yes, I need mocked test.

So, is it meaning that testing the behavior of the external client is better be delegated to E2E test?
Where can I put test for external client test?

san-ghun: Instead keeping `StripeClientTest`, I am adding `OrderPaymentServiceTest` to introduce mock test using `Mockito`.

src/main/kotlin/ecommerce/controller/api/OrderController.kt
san-ghun: Here I am applying a logger with `SLF4J`.

```kotlin

import org.slf4j.Logger
import org.slf4j.LoggerFactory

interface Loggable {
    val logger: Logger
        get() = LoggerFactory.getLogger(this::class.java)
}
```

san-ghun:
Hey GuMak!

Thank you for the warm welcome.
I've sent you the secret key values via DM.

## Changes

1. Apply the logger
    - Finally, apply logger with `SLF4J` to replace `println`
2. Add `currency` property into `Order` entity to handle the value in entity level
3. Remove tests that check the results of external APIs
4. Introduce mocked tests to check internal logic
5. Move credential value as an environment variable

## Questions

Here some questions I've got from your comments.

- [ToDo security issues to handle?](https://github.com/herotechcourse/spring-ecommerce-order/pull/22#discussion_r2280513081)
- [Where to delegate the external client test?](https://github.com/herotechcourse/spring-ecommerce-order/pull/22#discussion_r2275375173)

Thank you!

README.md
```diff
@@ -2,36 +2,95 @@
...
+4. Implement business logic for payment processing - `@Transactional`
+    - [x] subtract order quantity from `option` quantity, when the order processed with success
+    - [x] if the ordered product exists as a `CartItem` it should be removed from table and transferred to `Orders`
+    - [x] **MUST** handle the error safely, when the payment approval API call fails
+        - [ ] inform to the user cleary for the reason of failure of payment
+        - [ ] display a proper error message why the payment failed
+            - when the order failure due to payment rejection
+            - e.g., expired session, invalid payment method, insufficient balance, etc.
+5. Implement **Orders** feature
+    - **MUST** be visible in `OrderRequest` -> `OrderResponse`
```
Hyeon9mak: This isn’t something mandatory for this mission, so there’s no need to rush.
San can refine it later whenever you want. It should be enough just to keep a TODO list and make a habit of checking it regularly so you don’t forget.

src/main/resources/application.properties
```diff
@@ -1,8 +1,11 @@
 spring.application.name=spring-ecommerce
+
 spring.h2.console.enabled=true
 spring.h2.console.path=/h2-console
+
 spring.datasource.url=jdbc:h2:mem:database
-jwt.secret-key=X1yk3dV+NQemyyJbg6WzioaRR1NArxZxjMvvroTZauk=
+
+jwt.secret-key=${JWT_SECRET_KEY}
```
Hyeon9mak, line 8: 😉 👍

src/test/kotlin/ecommerce/service/OrderPaymentServiceTest.kt
```diff
@@ -0,0 +1,86 @@
...
+class OrderPaymentServiceTest {
+    @Mock
+    lateinit var stripeClient: StripeClient
+
+    lateinit var orderPaymentService: OrderPaymentService
+
+    @BeforeEach
+    fun setUp() {
+        orderPaymentService = OrderPaymentService(stripeClient)
+    }
```
Hyeon9mak, line 26: > So, is it meaning that testing the behavior of the external client is better be delegated to E2E test?
Where can I put test for external client test?

We can’t always guarantee that the external client will be available, and if calling it incurs costs, that could become a heavy burden.

More importantly, the purpose of your test isn’t to verify the external client’s behavior, but to ensure that our code correctly handles the business requirements.

That’s why it’s better to hide the external client behind mocks, or even better, extract the critical business logic into a separate domain object and write unit tests around it. 😄

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
```diff
@@ -12,26 +12,26 @@ import org.springframework.web.bind.annotation.ControllerAdvice
...
-        println("RuntimeException occurred: $errorBody")
+        logger.error("RuntimeException occurred: $errorBody")
         return ResponseEntity.internalServerError().build()
     }
 
     @ExceptionHandler(NotFoundException::class)
     fun handleNotFoundException(e: NotFoundException): ResponseEntity<Void> {
-        println("NotFoundException occurred: " + e.message)
+        logger.error("NotFoundException occurred: " + e.message)
         return ResponseEntity.notFound().build()
```
Hyeon9mak: What do each of the log levels mean in our context?
For example:

- DEBUG logs could be used during deep development,
- INFO logs might be something we check in daily reports,
- WARN logs could be ignored most of the time but trigger alerts if they repeat for a specific API,
- ERROR logs should always trigger alerts so that developers are immediately aware.

If we don’t define such rules, it might end up requiring developers to manually check every log for every situation.
What do you think about this approach?

src/main/kotlin/ecommerce/model/Order.kt
```diff
@@ -0,0 +1,65 @@
...
+    val createdAt: LocalDateTime = LocalDateTime.now(),
+    @Column(nullable = false)
+    val updatedAt: LocalDateTime = LocalDateTime.now(),
+    @Column(nullable = false)
+    var status: OrderStatus = OrderStatus.PRE_PAYMENT,
+    @Column(nullable = false)
+    var paymentAmount: Double = 0.0,
+    @Column(nullable = false)
+    var currency: String = "USD",
+    @Column
```
Hyeon9mak: > I have searched and considered about how to deal with the order and payment domain. I am placing new property currency into Order entity, and changing the initiatePayment() method into this,

Actually, what I wanted to suggest was either assigning a default value through the constructor or using an enum class. Imagine a teammate receives a requirement like: 'Please change the currency unit from USD to KRW.' Without enums, they would have to search for the literal string "usd" across all classes, which can be error-prone and inefficient.

In this way, the class file name can serve as a signpost that helps teammates locate the relevant business logic. This is what we usually refer to as the **role of the object**. 😉

Hyeon9mak (changes requested):
San, great job reflecting the feedback!

I’ve left comments on a few simple points that should be checked at this stage, as well as answers to the questions you raised.
Since I wasn’t sure if all your concerns were fully addressed, I’m leaving one last request changes. 😉 

Feel free to ask any additional questions anytime!

README.md
san-ghun: Thank you! I am going to make a section for TODO to keep reminding. I will keep that in mind!

src/test/kotlin/ecommerce/service/OrderPaymentServiceTest.kt
san-ghun: > That’s why it’s better to hide the external client behind mocks, or even better, extract the critical business logic into a separate domain object and write unit tests around it. 😄

Thanks for your clear explanation, it makes my mind even clear.
It is still challenge for me to extract the critical business logic into separate domain object in practice, however now I know how to try on!

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
san-ghun: Yes, truly. 
With your review, I am setting WARN logs on client error responses and ERROR logs on server error responses.
Still considering where to put DEBUG logs and INFO logs, but it is now much clear for me how to apply each log levels. Thanks!

san-ghun: I have an additional question.

I've saw some Kotlin codebases, recently. And I've saw some devs using `assert()` and `check()`, including `require()`. I've also used `require()` often when I am validating input parameter and arguments. However never `assert()` and `check()`.

Do you use `assert()` and `check()` on your project?

src/main/kotlin/ecommerce/model/Order.kt
san-ghun: Oh well, I guess I misunderstood your intension. 
Yes, having an enum to represent currency unit is a great idea. 
I've also added validation of currency for 'Order' entity, and added test for it.

Thanks for you suggestion!

I really need to get into the habit of reading code from teammates' perspective.

san-ghun:
Hey GuMak!

Thanks for your considerable review!

## Changes

1. Refine logging strategy by applying different log levels to the category of error response.
	- WARN log to 4xx client error
	- ERROR log to 5xx server error
2. Introduce `OrderCurrency` enum class to represent currency of `Order`
	- change default value of `OrderPlaceForm` DTO's currency value to get the value from the enum.
	- remove default value of currency in `Order` entity, rather add validation to check valid currency is given.

### Question

I am leaving a question to you.

- [`assert()` and `check()`](https://github.com/herotechcourse/spring-ecommerce-order/pull/22#discussion_r2295943641)

Thank you!

Hyeon9mak (approved):
Hello San,
Great job sticking with it through the final mission over such a long period!
It looks like you've put a lot of thought into how to collaborate effectively with your teammates. 😄 

> Do you use assert() and check() in your project?

Sure!
If you check which exceptions each of them throws, you can better predict when it's appropriate to use one over the other.

image
image

All requirements have been met, so I'm approving!
I hope the comments I've left along the way were helpful. 😄 
Wishing you an enjoyable time with the rest of your team project! 👍

PR #23, opened 2025-08-11
[spring-ecommerce-order - Step-2] Submit Vito's mission

VittorioDeMarzi:
# eCommerce — Step 2 (Payments, Orders & Deployment)

Hi there 👋 — thanks a lot for reviewing this step! I’d really appreciate your feedback on the implementation and the direction.

## Summary

This PR introduces Stripe-based payments, basic failure handling, initial `Order`/`Payment` domain modeling, stock updates, cart cleanup, CORS configuration, and a simple EC2 deployment flow with a start script and a short README.

### 1) Stripe Payment Integration

* Uses Stripe **Payment Intents API** to create a payment.
* For now, requests are sent with `confirm=true`, so there’s no separate confirmation step on our side.
* Uses **sandbox secret key** for development/testing.

### 2) Payment Failure Handling

* Exceptions from the Stripe client are currently caught in `PaymentService`.
* On failure, I return a `PaymentResponse` with status `failed` and the error message (temporary approach until we settle on a cleaner error model).

### 3) `Order` and `Payment` Entities (separate tables)

**Why separate?**

* Payments can fail/retry independently of orders.
* Clearer lifecycle boundaries (order creation vs. payment processing).
* Easier to extend with refunds, partial payments, multiple attempts.

**Current state / gap**

* Both entities store the `paymentIntentId`, but there is **no explicit relationship yet**.
* This is risky: for failed payments the `paymentIntentId` can be `null`, so we can’t reliably map a payment attempt back to its order.
  👉 I plan to add a relationship (likely `Order 1..1 Payment` or `Order 1..n PaymentAttempts`) so we always keep referential integrity.

### 4) Stock Update

* When a payment succeeds:

  * The order is marked `succeeded`.
  * The **stock of the selected product option** is decreased by the ordered quantity.

### 5) Cart Cleanup

* After placing an order, the purchased items are removed from the cart.

### 6) CORS Configuration (via `WebMvcConfigurer`)

* Resolves client–server **cross-origin** issues.
* **Allowed origins:** `*` (temporary, for development).
* **Allowed methods:** `GET, POST, PUT, DELETE, OPTIONS` (can add `HEAD` if needed).
* **Allowed headers:** `*`
* **Max age:** `1800` seconds.

### 7) Deployment to AWS EC2

* Built the JAR **locally** (EC2 instance has limited memory for Gradle build).
* Uploaded the JAR to `~/app` and added a simple deployment script there.
* Included a small README with the script.

**Automated Deployment Script (`start.sh`)**

* Stops any running instance of the app.
* Copies the latest JAR into the deployment location.
* Starts the new version in the background with `nohup`.

## Environments / Test Data

* Test and production profiles are separated.
* `data.sql` is **kept for tests only**.
* Currently there’s **no admin user** and **no seeded products** in prod profile.

  * To test quickly: sign up via the app, then promote your user to admin via the **H2 console**.
  * If that’s inconvenient, I can re-enable `data.sql` seeding for review.

## How to review (suggested path)

1. Set Stripe sandbox secret in your env/properties.
2. Run the backend.
3. Create a user, (optionally) promote to admin via H2.
4. Place an order with Stripe test cards (e.g., standard Visa).
5. Verify:
   * `Payment` created with the expected status.
   * `Order` status updated on success.
   * **Stock** decreased.
   * **Cart** items cleared.
6. Try a failing payment method and check the returned error payload and persisted state.

## Feedback requested

* **Exception handling & API error design:**
  Is there a cleaner way you’d prefer? (e.g., controller advice + structured error codes instead of building a failed `PaymentResponse` in the service)
* **Entity relationships:**
  Would you go for `Order 1..1 Payment` or model payments as attempts (`Order 1..n PaymentAttempt`)?
* **Stripe flow:**
  I’m using `confirm=true`. Would you recommend moving to explicit confirm or webhooks next?

## Improvements

* Switch to **PostgreSQL** for persistent data.
* Add an explicit **Order ↔ Payment** relation (and possibly `PaymentAttempt`).
* Introduce a **sealed `PaymentResponse`** hierarchy for `Success`/`Failure`.
* Improve Stripe flow (idempotency keys, consider **webhooks** for source-of-truth updates).
* Add endpoints to **list/fetch Orders and Payments** for the user/admin.
* E2E test for the whole process of order/payment

---

Thanks again for taking the time to review this!
If anything is hard to test (e.g., admin creation), tell me and I’ll re-enable `data.sql` seeding to speed things up. 🙏


Start.sh
```diff
@@ -0,0 +1,28 @@
...
+
+OLD_PID=$(pgrep -f "$APP_JAR" || true)
+if [ -n "$OLD_PID" ]; then
+  echo "Stopping old process: $OLD_PID"
+  kill -TERM "$OLD_PID" || true
+  sleep 3
+fi
+
+cp -f "$JAR" "$APP_JAR"
+nohup java -jar "$APP_JAR" > "$LOG_FILE" 2>&1 &
```
pkeugine, line 24: Nice job using the script to deploy the application 👍 
What do you think `nohup`, `2>&1`, and `&` mean here?
I ask this because these knowledge are important when it comes to deploying the application,
and they come out often in tech interviews :)

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,86 @@
...
+            "Creating PaymentIntent amount={} {} method={}",
+            amount,
+            req.currency,
+            req.paymentMethod,
+        )
+
+        return try {
+            val response =
+                restClient.post()
+                    .uri("https://api.stripe.com/v1/payment_intents")
```
pkeugine: When working with 3rd party API, I recommend taking care of their URLs in a separate properties file.
(even `application.properties` file is fine).

It would be great for you to find out why, and think about refactoring it 🙂

src/test/kotlin/ecommerce/stripe/StripeClientTest.kt
```diff
@@ -0,0 +1,68 @@
...
+        "spring.jpa.hibernate.ddl-auto=none",
+    ],
+)
+@ActiveProfiles("test")
+class StripeClientTest {
+    @Autowired
+    private lateinit var stripeClient: StripeClient
+
+    @Test
+    fun test1() {
```
pkeugine: It would be great to name this test with a meaningful one.

src/test/kotlin/ecommerce/stripe/StripeClientTest.kt
```diff
@@ -0,0 +1,68 @@
...
+            stripeClient.createCheckoutSession(
+                OrderRequest(
+                    Currency.EUR,
+                    "pm_card_visa",
+                ),
+                amount = 1000,
+            )
+        assertThat(actual).isNotNull
+        assertThat(actual?.amount).isEqualTo(10.0)
+    }
```
pkeugine: I'm unable to make some tests pass when I clone the repository and run them immediately.
I've read your PR message, but I think this is a great opportunity for us to make a PR more reviewer-friendly one.

* First, rather than using AI to generate a long, detailed message, it would be great for us to keep only the important information in there. When working with teammates, the person who reviews our PR are teammates who have their own agenda, and uses their spare time to read and check the PR. (We are also one of them too) This means rather than making a PR "full of information", it's great to make it "only have what we really need".

* This means we don't want reviewers to find and debug their way through the tests, but rather let them go through it as quick as possible. (Best : make tests pass without extra work or even disable / mock / stub the 3rd party related tests)

* When the ingredient to successful tests is related to security, send them to the reviewer in a secure, team-approved way. (For now, I think it's great to send it via DM)

Since this is our last step of this level, I thought this is a great time for us to go over these information. 🙂

README.md
```diff
@@ -103,5 +103,76 @@ Content-Type: application/json
    - No need to create a separate HTTP API. 
    - This logic should be implemented in the Service class or Entity class for future reuse.
 
+# Step2 - HTTP Clients
```
pkeugine, line 106: Addition to the previous comment, it'll be great to go over the commit count of a PR.
Though it varies from team to team, it's great to keep the amount of commits not too much, since it makes it hard for the reviewer to follow the process. If the team's merge strategy is not squash merge, those commit logs will all go into the main branch, which results in extra-long commit history.

When you refactor often while working on a feature, try out git's rebase & fixup features. Using git's various features to manipulate commit logs definitely helps us control the log in a more clean and meaningful way. Since you seem like a real developer now, I thought this would be a great time for me to mention this!

Though I have to say, some teams don't really care too much about it, and some developers have a different opinion from mine, so this is just a suggestion. 🙂

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val stripeClient: StripeClient,
+    private val paymentJpaRepository: PaymentJpaRepository,
+) {
+    fun createPaymentIntent(
+        request: OrderRequest,
+        amount: BigDecimal,
+    ): PaymentResponse {
+        try {
+            val paymentResponse =
+                stripeClient.createCheckoutSession(request, amount.multiply(BigDecimal.valueOf(100)).toInt())
```
pkeugine: Nice use of `BigDecimal` when working with money! 🤑 
However, if we `toInt()` it, there would still be an edge case that we should consider.
How do you think about this?

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import ecommerce.client.StripeClient
+import ecommerce.dto.OrderRequest
+import ecommerce.dto.PaymentResponse
+import ecommerce.enum.PaymentStatus
+import ecommerce.exception.StripePaymentException
+import ecommerce.model.Payment
+import ecommerce.repository.PaymentJpaRepository
+import ecommerce.repository.getByPaymentIntentIdOrThrow
+import jakarta.transaction.Transactional
```
pkeugine: I think the annotation that you were trying to import was `org.springframework.transaction.annotation.Transactional`, not `jakarta.transaction.Transactional`!

[We went over this before](https://github.com/herotechcourse/spring-ecommerce-order/pull/7#discussion_r2249306942), so I hope we acknowledge their difference and import Spring's annotation. 🙂

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,69 @@
...
+                Payment(
+                    amount = amount,
+                    currency = request.currency,
+                    paymentMethod = request.paymentMethod,
+                    paymentIntentId = "Payment Intent Failed",
+                    status = PaymentStatus.FAILED,
+                )
+            paymentJpaRepository.save(payment)
+            return paymentResponse
+        }
```
pkeugine, line 64: I see you've `return`ed a response in a catch statement rather than throwing a new exception 😮 
This is to prevent rollback from happening right? So we can persist failed payment?

README.md
```diff
@@ -103,5 +103,76 @@ Content-Type: application/json
...
+## Step 2-1 - External API: Place Order with Stripe
```
pkeugine, line 107: I see you've changed the database vendor to PostgreSQL.
What were the selling points of that vendor?

pkeugine (changes requested):
Hello Vito, and nice work with the additional features. 👍 

Since we've already started team project, I added some comments that you can easily fix and respond to, while keeping the important topics. Also, you've done well so there weren't so much to cover. It won't be too hard to check the reviews as you work on your next project.

Once you go over the reviews, I think we'll be ready to merge this PR!
Let's keep it up till the very end :)

Happy Hacking! 🔥 🔥 🔥

Start.sh
VittorioDeMarzi: Thank you 🙂

- nohup → Prevents the process from being terminated when the terminal session is closed (it ignores the SIGHUP signal). Useful when running processes in the background on a server.
- 2>&1 → Redirects stderr (2) to stdout (1), so both standard output and error messages go into the same log file.
- & → Runs the process in the background, so the shell is free for other commands while the application keeps running.

This way, the application keeps running even after the SSH session ends, and I can log both normal output and errors in a single file for easier debugging.

src/main/kotlin/ecommerce/client/StripeClient.kt
VittorioDeMarzi: Thanks for the suggestion!
I’ve moved Stripe’s endpoints into application.yml using @ConfigurationProperties (stripe.base-url and stripe.payment-intents-path) and configured the RestClient with baseUrl.

This way:
- The code is more maintainable
- I can easily switch URLs for different environments (dev, test, prod)
- It removes hardcoded strings from the codebase

src/test/kotlin/ecommerce/stripe/StripeClientTest.kt
VittorioDeMarzi: You’re absolutely right. I’ll make sure my future PR messages are more concise and focused on what the reviewer really needs to know.

README.md
VittorioDeMarzi: I actually thought the goal was to commit every single modification with a clear and meaningful message, so it creates a sort of “story” of my steps and thinking process.
Thanks for pointing this out, it’s good to know that keeping commits fewer and more focused can be better for reviewers.
I’ve just read about rebase and fixup, and I’ll start using them in the future to keep my commit history cleaner and more meaningful.

src/main/kotlin/ecommerce/service/PaymentService.kt
VittorioDeMarzi: I hadn’t thought about that, using toInt here can silently truncate decimals and even overflow for large amounts.
A safer approach would be to round explicitly and use Long instead.

src/main/kotlin/ecommerce/service/PaymentService.kt
VittorioDeMarzi: Yes, you’d already pointed that out, and I’ve read the difference carefully.
I need to start importing more carefully and double-check the imports, since the IDE often does it automatically.

src/main/kotlin/ecommerce/service/PaymentService.kt
VittorioDeMarzi: Exactly. This is the approach I took; otherwise, throwing an exception would roll everything back and I wouldn’t be able to persist failed payments in the DB.
 Is this approach bad, or acceptable?
 
These days transactional is a big topic here around. I could even try to avoid the rollback for a kind of exception, or save the failed payment in another transactional function with a new propagation.
What do you think? This would be a gold hint for the final project too :)

README.md
VittorioDeMarzi: Actually, I didn’t change the database vendor, it’s still the same in this project.
I only mentioned PostgreSQL in the README as one of the possible future improvements, since it has excellent support for transactions and data integrity.

src/main/kotlin/ecommerce/service/PaymentService.kt
pkeugine: I think it's fine for now, and you've done a good job.
Like you said, you can also avoid the rollback for certain exceptions, or deal with it using other components in the application, but I think this method fits our current needs 👍

README.md
pkeugine: Oh I misunderstood the PR message 😅 
It would be great to compare various vendors and choose the best one for your team project 👍

pkeugine (approved):
Hello Vito, and well done finishing this mission 👍 
You've come a long way, and now you're ready to dive into a team project!

I hope you do well on it just like you did in these missions,
and I wish you all the luck.

As always, in team project, and beyond
Happy Hacking! 🚀 🚀 🚀

PR #24, opened 2025-08-14
[Spring-Ecommerce-Order Step 2] Submit arekiu's Mission

arekiu:
Good day Pk! Nice to meet you again in this mission

## Overview
This pull request introduces the "Place Order" and "Orders" features, completing Step 2 of the spring-ecommerce-order mission. The key additions include:

- Stripe Integration: The "Place Order" feature is now integrated with the Stripe Payment Create API to handle real-time payment processing.

- Order and Stock Management: When an order is successfully placed, the stock of the selected product option is decreased, and the item is removed from the user's cart if present.

- Order Details: The "Orders" feature displays essential information, including order date and time, status, purchased items, etc.

- Product Options: Products now support options, with validation to make sure that each product has at least one option and that option names are unique within a product.

All changes are built upon the existing codebase and adhere to the mission's requirements.

## Notes for the Reviewer

- StripeClient Error Parsing: The parseStripeError method in StripeClient uses jacksonObjectMapper to parse the error response from Stripe. Is this implementation the best approach to manage error formats from the Stripe API?

- Payment Failure Logic in OrderService: When a payment fails, the OrderService throws an IllegalStateException with a message from the payment response. Is this the most appropriate way to handle payment failures?

I'm also currently working on the group project, so I apologize if I miss important points to discuss. I'm looking forward to hear your feedback on these and other aspects of the implementation. Also, I will send the .env file on Slack so you can test the project.

Thank you for your time and feedback!


src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,91 @@
...
+        return try {
+            val root = jacksonObjectMapper().readTree(json)
+            println(root.toString())
+            val errNode = root.path("error")
+            PaymentErrorResponse(
+                code = errNode.path("code").asText(null),
+                // converts value to a string or provides null if absent/missing.
+                declineCode = errNode.path("decline_code").asText(null),
+                message = errNode.path("message").asText(null),
+            )
```
pkeugine: I think we can remove `println` part since it's for debugging this method while it's in development.
Also, since the code describes itself, the comment can be taken out :)

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,91 @@
...
+        } catch (ex: Exception) {
+            null
+        }
+    }
```
pkeugine: **Q. Is this implementation the best approach to manage error formats from the Stripe API?**

Well, I'm not sure if there is a `best` approach on handling 3rd party api call, since the situation and api characteristics would be different depending on what we are trying to do, and how they are made.

However, what we can do is, just like how we handle domain classes, we can make a class with fields that match the response of Stripe's api, and make a small parser that handles json to class mapping. This way, we can let `PaymentErrorResponse` be in a format that we want. Let's say we added a new payment method using some other company's api. We can then add that company related class and parser as well. Some companies might have a different `code`, `declineCode`, or `message` (they might not even have one 😂 ), so our `PaymentErrorResponse` will have fields and codes that are related to our service rather than Stripe's.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,91 @@
...
+            )
+        } catch (e: Exception) {
+            throw IllegalArgumentException("Stripe error: ${e.message}", e)
+        }
+    }
+
+    private fun parseStripeError(json: String?): PaymentErrorResponse? {
+        if (json.isNullOrBlank()) return null
+        return try {
+            val root = jacksonObjectMapper().readTree(json)
```
pkeugine: We can also inject `ObjectMapper` dependency through constructor and use it like a singleton.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,91 @@
...
+                    "lost_card" -> "Payment failed: lost card."
+                    "stolen_card" -> "Payment failed: stolen card."
+                    else -> "Payment was declined by your bank."
+                }
+            "expired_card" -> "Payment failed: your card has expired."
+            "processing_error" -> "Payment failed due to a processing error."
+            else -> error?.message ?: "Payment failed for an unknown reason."
+        }
+    }
+}
```
pkeugine: I think we do need a way to handle error messages and convert it to our own, so I do agree that this process is necessary.

However, just like our previous review, `PaymentErrorResponse` seems like it should be open to situations where it handles more than Stripe's api response. So rather than using `PaymentErrorResponse` to convert error message, I think this should be done in Stripe related logic before we use the general `PaymentErrorResponse`.

Or to just make it more simple, change the name to `StripePaymentErrorResponse` :)
Since `PaymentErrorResponse` is only used in `StripeClient`, I think this way of refactoring seems okay. But to make it better, it'll be great to have Stripe related package and manage all the related classes there.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,91 @@
...
+            throw IllegalArgumentException("Stripe error: ${e.message}", e)
+        }
+    }
+
+    private fun parseStripeError(json: String?): PaymentErrorResponse? {
+        if (json.isNullOrBlank()) return null
+        return try {
+            val root = jacksonObjectMapper().readTree(json)
+            println(root.toString())
+            val errNode = root.path("error")
```
pkeugine: You've done a great job managing this project with clear names in variables.
In this class, you've also used a variable called `errorJson`.
This seems like a minor point, but I think the name `errNode` feels odd and feels like it's not written by you.

Why don't we keep the convention here too?
I think this kind of consistency is also important when working as a team.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,91 @@
...
+            PaymentErrorResponse(
+                code = errNode.path("code").asText(null),
+                // converts value to a string or provides null if absent/missing.
+                declineCode = errNode.path("decline_code").asText(null),
+                message = errNode.path("message").asText(null),
+            )
+        } catch (ex: Exception) {
+            null
+        }
```
pkeugine: It would be great to add in a log here since we're treating this exceptional situation as a `success` by returning `null`, but we want to know what really happened. If there's something we have to handle, we must know what went wrong. It's highly likely that the 3rd party api spec changed. I think this is especially important in payment related logic. How do you think about this?

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,128 @@
...
+    ) {
+        if (paymentResponse.status == "succeeded") {
+            order.status = OrderStatus.COMPLETED
+            cart.clearCartProducts()
+            cartRepository.save(cart)
+        } else {
+            order.status = OrderStatus.CANCELLED
+            throw IllegalStateException("Payment failed: ${paymentResponse.errorMessage}")
+        }
+    }
```
pkeugine: **Q. Is this the most appropriate way to handle payment failures?**

The quick and easy way to check is to see if cancelled payment got persisted in the database as you expected it to. Did it get persisted?

pkeugine (changes requested):
Hello Alex, and nice work with the final step of this mission :)

Since we're already working on the team project, I added some reviews that you can check quickly. It still has some important topics in it, so please check them out when you have the time.

I'll merge this PR when you request for a review again! Until then,
Happy Hacking! 😄 😄 😄

src/main/kotlin/ecommerce/stripe/StripeClient.kt
arekiu: Right! I forgot to remove the print, and the comment I did for myself because I didn´t know what exactly it was doing with that "null"

src/main/kotlin/ecommerce/stripe/StripeClient.kt
arekiu: Thanks, that's a good point. I've refactored it to inject the ObjectMapper bean instead.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
arekiu: Thanks for the detailed review and the explanation. I think that in fact, is a better approach.

I've implemented the suggested changes for the error handling. The logic is now decoupled by mapping the Stripe error to a dedicated DTO before converting it to our internal response type.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
arekiu: Thanks again for all the suggestions!
I've refactored it so that the StripeClient is now fully responsible for translating the specific Stripe error codes into the final more readable messages. The convertMessageFromError function now works directly with the StripeErrorDetail DTO, and I've removed the intermediate PaymentErrorResponse class completely. Also I created a dedicated package for Stripe related DTOs

src/main/kotlin/ecommerce/stripe/StripeClient.kt
arekiu: You are right! I didn´t know how to deal to parse the error, and in that moment I found that example online. So possibly I copied the names. But now that part of the code looks much better!

src/main/kotlin/ecommerce/stripe/StripeClient.kt
arekiu: Great suggestion! I've implemented the changes. The StripeClient now throws a custom exception on a parsing failure, and the global handler takes care of it.

src/main/kotlin/ecommerce/service/OrderService.kt
arekiu: Ohh interesting point!
The transaction was not finishing because I throw the exception. I've adjusted the logic to make sure that the CANCELLED status is always committed to the database. The service method now completes successfully and returns the final order, and the controller returns HTTP status based in the result.

arekiu:
Hi Pk! I hope you´re doing well.

Here I refactored some files related to order and Stripe. Sorry for the delay, I´m currently working a lot on the team project!

pkeugine (approved):
Hello Alex, and well done refactoring the project.
It must have been challenging taking care of both the team project and this mission, but you've done well 👍 

I think we are good to go, and this is a great time for us to wrap up this mission.
I hope you gained knowledge and skill throughout these missions,
and I'm sure you're doing great in your team project.

Wish you all the luck, and as always,
Happy Hacking! 🔥 🔥 🔥

PR #25, opened 2025-08-15
[Spring-Ecommerce-Order Step 2] Submit Farhana's Mission

farhanaahmed:
Sorry for the delay in submitting this PR 🙏.

In this step, I have implemented the following:

### Step 2-1: Place Order Feature
- Integrated **Stripe Payment Create API** using the sandbox key.
- Implemented order placement with selected product option and quantity.
- Reduced stock for the selected option after a successful order.
- Removed ordered product from the user's cart if present.
- Added safe handling for payment approval API failures.
- Displayed clear error messages for payment failures (e.g., expired session, invalid method, insufficient balance).

### Step 2-2: Orders Feature
- Implemented **Orders API** to retrieve detailed order information, including:
  - Order date/time
  - Order status
  - Purchased items
  - Stripe checkout session ID
  - Payment amount
- Stored optional payment-related fields in the database.

### Step 2-3: Deployment
- Deployed the order service to AWS.
- Created a deployment script to automate build, upload, and restart.
- Configured CORS to allow smooth client-server interaction.

---

**Additional Notes:**
- Incorporated feedback from Step 1 to maintain best practices.
- Added the **Stripe Secret Key** to the `ec2-farhana` instance inside `env/app.env` (required for test cases to pass).

Thank you very much for your valuable time and feedback! 🙌


build.gradle.kts
```diff
@@ -33,11 +33,10 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
     implementation("nz.net.ultraq.thymeleaf:thymeleaf-layout-dialect:3.2.0")
+    implementation("org.springframework.boot:spring-boot-starter-actuator")
```
Wordbe, line 36: For what kind of monitoring did you add this? 🙂

scripts/deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+RELEASE_DIR="$HOME/releases"
+APP_DIR="$HOME/app"
+APP_JAR_LINK="$APP_DIR/app.jar"
+
+LATEST_JAR=$(ls -t "$RELEASE_DIR"/*.jar | head -n1)
+mkdir -p "$APP_DIR"
+ln -sfn "$LATEST_JAR" "$APP_JAR_LINK"
+
+if systemctl list-unit-files | grep -q '^ecommerce.service'; then
+  sudo systemctl restart ecommerce
```
Wordbe, line 31: If there’s a way to register and run the application using systemctl, it would be good to document it separately in the deployment guide. 📄

src/main/kotlin/ecommerce/config/CorsConfig.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import org.springframework.context.annotation.Configuration
+import org.springframework.web.servlet.config.annotation.CorsRegistry
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+
+@Configuration
+class CorsConfig : WebMvcConfigurer {
+    override fun addCorsMappings(registry: CorsRegistry) {
+        registry.addMapping("/**")
+            .allowedOrigins("*") // <-- important for test expecting "*"
```
Wordbe: You could enhance security by controlling only specific origins. 🙂

src/main/kotlin/ecommerce/config/StripeProperties.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.config
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties("stripe")
+data class StripeProperties(
```
Wordbe: Is a data class really necessary? 🤔

src/main/kotlin/ecommerce/controller/OrderController.kt
```diff
@@ -0,0 +1,40 @@
...
+
+@RestController
+@RequestMapping("/orders")
+class OrderController(
+    private val orderService: OrderService,
+    private val memberRepository: MemberRepositoryJpa,
+) {
+    @PostMapping
+    fun placeOrder(
+        @LoginMember member: MemberResponse,
```
Wordbe: It seems strange that the response object is being used as the request object! 🧐

src/main/kotlin/ecommerce/controller/OrderController.kt
```diff
@@ -0,0 +1,40 @@
...
+        @Valid @RequestBody req: PlaceOrderRequest,
+    ): ResponseEntity<PlaceOrderResponse> {
+        val member =
+            memberRepository.findById(member.id)
+
+        if (member.isEmpty) {
+            return ResponseEntity.badRequest().build()
```
Wordbe: Delegate the role to memberService and unify the structure to return exceptions. Also, it would be good if a common exception-handling object handled the exception responses. 🙂

src/main/kotlin/ecommerce/dto/CartRequest.kt
```diff
@@ -1,3 +1,3 @@
 package ecommerce.dto
 
-data class CartRequest(val productId: Long)
+data class CartRequest(val productOptionId: Long)
```
Wordbe: Is data class needed?

src/main/kotlin/ecommerce/dto/PaymentRequest.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+data class PaymentRequest(
```
Wordbe: Same here.

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
```diff
@@ -0,0 +1,15 @@
+package ecommerce.dto
+
+import jakarta.validation.constraints.Min
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.NotNull
+
+data class PlaceOrderRequest(
```
Wordbe, line 7: Same here.

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
```diff
@@ -0,0 +1,15 @@
...
+    @field:NotNull
+    val optionId: Long,
+    @field:Min(1)
+    val quantity: Long,
+    @field:NotBlank
+    val paymentMethod: String,
```
Wordbe: Since this field only has predefined values, an enum seems fitting. 🙂

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
```diff
@@ -0,0 +1,15 @@
...
+    val currency: String = "usd",
```
Wordbe, line 14: Here as well.

src/main/kotlin/ecommerce/dto/PlaceOrderResponse.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+data class PlaceOrderResponse(
```
Wordbe: .

src/main/kotlin/ecommerce/dto/PlaceOrderResponse.kt
```diff
@@ -0,0 +1,7 @@
...
+    val orderId: Long?,
+    val paymentStatus: String,
```
Wordbe, line 5: .

src/main/kotlin/ecommerce/dto/StripeIntentResponse.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+data class StripeIntentResponse(
```
Wordbe: .

src/main/kotlin/ecommerce/dto/StripeIntentResponse.kt
```diff
@@ -0,0 +1,7 @@
...
+    val id: String,
+    val status: String,
+    val client_secret: String? = null,
```
Wordbe: Please follow Kotlin field naming conventions. 🙂
If you need mapping, please refer to `@JsonProperty`

src/main/kotlin/ecommerce/entity/Payment.kt
```diff
@@ -0,0 +1,32 @@
...
+    @Column(nullable = false)
+    val amount: Long,
+    @Column(nullable = false)
+    val currency: String,
+    @Column(nullable = false)
+    val status: String,
+    @Column(nullable = true)
+    val paymentMethod: String? = null,
+    @Column(nullable = true)
+    val failureReason: String? = null,
```
Wordbe, line 35: Most payments will likely succeed, so failReason will mostly be null. It might be good to separate it into another table, or managing it through log messages could be sufficient. 💡

src/main/kotlin/ecommerce/entity/OrderItem.kt
```diff
@@ -0,0 +1,27 @@
...
+@Table(name = "order_items")
+class OrderItem(
+    @Id @GeneratedValue(strategy = GenerationType.IDENTITY)
+    val id: Long? = null,
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "order_id")
+    val order: Order,
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "product_option_id")
+    val productOption: Option,
```
Wordbe, line 22: orderItem should contain snapshot data at the time of the order. For example, if product name information changes, the order information referencing it could also change. Think about the structure carefully. 📦

src/main/kotlin/ecommerce/entity/Product.kt
```diff
@@ -33,16 +35,25 @@ class Product(
...
-        return id?.hashCode() ?: 0
+        if (this.id != null && other.id != null) {
+            return this.id == other.id
+        }
+
+        return false
     }
+
+    override fun hashCode(): Int = id?.hashCode() ?: 31
 }
```
Wordbe, line 59: Why did you create this hashCode like this?

src/main/kotlin/ecommerce/infrastructure/StripeClient.kt
```diff
@@ -0,0 +1,46 @@
...
+            restClient.post()
+                .uri("https://api.stripe.com/v1/payment_intents")
+                .header(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secret}")
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                .body(body)
+                .retrieve()
+                .toEntity(StripeIntentResponse::class.java)
+                .body ?: throw IllegalArgumentException("Stripe error: empty body")
+        } catch (e: RestClientResponseException) {
+            val reason = e.responseBodyAsString.take(500)
```
Wordbe, line 43: I think it’s better to log all error messages. They can serve as hints for solving problems. 🙂

src/main/kotlin/ecommerce/repository/OptionRepositoryJpa.kt
```diff
@@ -1,6 +1,11 @@
 package ecommerce.repository
 
 import ecommerce.entity.Option
+import jakarta.persistence.LockModeType
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Lock
 
-interface OptionRepositoryJpa : JpaRepository<Option, Long>
+interface OptionRepositoryJpa : JpaRepository<Option, Long> {
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
```
Wordbe, line 9: What’s the reason for introducing pessimistic locking? It would also be good to understand the trade-offs. ⚖️

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -6,43 +6,53 @@ import ecommerce.entity.Cart
...
-                product = product,
-                cart = cart,
-                quantity = 1,
-            )
+        val existing =
+            cartItemRepositoryJpa
+                .findByCartIdAndProductOptionId(cartId!!, option.id!!)
 
-        cartItemRepositoryJpa.save(carItemEntity)
+        if (existing != null) {
```
Wordbe, line 36: Try using a clear method instead of a setter. 😀

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,123 @@
...
+        if (stripeRes.status != "succeeded") {
+            throw IllegalArgumentException("Payment not approved (status=${stripeRes.status}).")
+        }
+
+        val orderId = persistAfterStripeSuccess(member, option.id!!, req.quantity, amountMinor, req.currency, stripeRes)
+        return PlaceOrderResponse(orderId, stripeRes.status, "Order placed and paid successfully.")
+    }
+
+    @Transactional
+    protected fun persistAfterStripeSuccess(
```
Wordbe: Check if the Transactional on this method really works. If not, why could that be? Hint: proxy. 🔍

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,123 @@
...
+                order = order,
+                amount = amountMinor,
+                currency = currency,
+                status = "PAID",
+                stripeSessionId = stripeRes.id,
+                paymentMethod = "stripe",
+            ),
+        )
+
+        option.quantity -= requestedQty
```
Wordbe: Do not use setter. And think about why this is recommend.

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,123 @@
...
+        cartRepository.findByMemberId(member.id!!)?.let { cart ->
+            cartItemRepository.findByCartIdAndProductOptionId(cart.id!!, option.id!!)?.let {
+                cartItemRepository.delete(it)
+            }
+        }
+
+        return order.id!!
+    }
+
+    private fun toMinorUnits(
```
Wordbe: Is this really the responsibility of OrderService? If not, who would be better suited to handle it? 🙂

src/test/kotlin/ecommerce/controller/OrderControllerTest.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+import java.util.Optional
+import org.mockito.kotlin.any as ktAny
+
+@WebMvcTest(OrderController::class)
+@AutoConfigureMockMvc
+class OrderControllerTest
+    @Autowired
+    constructor(
```
Wordbe, line 40: I’m curious why this was declared in the constructor!

Wordbe (changes requested):
Hello Farhana!

Thank you for writing and submitting your code so carefully.
You did a great job writing the deployment script and the test code as well.
Please check the comments and apply the parts you can. Repeated points have been reviewed with `.`

Thank you! 😊

build.gradle.kts
farhanaahmed: I added the `actuator dependency` so that I can use the /actuator/health endpoint in my deploy.sh script and also check it manually on the EC2 instance. This helps me quickly verify that the application is running healthily after deployment. For now, I’m only using the health endpoint, so it’s a lightweight monitoring setup.

scripts/deploy.sh
farhanaahmed: Thanks for the suggestion. 😄 
I added a short systemd setup guide in the `README.md` under “Run as a systemd service”.
The deploy script remains the same. It automatically uses systemctl restart ecommerce when the service file is present and falls back to nohup otherwise.
Let me know if anything else is needed!

src/main/kotlin/ecommerce/config/CorsConfig.kt
farhanaahmed: Thanks for the suggestion 🙂. You were right that allowing all regions was too permissive.
I updated the `CorsConfig` to only allow `http://localhost:3000 (our development frontend)` now and adjusted the tests accordingly.

src/main/kotlin/ecommerce/config/StripeProperties.kt
farhanaahmed: Thanks for pointing this out 🙂. I updated `StripeProperties` to a regular class so it follows Spring Boot’s usual convention for `@ConfigurationProperties`. This also makes it easier to extend later if needed.

src/main/kotlin/ecommerce/controller/OrderController.kt
farhanaahmed: Thanks for the nudge. 😄  Using a response object as input was indeed awkward.
Initially, I didn’t want to create another DTO class because I wanted to keep things simple and avoid introducing unnecessary types. That’s why I reused the existing `MemberResponse`.
But I agree with your point that it was confusing to mix response DTOs as input. To address that, I introduced a `LoggedInMember` DTO dedicated to the `@LoginMember resolver`. In the controller, I still fetch the persisted `Member` from the repository before calling the service.
This way, the controller is clearer, avoids nullable-ID issues, and remains decoupled from JPA entities.

src/main/kotlin/ecommerce/controller/OrderController.kt
farhanaahmed: Thanks for your suggestion. I have updated `OrderController` to use `MemberService` instead of calling `MemberRepositoryJpa` directly. The controller now delegates member lookup and exception handling to the service, so errors are unified through our global exception handler.

src/main/kotlin/ecommerce/dto/CartRequest.kt
farhanaahmed: Thank you for pointing that out! I’ve updated `CartRequest` to be a regular class instead of a data class, since we don’t need features like copy or equals for this request DTO.

src/main/kotlin/ecommerce/dto/PaymentRequest.kt
farhanaahmed: Thanks for catching this. I’ve updated `PaymentRequest` to a regular class instead of a data class, since we don’t need the extra functionality (copy, equals, etc.) here either.

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
farhanaahmed: Thanks for pointing it out 🙂. For `PlaceOrderRequest`, I decided to keep it as a data class because in the tests and in request handling, structural equality is very useful. When Spring/Jackson deserializes JSON into a new instance, the auto-generated equals from the data class ensures it matches correctly in Mockito stubs and assertions. If I switch it to a regular class, the tests start failing since equality falls back to reference-based comparison.

So, while for other DTOs like `CartRequest` or `PaymentRequest` a regular class works fine, I think keeping `PlaceOrderRequest` as a data class is justified here to make testing and validation more consistent.

src/main/kotlin/ecommerce/dto/PlaceOrderResponse.kt
farhanaahmed: Thanks for catching this. I’ve updated `PlaceOrderResponse` to a regular class instead of a data class, since we don’t need the extra functionality (copy, equals, etc.) here.

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
farhanaahmed: Thanks for the suggestion! I’ve updated `Payment` to use a `PaymentMethod` enum and adjusted the DTOs, service, persistence, and tests accordingly. This makes the code more type-safe and prevents invalid values from slipping through.

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
farhanaahmed: Thanks for the suggestion! I’ve updated `PlaceOrderRequest` to also use the `PaymentMethod` enum here.

src/main/kotlin/ecommerce/dto/PlaceOrderResponse.kt
farhanaahmed: Thanks for the suggestion. I’ve updated `PlaceOrderResponse` to a regular class instead of a data class since we don’t need the extra functionality (copy, equals, etc.) here.

src/main/kotlin/ecommerce/dto/StripeIntentResponse.kt
farhanaahmed: Thanks for the review! I’ve changed `StripeIntentResponse` to a regular class as suggested.

src/main/kotlin/ecommerce/dto/StripeIntentResponse.kt
farhanaahmed: Thanks for pointing this out! I’ve renamed `client_secret → clientSecret` to follow Kotlin conventions and added `@JsonProperty` for correct JSON mapping.

src/main/kotlin/ecommerce/entity/Payment.kt
farhanaahmed: Thanks for the suggestion! For now, I’ve kept `failureReason` as a nullable column since it keeps the implementation simple. I agree that separating it into another table or handling it through logs could be a good improvement if payment failure tracking becomes more important later. 🙂

src/main/kotlin/ecommerce/entity/OrderItem.kt
farhanaahmed: Thanks for calling this out! I’ve updated `OrderItem` to store immutable snapshot fields (productNameSnapshot, optionNameSnapshot, productImageUrlSnapshot) and treat price as the unit price at purchase time. `OrderService` now populates these values when creating order items. This keeps order history stable even if catalog names or images change later. Tests were also added to verify the snapshots.

src/main/kotlin/ecommerce/entity/Product.kt
farhanaahmed: I initially used `id?.hashCode() ?: 31` because 31 is a common default in Java hashCode implementations (e.g., String.hashCode) and gave a consistent non-null fallback before persistence. However, I agree that using a fixed number like 31 can be confusing, so I’ve updated it to fall back to `super.hashCode()` for transient entities for better clarity.

src/main/kotlin/ecommerce/infrastructure/StripeClient.kt
farhanaahmed: Thanks for the suggestion! I’ve added logging in `StripeClient` to capture both Stripe API error responses and unexpected failures. This should help with troubleshooting issues more effectively.

src/main/kotlin/ecommerce/repository/OptionRepositoryJpa.kt
farhanaahmed: Thanks for the question. I added pessimistic locking on `Option` to prevent overselling when multiple orders hit the same product option at the same time. During checkout, we: (1) load the option, (2) verify stock, and (3) decrement. With `@Lock(PESSIMISTIC_WRITE)`, the row is locked for the duration of the transaction, so two checkouts can’t pass the stock check concurrently.

Trade-offs:

Pros: Strong guarantee against race conditions and negative stock; simpler logic (no retry loop).
Cons: Lower throughput under contention (blocking), potential deadlocks if lock order is inconsistent, and longer transactions increase wait times.

But for a demo project, I think it’s the simplest, safest way to keep stock accurate, and no extra retry logic or versioning is needed. Let me know your thoughts on this.

src/main/kotlin/ecommerce/service/CartService.kt
farhanaahmed: Thanks for the suggestion! I’ve added an `increaseQuantity()` method inside `CartItem` and updated the service to use it instead of directly modifying the field. This makes the intent clearer and keeps the entity logic encapsulated.

src/main/kotlin/ecommerce/service/OrderService.kt
farhanaahmed: Thanks for the suggestion! 🙏 I refactored the code to avoid mutating the field directly. Instead, I used the `decreaseQuantity(amount: Long)` method inside the `Option` entity, so the entity itself controls how its stock changes. This way, we can encapsulate validation and keep the domain logic in one place, rather than scattering it across services.

src/main/kotlin/ecommerce/service/OrderService.kt
farhanaahmed: Thanks for the feedback. You’re absolutely right. Converting a price to minor currency units is a formatting concern rather than an order-orchestration one.
I’ve extracted that logic into a small `MoneyUtil` utility object, which keeps `OrderService` focused on its primary responsibility (workflow and business rules) and makes the conversion reusable for any future price handling code.

src/test/kotlin/ecommerce/controller/OrderControllerTest.kt
farhanaahmed: Thanks for asking! 🙏 I used constructor injection for the real Spring beans like `MockMvc` and `ObjectMapper` because it’s idiomatic in Kotlin. It is more concise, immutable, and guarantees the beans are injected before the tests run. For the `@MockitoBean` fields, they’re created dynamically by Spring after the context is already built, so they can’t be injected via a constructor; declaring them as lateinit var fields allows Spring to replace the real beans with mocks at runtime. This way, the setup stays clean and consistent while ensuring both real and mock beans work as intended.

src/main/kotlin/ecommerce/service/OrderService.kt
farhanaahmed: Thanks for the hint about proxying! 🙏  You’re right that the `@Transactional` on a method called via self-invocation won’t be applied (and it’s also ignored on private/protected methods).
I first tried making `persistAfterStripeSuccess` method public, but that didn’t work either since the call was still happening inside the same bean, so the proxy never got a chance to intercept it.
To fix this properly, I extracted the method into a dedicated `OrderPersistenceService` with `@Service` and `@Transactional` annotations.Now,`OrderService` delegates to it. No logic is changed, just a clearer proxy boundary so the transaction can be applied.
Existing tests confirm that order item, payment, and stock updates persist atomically.

farhanaahmed:
Hello Suno, 
I hope you’re doing well! 😄 
I’ve carefully gone through your comments and tried to incorporate your feedback as much as I could across the codebase.
Could you please review the changes when you have time? 
Thanks a lot for your time and guidance. 🙏

build.gradle.kts
Wordbe: Okay Thank you for your exaplanation.

scripts/deploy.sh
Wordbe: It's enough. Thank you!

src/main/kotlin/ecommerce/config/CorsConfig.kt
Wordbe: Great Job!

src/main/kotlin/ecommerce/controller/OrderController.kt
Wordbe: Yes, that looks much better now!

src/main/kotlin/ecommerce/dto/PlaceOrderRequest.kt
Wordbe: I understand that you used it for testing. Yes, let’s keep it as is. Thank you. 🙏

src/main/kotlin/ecommerce/entity/Payment.kt
Wordbe: Ok. Great!

src/main/kotlin/ecommerce/entity/OrderItem.kt
Wordbe: Thank you for reflecting the suggestions. 🙌

src/main/kotlin/ecommerce/entity/Product.kt
Wordbe: The `hashCode` of a `String` seems to roughly iterate over the characters, multiply each by 31, and accumulate the sum to get the final hash value. It’s not simply returning 31. For a `Product` with `id = null`, always returning 31 could be problematic.

src/main/kotlin/ecommerce/repository/OptionRepositoryJpa.kt
Wordbe: Yes, I think it’s fine for our project demo. But as order requests increase, this could become an issue, so I asked to encourage you to think about the trade-offs. You explained it well, thank you. 🙏

src/main/kotlin/ecommerce/service/OrderService.kt
Wordbe: Great work, thank you.

src/main/kotlin/ecommerce/service/OrderService.kt
Wordbe: Goood.

src/test/kotlin/ecommerce/controller/OrderControllerTest.kt
Wordbe: So you made it for more Kotlin-style code. Good, thank you. 🙂

Wordbe (approved):
Hello Farhana! 🙂

You’ve reflected the review really well.
It was great to see you carefully read through all the feedback and clearly share your thoughts, it made giving feedback easier. Thank you.
At this point, I’ll approve and merge this PR.
With this, you’ve completed all the missions, congratulations! 🎉
Wishing you the best as you continue your journey. Thank you!

PR #28, opened 2025-08-16
[spring-ecommerce-order - Step-2] Submit saragrosser's mission

saragrosser:
PR: Step 2 – External API, Orders, and Deployment

This PR builds on Step 1 by extending the service with payment processing, order management, and a full deployment flow. It represents the transition from purely local persistence into integrating with an external API (Stripe), handling real-world concerns like stock management, error propagation, and CORS/security configuration.

The work for this step was divided into three parts:

⸻

Step 2.1 – External API Integration
• Introduced StripeProperties bound to stripe.secret-key.
• Implemented a StripeClient using Spring RestClient to interact with Stripe’s Payment Create API.
• Added the Place Order flow:
• Decreases product option stock after purchase.
• Removes purchased items from the cart.
• If payment approval fails, the error is surfaced clearly to the user.
• Wrote acceptance & service tests to validate payment handling and failure scenarios.

⸻

Step 2.2 – Orders
• Implemented an Orders feature to persist and query order history.
• Orders include:
• Date and time of purchase.
• Order status.
• Purchased items.
• Checkout Session ID (from Stripe).
• Payment amount.
• Added repository and service tests to verify order creation and retrieval.

⸻

Step 2.3 – Deployment
• Wrote a deployment script (scripts/deploy.sh) to automate building, uploading, and restarting the service on AWS EC2.
• Configured .env and DotenvConfig so secrets like STRIPE_SECRET_KEY and CORS origins are injected safely without being committed.
• Added CORS configuration to allow calls from a different client origin.
• Successfully deployed the app to AWS EC2, verified via /api/products endpoint.

⸻

Challenges Faced
• Managing error handling around Stripe payment failures and surfacing useful error messages.
• Handling cart updates atomically when placing orders.
• Getting deployment scripts right to run on port 80 securely with .env loading.

⸻

Key Learnings
• Deepened understanding of external API integration and error handling patterns.
• Learned how to tie business logic together: payment, cart updates, and stock management in one flow.
• Strengthened knowledge of deployment pipelines: .env handling, CORS/security, and process management on AWS EC2.

⸻

✅ With this step, the service is no longer local-only—it now integrates with external systems, persists orders, and runs in a deployed environment ready to be consumed by the client.

Thank you very much for reviewing—I welcome all suggestions and improvements 🙏


.env.example
```diff
@@ -0,0 +1,14 @@
+# --- Stripe ---
+STRIPE_SECRET_KEY=
```
da-nyee, line 2: `.env` 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.http.MediaType
+import org.springframework.stereotype.Component
+import org.springframework.web.client.RestClient
+import java.net.SocketTimeoutException
+
+@Component
+class StripeClient(
+    private val stripeProperties: StripeProperties,
+) {
+    private val restClient = RestClient.create()
```
da-nyee: What about setting timeouts for the client?

da-nyee: Have you heard of `WebClient`? How is it different from this?

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,50 @@
...
+        currency: String = "usd",
+    ): StripePaymentResponse {
+        val body =
+            listOf(
+                "amount=$amount",
+                "currency=$currency",
+                "confirm=true",
+                "automatic_payment_methods[enabled]=true",
+                "automatic_payment_methods[allow_redirects]=never",
+            ).joinToString("&")
```
da-nyee: Would it make sense to introduce a body object?

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,50 @@
...
+                .onStatus({ it.is4xxClientError }) { _, _ ->
+                    throw PaymentClientException("Invalid request to Stripe")
+                }
+                .onStatus({ it.is5xxServerError }) { _, _ ->
+                    throw PaymentServerException("Stripe server error")
+                }
+                .body(StripePaymentResponse::class.java)!!
+        } catch (e: SocketTimeoutException) {
+            throw PaymentServerException("Stripe request timed out") as Throwable
+        }
```
da-nyee: What about logging an exception with `KotlinLogging` or another logger?

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,50 @@
...
+                .body(body)
+                .retrieve()
+                .onStatus({ it.is4xxClientError }) { _, _ ->
+                    throw PaymentClientException("Invalid request to Stripe")
+                }
+                .onStatus({ it.is5xxServerError }) { _, _ ->
+                    throw PaymentServerException("Stripe server error")
+                }
+                .body(StripePaymentResponse::class.java)!!
+        } catch (e: SocketTimeoutException) {
```
da-nyee: You might consider handling exceptions more precisely, following the [docs](https://docs.stripe.com/error-handling?lang=java#payment-errors) 👀

src/test/kotlin/ecommerce/client/StripeClientTest.kt
```diff
@@ -0,0 +1,20 @@
...
+import ecommerce.exception.PaymentClientException
+import org.assertj.core.api.Assertions.assertThatThrownBy
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+
+@SpringBootTest
+class StripeClientTest {
+    @Autowired
+    lateinit var stripeClient: StripeClient
```
da-nyee: I'd recommend mocking the client and stubbing the API call 👀

src/test/kotlin/ecommerce/client/StripeClientTest.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    @Test
+    fun `createPaymentIntent with invalid key maps to PaymentClientException`() {
```
da-nyee: What about testing success cases and more detailed failure cases?

scripts/deploy.sh
```diff
@@ -0,0 +1,45 @@
...
+cd "$DEPLOY_DIR"
+
+
+nohup java -jar "$TARGET" \
+  --spring.profiles.active=prod \
+  > app.out 2>&1 &
+
+NEW_PID=$!
+echo "[deploy] Started PID=${NEW_PID}"
+echo "[deploy] Logs: tail -f ${DEPLOY_DIR}/app.out"
```
da-nyee, line 45: Deployment script 👍

da-nyee (changes requested):
Hi Sari! 🙌

You're doing great so far 💯 
There are a few things that could be improved though, so I've left some comments — please take a look when you have time.

If anything's unclear, just let me know 🙇‍♀️

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: I’ve updated the implementation by setting the timeouts directly inside StripeClient.kt. Regarding your second point, yes, I’ve looked into WebClient. From what I understand, it’s a non-blocking, reactive alternative to RestClient that supports more advanced configurations and async use cases. For our current use case (simple synchronous requests with Stripe), I thought RestClient would be enough.

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: Thanks for the suggestion 🙏 I added logging with KotlinLogging around the timeout/connection handling. Now whenever a request fails or times out, the exception details get logged before we rethrow it as a PaymentServerException. This should make troubleshooting Stripe issues much easier.

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: Makes sense  I updated the code to follow the Stripe documentation. Instead of using generic onStatus handlers, I’m now catching HttpClientErrorException and HttpServerErrorException directly. That lets me parse Stripe’s error JSON and pass along the actual message in our exceptions. It feels more precise and keeps the error context intact.

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: I refactored that to use a dedicated StripePaymentIntentRequest data class for the request body. It's definitely a lot cleaner and safer now since it's type-safe, so thanks for the great suggestion!

src/test/kotlin/ecommerce/client/StripeClientTest.kt
saragrosser: Thank you Danyee I have replaced integration test with unit test for StripeClient it took me a while to refactor but i finally was able to add more detailed cases

src/test/kotlin/ecommerce/client/StripeClientTest.kt
saragrosser: I have refactor the whole file thank you for motivating me to create better  tests

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,87 @@
...
+    private val objectMapper: ObjectMapper,
+    private val restClient: RestClient = defaultRestClient(),
+) {
+    companion object {
+        private fun defaultRestClient(): RestClient =
+            RestClient.builder()
+                .requestFactory(
+                    SimpleClientHttpRequestFactory().apply {
+                        setConnectTimeout(Duration.ofSeconds(3).toMillis().toInt())
+                        setReadTimeout(Duration.ofSeconds(10).toMillis().toInt())
```
da-nyee: 👍

da-nyee: Just wondering — is there any particular reason you set the timeouts to 3s and 10s respectively?

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,87 @@
...
+        currency: String = "usd",
+    ): StripePaymentResponse {
+        // Stripe expects application/x-www-form-urlencoded
+        val form =
+            LinkedMultiValueMap<String, String>().apply {
+                add("amount", amount.toString()) // smallest currency unit
+                add("currency", currency)
+                // Optionally enable automatic payment methods:
+                // add("automatic_payment_methods[enabled]", "true")
+            }
```
da-nyee: I guess you could use `StripePaymentIntentRequest` here with `toString()` or something similar 🤔

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.springframework.http.client.SimpleClientHttpRequestFactory
+import org.springframework.stereotype.Component
+import org.springframework.util.LinkedMultiValueMap
+import org.springframework.web.client.HttpClientErrorException
+import org.springframework.web.client.HttpServerErrorException
+import org.springframework.web.client.ResourceAccessException
+import org.springframework.web.client.RestClient
+import java.time.Duration
+
+private val logger = KotlinLogging.logger {}
```
da-nyee, line 21: 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,87 @@
...
+                .uri("https://api.stripe.com/v1/payment_intents")
+                .header(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secretKey}")
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                .body(form)
+                .retrieve()
+                .body(StripePaymentResponse::class.java)!!
+        } catch (e: HttpClientErrorException) {
+            // 4xx from Stripe
+            val msg = parseStripeMessage(e.responseBodyAsString) ?: "Invalid request to Stripe."
+            throw PaymentClientException("Stripe Client Error: $msg")
```
da-nyee, line 69: What about logging something before throwing the exception? 👀

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,87 @@
...
+        } catch (e: HttpServerErrorException) {
```
da-nyee: 👍

src/test/kotlin/ecommerce/client/StripeClientTest.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.springframework.web.client.RestClient
+import org.springframework.web.client.RestClient.RequestBodySpec
+import org.springframework.web.client.RestClient.RequestBodyUriSpec
+import org.springframework.web.client.RestClient.ResponseSpec
+
+@ExtendWith(org.mockito.junit.jupiter.MockitoExtension::class)
+class StripeClientTest {
+    private val stripeProps = StripeProperties(secretKey = "sk_test_invalid")
+    private val objectMapper = jacksonObjectMapper()
+    private val restClient: RestClient = mock()
```
da-nyee, line 33: 👍

src/test/kotlin/ecommerce/client/StripeClientTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        stubCommonChain()
+        val fakeResponse =
+            StripePaymentResponse(
+                id = "pi_123",
+                status = "succeeded",
+                amount = 1000L,
+                currency = "usd",
+                lastPaymentError = null,
+            )
+        whenever(responseSpec.body(StripePaymentResponse::class.java)).thenReturn(fakeResponse)
```
da-nyee, line 65: 👍

src/test/kotlin/ecommerce/client/StripeClientTest.kt
```diff
@@ -0,0 +1,127 @@
...
+        whenever(responseSpec.body(StripePaymentResponse::class.java)).thenThrow(httpEx)
+
+        val client = buildClient()
+
+        assertThatThrownBy { client.createPaymentIntent(1000, "usd") }
+            .isInstanceOf(PaymentServerException::class.java)
+    }
+
+    @Test
+    fun `IO timeout maps to PaymentServerException`() {
```
da-nyee, line 117: Test cases in more detail 👍

da-nyee (changes requested):
Hi Sari! 🙌

It seems almost ready to merge, but I'll mark this as `Request changes` because there are a few more comments and some small things to improve.

Just one more step to go, so keep it up 💪

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: The idea was to strike a balance between responsiveness and resilience.
	•	3s connect timeout → Stripe is a very reliable API, so establishing a TCP connection should normally be almost instant. Setting it to 3s avoids hanging too long if there’s a network issue.
	•	10s read timeout → Once connected, the actual request/response can take a bit longer (e.g. if Stripe is processing something or the network is briefly slow). 10s felt like a safe upper bound to allow for that without keeping the user waiting forever.

That said, I’ve now moved these values into StripeProperties so they’re configurable. We can adjust them anytime based on real-world performance or team preference.

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: thank you! I refactored the code to build the request using StripePaymentIntentRequest and then convert it to a form map with a small helper. That way i'm not not duplicating fields, and the request logic stays consistent with the rest of the DTOs.

src/main/kotlin/ecommerce/client/StripeClient.kt
saragrosser: Sorry about that, I refactored and added logging for the different error cases. 4xx responses are logged as warnings with Stripe’s error message, while 5xx and timeout errors are logged as errors.

src/test/kotlin/ecommerce/client/StripeClientTest.kt
saragrosser: thank you!

src/main/kotlin/ecommerce/config/StripeProperties.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.annotation.Configuration
+
+@ConfigurationProperties(prefix = "stripe")
+data class StripeProperties(
+    val secretKey: String,
+    val connectTimeoutMs: Long = 3000,
+    val readTimeoutMs: Long = 10000,
```
da-nyee, line 11: Good place 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private fun StripePaymentIntentRequest.toForm(): LinkedMultiValueMap<String, String> =
+        LinkedMultiValueMap<String, String>().apply {
+            add("amount", amount.toString())
+            add("currency", currency)
+            add("confirm", confirm.toString())
+            add("automatic_payment_methods[enabled]", automaticPaymentMethodsEnabled.toString())
+            add("automatic_payment_methods[allow_redirects]", automaticPaymentMethodsAllowRedirects)
+        }
```
da-nyee, line 90: 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,102 @@
...
+                .header(HttpHeaders.AUTHORIZATION, "Bearer ${stripeProperties.secretKey}")
+                .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                .body(form)
+                .retrieve()
+                .body(StripePaymentResponse::class.java)!!
+        } catch (e: HttpClientErrorException) {
+            // 4xx from Stripe
+            val raw = e.responseBodyAsString
+            val msg = parseStripeMessage(raw) ?: "Invalid request to Stripe."
+            logger.warn { "Stripe 4xx: status=${e.statusCode.value()}, message=$msg, body=$raw" }
```
da-nyee, line 68: Using warn level for 4XX and error level for 5XX 👍

da-nyee (approved):
Hi Sari! 🙌

It's time to wrap up the last mission — you made it 💯 
I really hope the feedback was helpful for you, and that you continue to keep your interest in software development.

Congratulations on your graduation! 🥳

PR #30, opened 2025-08-18
[spring-ecommerce-order - Step-2] Submit Apoorva-S-Natesh's mission

Apoorva-S-Natesh:
Hi, 
This PR introduces Docker-based deployment for the Order Service, alongside Stripe-based payments, basic failure handling, Order/Payment domain modeling, stock updates, cart cleanup, and CORS configuration. It replaces the previous direct JAR deployment with a containerized approach to improve reproducibility, environment consistency, and deployment automation.

### 1) Stripe Payment Integration
- Uses Stripe Payment Intents API to create a payment.  
- Requests sent with `confirm=true`, so no separate confirmation step on our side.  
- Sandbox secret key used for development/testing.  

### 2) Payment Failure Handling
- Exceptions from Stripe client caught in `PaymentService`.  
- On failure, a `PaymentResponse` with `status=failed` and error message is returned
- 
### 3) Order and Payment Entities (separate tables)
- Payments are decoupled from orders for clearer lifecycle boundaries and support for retries/refunds.  

### 4) Stock Update
- On successful payment:  
  - Order is marked as succeeded.  
  - Stock of purchased product option is decreased by the order quantity.  

### 5) Cart Cleanup
- Purchased items are removed from the cart after order completion.  

### 6) CORS Configuration (via WebMvcConfigurer)
- Allowed origins: `*` (temporary, for development).  
- Allowed methods: GET, POST, PUT, DELETE, OPTIONS.  
- Allowed headers: `*`  
- Max age: 3600 seconds.  

### 7) Dockerized Deployment
- Created a `Dockerfile` to build the Kotlin app into a lightweight Docker image.  
- Includes configuration to securely handle `.env` variables without baking secrets into the image.  
- Deployment script (`deploy.sh`) automates container build, stop/remove old container, and run new container.  

### 8) Deployment to AWS EC2 with Docker
- Built the Docker image locally (avoiding heavy Gradle builds on EC2).  
- Pushes and runs the container on the EC2 instance.  
- `.env` file mounted as a volume, keeping secrets secure and out of the image.  
- Automated deployment script (`deploy.sh`) does the following:  
  1. Stops any running container for this service.  
  2. builds the latest Docker image.  
  3. Runs the new container in detached mode with proper environment variables.  

Next Steps / Gaps
Consider adding HTTPS and Docker health checks.
Fine-tune CORS and security headers for production.
Add CI/CD integration for fully automated builds and deployments.

Thank you for your feedback and time

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,8 @@
 .git
 .gradle
-#build
+build/
 out
 .idea
 *.iml
 .env
+!build/libs/spring-ecommerce-0.0.1-SNAPSHOT.jar
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.git
+.gradle
+#build
+out
+.idea
+*.iml
+.env
--- a/README.md
+++ b/README.md
@@ -62,3 +62,65 @@
 - [x] **Pagination** - Add `Pageable` support to controllers
 - [x] **Inventory Management** - Stock tracking and validation
 - [ ] **Performance** - Query optimization and caching
+
+## Place Order and Payment
+- [] **Setup Stripe** - add secret-key to application-properties or application.yml(External API)
+  - []key hidden from gitHub codeBase (environment variable management in springBoot)
+  - []check key is correctly loaded with StripeProperties (@ConfigurationProperties)
+  - []@EnableConfigurationProperties(StripeProperties::class) on class Application
+- [] **Entity** - OrderItem (id, quantity, price, ProductOption @ManyToOne - one directional)
+- [] **DTO** - PlaceOrderRequest(productOptionId, quantity,paymentDetails)
+        -[] Validate (valid ProductOptionID, positive quantity, notBlank payment) 
+  - PaymentResponse 
+- [] **Repository**
+- [] **Service** - StripeClient - 
+  - [] call Stripe Payment Intents API(/v1/payment_intents)
+  - [] Parameters - amount, currency, payment_method, confirm=true
+  - [] Order amount based on ProductOption Price(from Product) and quantity
+  - [] Handle API response to extract checkout session ID (payment Status)
+  - [] payment successful -> Decrease ProductOption quantity by ordered quantity 
+  - [] payment successful -> Remove CartItem from member's cart (CartItemService.deleteCartItemById)
+  - [] payment successful -> Update Cart Quantity (CartItem removed)
+  - [] Set timeouts for Stripe API calls using RestClient to avoid holding database connections during slow responses.
+  - [] Handle transaction propagation (REQUIRED for main transaction)
+- [] **Controller** - placeOrder endpoints(POST /api/orders/  @RequestBody placeOrderRequest) 
+  - verify endpoint requires authentication
+- [] **Tests** - Unit and integration tests
+  - [] Test API call with valid test cards from Stripe's documentation. (Verify Response fields -id, status)
+  - [] declined payments (insufficient funds, invalid card)
+- [] **Exception Handling**
+  - [] Exceptions from StripeClient - map to specific reasons (expired session, Invalid payment Method)
+  - [] Return user friendly error messages - ResponseEntity.badRequest().body(ErrorResponse(message))
+- [] Validate stock availability (fetch user's cart, CartItem, ProductOption) - decide where should this validation happen
+- [] @Transactional (Stock update and CartItem removal)
+- [] Server side polling to handle payment - Synchronous status check
+
+## Orders - Store and display order details
+- [] **Entity** 
+  - [] Order (OrderItem @OneToMany, memberId or Member @ManyToOne, Payment @OneToOne)
+  - [] Member (Order @OneToMany)
+  - [] Payment (id, checkout_session_id, amount, currency, status, payment_method, created, and last_payment_error, linked to Order.)
+- [] **DTO** - OrderResponse (id, LocalDateTme, status, purchasedItem, checkoutSessionId, amount)
+- [] **Repository** 
+- [] **Service** - Create an order, process payment, update stock, remove cart items
+- [] **Controller** 
+- GET /api/orders - return a list of orders for authenticated user, 
+- GET /api/orders/{orderId} - Retrieve details of a specific order for the authenticated user.
+- GET /api/admin/orders - Admin can view all orders
+ - [] **Tests** - Integration tests (no orders, invalid user, failed payment)
+
+## Deployment
+- [] Deployment Script (shell script) - automate deployment
+  - [] steps to copy the build, start application with nohup
+- [] Configure environment variable - secure stripe API key
+- [] Configure CORS to allow client requests(CorsConfiguration)
+- [] Set allowedOrigins to * allow methods (GET, POST, etc.), set maxAge to 1800 seconds.
+- [] Test CORS with AcceptanceTest to verify headers
+- [] Deploy to AWS EC2
+
+## Next Steps - Webhook
+- [] setup Webhook endpoint - handle asynchronous payment flows (WebHook controller POST /api/webhooks/stripe)
+- [] Configure Stripe dashboard - add new webhook endpoint with the URL
+- [] Verify webhook Signature (Stripe signature - verify authenticity)
+- [] Process webhook events - parse JSON payload - exteact event type and data
+- [] test (payment_intent.succeeded and payment_intent.failed)
--- a/README.md
+++ b/README.md
@@ -64,50 +64,50 @@
 - [ ] **Performance** - Query optimization and caching
 
 ## Place Order and Payment
-- [] **Setup Stripe** - add secret-key to application-properties or application.yml(External API)
-  - []key hidden from gitHub codeBase (environment variable management in springBoot)
-  - []check key is correctly loaded with StripeProperties (@ConfigurationProperties)
-  - []@EnableConfigurationProperties(StripeProperties::class) on class Application
-- [] **Entity** - OrderItem (id, quantity, price, ProductOption @ManyToOne - one directional)
-- [] **DTO** - PlaceOrderRequest(productOptionId, quantity,paymentDetails)
-        -[] Validate (valid ProductOptionID, positive quantity, notBlank payment) 
-  - PaymentResponse 
-- [] **Repository**
-- [] **Service** - StripeClient - 
-  - [] call Stripe Payment Intents API(/v1/payment_intents)
-  - [] Parameters - amount, currency, payment_method, confirm=true
-  - [] Order amount based on ProductOption Price(from Product) and quantity
-  - [] Handle API response to extract checkout session ID (payment Status)
-  - [] payment successful -> Decrease ProductOption quantity by ordered quantity 
-  - [] payment successful -> Remove CartItem from member's cart (CartItemService.deleteCartItemById)
-  - [] payment successful -> Update Cart Quantity (CartItem removed)
+- [x] **Setup Stripe** - add secret-key to application-properties or application.yml(External API)
+  - [x] key hidden from gitHub codeBase (environment variable management in springBoot)
+  - [] check key is correctly loaded with StripeProperties (@ConfigurationProperties)
+  - [x] @EnableConfigurationProperties(StripeProperties::class) on class Application
+- [x] **Entity** - OrderItem (id, quantity, price, ProductOption @ManyToOne - one directional)
+- [x] **DTO** - PlaceOrderRequest(productOptionId, quantity,paymentDetails)
+        -[x] Validate (valid ProductOptionID, positive quantity, notBlank payment) 
+  - [x] PaymentResponse 
+- [x] **Repository**
+- [x] **Service** - StripeClient - 
+  - [x] call Stripe Payment Intents API(/v1/payment_intents)
+  - [x] request (Parameters - amount, currency, payment_method, confirm=true)
+  - [x] Order amount based on ProductOption Price(from Product) and quantity
+  - [x] Handle API response to extract checkout session ID (payment Status)
+  - [x] payment successful -> Decrease ProductOption quantity by ordered quantity 
+  - [x] payment successful -> Remove CartItem from member's cart (CartItemService.deleteCartItemById)
+  - [x] payment successful -> Update Cart Quantity (CartItem removed)
   - [] Set timeouts for Stripe API calls using RestClient to avoid holding database connections during slow responses.
-  - [] Handle transaction propagation (REQUIRED for main transaction)
-- [] **Controller** - placeOrder endpoints(POST /api/orders/  @RequestBody placeOrderRequest) 
+  - [x] Handle transaction propagation (REQUIRED for main transaction)
+- [x] **Controller** - placeOrder endpoints(POST /api/orders/  @RequestBody placeOrderRequest) 
   - verify endpoint requires authentication
-- [] **Tests** - Unit and integration tests
-  - [] Test API call with valid test cards from Stripe's documentation. (Verify Response fields -id, status)
-  - [] declined payments (insufficient funds, invalid card)
-- [] **Exception Handling**
-  - [] Exceptions from StripeClient - map to specific reasons (expired session, Invalid payment Method)
-  - [] Return user friendly error messages - ResponseEntity.badRequest().body(ErrorResponse(message))
-- [] Validate stock availability (fetch user's cart, CartItem, ProductOption) - decide where should this validation happen
-- [] @Transactional (Stock update and CartItem removal)
-- [] Server side polling to handle payment - Synchronous status check
+- [x] **Tests** - Unit and integration tests
+  - [x] Test API call with valid test cards from Stripe's documentation. (Verify Response fields -id, status)
+  - [x] declined payments (insufficient funds, invalid card)
+- [x] **Exception Handling**
+  - [x] Exceptions from StripeClient - map to specific reasons (expired session, Invalid payment Method)
+  - [x] Return user friendly error messages - ResponseEntity.badRequest().body(ErrorResponse(message))
+- [x] Validate stock availability (fetch user's cart, CartItem, ProductOption) - decide where should this validation happen
+- [x] @Transactional (Stock update and CartItem removal)
+- [x] Server side polling to handle payment - Synchronous status check
 
 ## Orders - Store and display order details
-- [] **Entity** 
-  - [] Order (OrderItem @OneToMany, memberId or Member @ManyToOne, Payment @OneToOne)
-  - [] Member (Order @OneToMany)
-  - [] Payment (id, checkout_session_id, amount, currency, status, payment_method, created, and last_payment_error, linked to Order.)
-- [] **DTO** - OrderResponse (id, LocalDateTme, status, purchasedItem, checkoutSessionId, amount)
-- [] **Repository** 
-- [] **Service** - Create an order, process payment, update stock, remove cart items
-- [] **Controller** 
+- [x] **Entity** 
+  - [x] Order (OrderItem @OneToMany, memberId, Payment @OneToOne)
+  - [x] Payment (id, checkout_session_id, amount, currency, status, payment_method, created, and last_payment_error, linked to Order.)
+- [x] **DTO** 
+  -   [x] - OrderResponse (id, LocalDateTme, status, purchasedItem, checkoutSessionId, amount)
+  -   [x] - OrderItemResponse (quantity, price, productOptionId)
+- [x] **Service** - Create an order, process payment, update stock, remove cart items
+- [x] **Controller** 
 - GET /api/orders - return a list of orders for authenticated user, 
 - GET /api/orders/{orderId} - Retrieve details of a specific order for the authenticated user.
 - GET /api/admin/orders - Admin can view all orders
- - [] **Tests** - Integration tests (no orders, invalid user, failed payment)
+ - [x] **Tests** - Integration tests (no orders, invalid user, failed payment)
 
 ## Deployment
 - [] Deployment Script (shell script) - automate deployment
--- a/README.md
+++ b/README.md
@@ -110,13 +110,13 @@
  - [x] **Tests** - Integration tests (no orders, invalid user, failed payment)
 
 ## Deployment
-- [] Deployment Script (shell script) - automate deployment
-  - [] steps to copy the build, start application with nohup
-- [] Configure environment variable - secure stripe API key
-- [] Configure CORS to allow client requests(CorsConfiguration)
-- [] Set allowedOrigins to * allow methods (GET, POST, etc.), set maxAge to 1800 seconds.
-- [] Test CORS with AcceptanceTest to verify headers
-- [] Deploy to AWS EC2
+- [x] Deployment Script (shell script) - automate deployment
+  - [X] steps to copy the build, start application with nohup
+- [X] Configure environment variable - secure stripe API key
+- [X] Configure CORS to allow client requests(CorsConfiguration)
+- [X] Set allowedOrigins to * allow methods (GET, POST, etc.), set maxAge to 1800 seconds.
+- [X] Test CORS with AcceptanceTest to verify headers
+- [X] Deploy to AWS EC2
 
 ## Next Steps - Webhook
 - [] setup Webhook endpoint - handle asynchronous payment flows (WebHook controller POST /api/webhooks/stripe)
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@
 ## Place Order and Payment
 - [x] **Setup Stripe** - add secret-key to application-properties or application.yml(External API)
   - [x] key hidden from gitHub codeBase (environment variable management in springBoot)
-  - [] check key is correctly loaded with StripeProperties (@ConfigurationProperties)
+  - [x] check key is correctly loaded with StripeProperties (@ConfigurationProperties)
   - [x] @EnableConfigurationProperties(StripeProperties::class) on class Application
 - [x] **Entity** - OrderItem (id, quantity, price, ProductOption @ManyToOne - one directional)
 - [x] **DTO** - PlaceOrderRequest(productOptionId, quantity,paymentDetails)
@@ -81,7 +81,7 @@
   - [x] payment successful -> Decrease ProductOption quantity by ordered quantity 
   - [x] payment successful -> Remove CartItem from member's cart (CartItemService.deleteCartItemById)
   - [x] payment successful -> Update Cart Quantity (CartItem removed)
-  - [] Set timeouts for Stripe API calls using RestClient to avoid holding database connections during slow responses.
+  - [x] Set timeouts for Stripe API calls using RestClient to avoid holding database connections during slow responses.
   - [x] Handle transaction propagation (REQUIRED for main transaction)
 - [x] **Controller** - placeOrder endpoints(POST /api/orders/  @RequestBody placeOrderRequest) 
   - verify endpoint requires authentication
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,7 +30,6 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("io.jsonwebtoken:jjwt-api:0.12.5")
     implementation("org.mindrot:jbcrypt:0.4")
-    implementation("org.springframework.boot:spring-boot-starter-webflux")
     runtimeOnly("io.jsonwebtoken:jjwt-impl:0.12.5")
     runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.12.5")
     runtimeOnly("com.h2database:h2")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -30,6 +30,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("io.jsonwebtoken:jjwt-api:0.12.5")
     implementation("org.mindrot:jbcrypt:0.4")
+    implementation("org.springframework.boot:spring-boot-starter-webflux")
     runtimeOnly("io.jsonwebtoken:jjwt-impl:0.12.5")
     runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.12.5")
     runtimeOnly("com.h2database:h2")
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -36,6 +36,7 @@ dependencies {
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
     testImplementation("io.rest-assured:rest-assured:$assureVersion")
+    testImplementation("org.mockito.kotlin:mockito-kotlin:5.3.1")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
     developmentOnly("org.springframework.boot:spring-boot-devtools")
 }
--- a/src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
+++ b/src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
@@ -88,17 +88,19 @@ class GlobalControllerAdvice {
                         "Authentication with the card issuer is required.",
                     )
 
-                "invalid_request_error" -> Triple(
-                    "INVALID_REQUEST",
-                    HttpStatus.BAD_REQUEST,
-                    e.message ?: "Invalid request parameters."
-                )
+                "invalid_request_error" ->
+                    Triple(
+                        "INVALID_REQUEST",
+                        HttpStatus.BAD_REQUEST,
+                        e.message ?: "Invalid request parameters.",
+                    )
 
-                "rate_limit" -> Triple(
-                    "RATE_LIMIT",
-                    HttpStatus.TOO_MANY_REQUESTS,
-                    "Too many requests. Please try again later."
-                )
+                "rate_limit" ->
+                    Triple(
+                        "RATE_LIMIT",
+                        HttpStatus.TOO_MANY_REQUESTS,
+                        "Too many requests. Please try again later.",
+                    )
 
                 "api_connection_error" ->
                     Triple(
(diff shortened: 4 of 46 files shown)


.dockerignore
```diff
@@ -0,0 +1,7 @@
+.git
+.gradle
+#build
```
Wordbe: Try removing the comments, and if you need the jar file inside build, it would be better to include only that file using `!` 🙂

README.md
```diff
@@ -62,3 +62,65 @@
 - [x] **Pagination** - Add `Pageable` support to controllers
 - [x] **Inventory Management** - Stock tracking and validation
 - [ ] **Performance** - Query optimization and caching
+
+## Place Order and Payment
+- [x] **Setup Stripe** - add secret-key to application-properties or application.yml(External API)
+  - [x] key hidden from gitHub codeBase (environment variable management in springBoot)
+  - [] check key is correctly loaded with StripeProperties (@ConfigurationProperties)
```
Wordbe: If you’ve finished, please check it off! ✅

README.md
```diff
@@ -62,3 +62,65 @@
...
+- [x] **Repository**
+- [x] **Service** - StripeClient - 
+  - [x] call Stripe Payment Intents API(/v1/payment_intents)
+  - [x] request (Parameters - amount, currency, payment_method, confirm=true)
+  - [x] Order amount based on ProductOption Price(from Product) and quantity
+  - [x] Handle API response to extract checkout session ID (payment Status)
+  - [x] payment successful -> Decrease ProductOption quantity by ordered quantity 
+  - [x] payment successful -> Remove CartItem from member's cart (CartItemService.deleteCartItemById)
+  - [x] payment successful -> Update Cart Quantity (CartItem removed)
+  - [] Set timeouts for Stripe API calls using RestClient to avoid holding database connections during slow responses.
```
Wordbe: Same here.

build.gradle.kts
```diff
@@ -30,6 +30,7 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("io.jsonwebtoken:jjwt-api:0.12.5")
     implementation("org.mindrot:jbcrypt:0.4")
+    implementation("org.springframework.boot:spring-boot-starter-webflux")
```
Wordbe: Are you using WebFlux in this project?

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -68,4 +71,61 @@ class GlobalControllerAdvice {
...
+
+                else ->
+                    Triple(
+                        "STRIPE_ERROR",
+                        HttpStatus.BAD_REQUEST,
+                        e.message ?: "An error occurred during payment processing.",
+                    )
+            }
+        return createErrorResponse(e, errorCode, status, listOf(message))
+    }
```
Wordbe: Is there a reason why `authentication_required`, `invalid_request_error`, `rate_limit`, and `api_connection_error` weren’t made into an enum? I think it would look cleaner that way. Also, it might be clearer if we don’t use `Triple`! 🙂

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -68,4 +71,61 @@ class GlobalControllerAdvice {
...
+
+    @ExceptionHandler(OrderProcessingException::class)
+    fun handleOrderProcessingException(e: OrderProcessingException): ResponseEntity<ErrorResponse> {
+        val status =
+            when ("BAD_REQUEST") {
+                in DeclineCode.entries.map { it.stripeCode } -> HttpStatus.BAD_REQUEST
+                else -> HttpStatus.BAD_REQUEST
+            }
```
Wordbe: Since it always returns `HttpStatus.BAD_REQUEST`, it doesn’t seem like there’s a reason for classifying the cases. 🔍

src/main/kotlin/ecommerce/config/ApplicationConfig.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.config
+
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+class ApplicationConfig
```
Wordbe, line 6: Is there a place where this is being used? 🤔

src/main/kotlin/ecommerce/config/CorsConfig.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+
+@Configuration
+class CorsConfig {
+    @Bean
+    fun corsConfigurer(): WebMvcConfigurer {
+        return object : WebMvcConfigurer {
+            override fun addCorsMappings(registry: CorsRegistry) {
+                registry.addMapping("/**")
+                    .allowedOrigins("*")
```
Wordbe: Blocking the allowance of all origins could help strengthen security. 🔒

src/main/kotlin/ecommerce/controller/OrderController.kt
```diff
@@ -0,0 +1,67 @@
...
+        user: AuthenticatedUser,
+    ): ResponseEntity<Any> {
+        return try {
+            val order = orderService.placeOrder(placeOrderRequest, user.userId)
+            ResponseEntity.ok(order)
+        } catch (e: OrderProcessingException) {
+            ResponseEntity.badRequest().body(ErrorResponse("BAD_REQUEST", e.message!!))
+        } catch (e: Exception) {
+            ResponseEntity.badRequest().body(ErrorResponse("BAD_REQUEST", e.message ?: "Unknown error"))
+        }
```
Wordbe: Try delegating exception handling to `GlobalControllerAdvice`. 🙂

src/main/kotlin/ecommerce/dto/OrderDtos.kt
```diff
@@ -0,0 +1,70 @@
...
+class OrderItemResponse(
+    val quantity: Int,
+    val price: Double,
+    val productOptionId: Long,
+)
+
+class OrderResponse(
+    val id: Long?,
+    val orderDate: LocalDateTime?,
+    val status: String?,
```
Wordbe: If `status` is defined, try using an enum for it. 🙂

src/main/kotlin/ecommerce/interceptor/AuthInterceptor.kt
```diff
@@ -17,21 +18,30 @@ class AuthInterceptor(private val tokenService: TokenService) : HandlerIntercept
         response: HttpServletResponse,
         handler: Any,
     ): Boolean {
+        logger.debug("Processing request: ${request.requestURI}, Authorization: ${request.getHeader(AUTH_HEADER)}")
         val token =
-            extractToken(request)
-                ?: throw AuthenticationException("Missing or invalid Authorization header")
+            extractToken(request) ?: run {
+                logger.warn("Missing or invalid Authorization header")
+                throw AuthenticationException("Missing or invalid Authorization header")
```
Wordbe: The logging and exceptions seem redundant. If you add common logging in `GlobalControllerAdvice`, it would be nice to just throw exceptions and have them logged automatically. Logging exceptions is important for operations. 
Also, since this is logic that throws exceptions, it would be better to log it at the error level. 🔥

src/main/kotlin/ecommerce/interceptor/AuthInterceptor.kt
```diff
@@ -17,21 +18,30 @@ class AuthInterceptor(private val tokenService: TokenService) : HandlerIntercept
...
+            extractUserId(claims) ?: run {
+                logger.warn("Invalid token payload: missing userId")
+                throw AuthenticationException("Invalid token payload")
+            }
 
+        logger.debug("Authenticated userId: $userId")
         storeAuthenticatedUser(request, claims, userId)
 
         if (requiresAdminAccess(request) && !hasAdminRole(claims)) {
+            logger.warn("Admin access required for ${request.requestURI}")
```
Wordbe: Same here.

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -25,7 +26,7 @@ class Member(
     @Column(name = "role", nullable = false)
     @Enumerated(EnumType.STRING)
     val role: Role = Role.USER,
-    @OneToOne(fetch = FetchType.LAZY)
+    @OneToOne(fetch = FetchType.LAZY, cascade = [CascadeType.ALL])
     @JoinColumn(name = "cart_id", nullable = true)
```
Wordbe: For the `cart` table, we could consider a unidirectional relationship that only manages `member_id`. What’s the reason for choosing a bidirectional one? 🙂

src/main/kotlin/ecommerce/model/Order.kt
```diff
@@ -0,0 +1,32 @@
...
+import jakarta.persistence.OneToOne
+import jakarta.persistence.Table
+import java.time.LocalDateTime
+
+@Entity
+@Table(name = "orders")
+class Order(
+    @OneToMany(cascade = [CascadeType.ALL], fetch = FetchType.LAZY, orphanRemoval = true)
+    @JoinColumn(name = "member_orders_id", nullable = false)
+    var orderItems: MutableList<OrderItem> = mutableListOf(),
```
Wordbe: Is there a reason the column name is `member_orders_id`? Also, it would be good to have a separate DDL file to manage the project’s data schema!

src/main/kotlin/ecommerce/model/OrderItem.kt
```diff
@@ -0,0 +1,32 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    var id: Long? = null,
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is OrderItem) return false
+        return id != null && other.id == id
+    }
+
+    override fun hashCode(): Int = id?.hashCode() ?: 0
```
Wordbe: What’s the reason for defining the `equals` and `hashCode` methods like this? If an entity doesn’t have an `id` before being persisted, it seems like all instances would be considered the same hashCode!

src/main/kotlin/ecommerce/model/Payment.kt
```diff
@@ -0,0 +1,29 @@
...
+@Entity
+class Payment(
+    @Column
+    var checkoutSessionId: String? = null,
+    @Column
+    var amount: Long? = null,
+    @Column
+    var currency: String? = "USD",
+    @Column
+    var status: String? = "PENDING",
```
Wordbe: Consider using enum 🙂

src/main/kotlin/ecommerce/service/MemberService.kt
```diff
@@ -30,9 +32,11 @@ class MemberService(
         val hashedPassword = passwordService.hashPassword(request.password)
         val member = request.toModel(hashedPassword)
 
-        val savedCart = Cart(member = member)
-        member.assignCart(savedCart)
         val savedMember = memberRepository.save(member)
+        val savedCart = cartRepository.save(Cart(member = savedMember))
+        savedMember.assignCart(savedCart)
+        memberRepository.save(savedMember)
```
Wordbe: If you’ve set up `Cascade`, try using it; if not, consider removing the `Cascade`. 🙂

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,164 @@
...
+    private val restClient = RestClient.create()
+
+    fun getMemberOrders(
+        memberId: Long,
+        page: Int,
+        size: Int,
+        sortBy: String,
+    ): Page<OrderResponse> {
+        validateMember(memberId)
+        val pageable = PageRequest.of(page, size, Sort.by(sortBy))
```
Wordbe: Providing what can be selected with `sortBy` as an enum would make it easier for the client to use and also easier for the server to validate. 🙂

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,164 @@
...
+        return orderRepository.findByMemberId(memberId, pageable).map { orderToResponse(it) }
+    }
+
+    fun getAllOrders(
+        page: Int,
+        size: Int,
+        sortBy: String,
+    ): Page<OrderResponse> {
+        val pageable = PageRequest.of(page, size, Sort.by(sortBy))
+        return orderRepository.findAll(pageable).map { orderToResponse(it) }
```
Wordbe: Same here.

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,164 @@
...
+        return productOptionRepository.findByIdOrNull(productOptionId)
+            ?: throw EntityNotFoundException("Product Option not found")
+    }
+
+    private fun calculateAmount(
+        productOption: ProductOption,
+        quantity: Int,
+    ): Long {
+        return (productOption.product.price * quantity * 100).toLong()
+    }
```
Wordbe: Shall we delegate the role to another object instead of `OrderService`?

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,164 @@
...
+
+    private fun processPayment(paymentRequest: PaymentRequest): PaymentResponse {
+        try {
+            return stripeClient.createCheckoutSession(paymentRequest)
+        } catch (e: StripePaymentException) {
+            val decline = DeclineCode.fromStripeCode(e.declineCode)
+            throw OrderProcessingException(decline.stripeCode, decline.userMessage, e)
+        }
+    }
```
Wordbe: Try separating the order service and the payment service. 🙂

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,164 @@
...
+
+@Service
+class OrderService(
+    private val orderRepository: OrderRepository,
+    private val productOptionRepository: ProductOptionRepository,
+    private val cartItemRepository: CartItemRepository,
+    private val memberRepository: MemberRepository,
+    private val stripeClient: StripeClient,
+) {
+    private val restClient = RestClient.create()
```
Wordbe: It’s not being used. Please remove this!

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,164 @@
...
+        val payment =
+            Payment(
+                checkoutSessionId = paymentResponse.id,
+                amount = paymentResponse.amount,
+                currency = paymentResponse.currency,
+                status = paymentResponse.status,
+                paymentMethod = paymentResponse.paymentMethod,
+            )
+        order.payment = payment
+        return orderToResponse(orderRepository.save(order))
```
Wordbe: Try doing it without using setters, and think about why that might be better! 🔧

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,107 @@
...
+
+        return executeStripeRequest(url = "https://api.stripe.com/v1/payment_intents", body = body)
+    }
+
+    fun confirmPayment(intentId: String): PaymentResponse {
+        return executeStripeRequest(
+            url = "https://api.stripe.com/v1/payment_intents/$intentId/confirm",
+            body = "",
+        )
+    }
```
Wordbe: This method is not being used.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
```diff
@@ -0,0 +1,107 @@
...
+            val resp =
+                restClient.post()
+                    .uri(url)
+                    .header(HttpHeaders.AUTHORIZATION, "Bearer $stripeKey")
+                    .contentType(MediaType.APPLICATION_FORM_URLENCODED)
+                    .body(body)
+                    .retrieve()
+                    .toEntity(String::class.java)
+
+            val paymentIntent = mapper.readValue(resp.body, StripePaymentIntent::class.java)
```
Wordbe: Try receiving it directly as a `StripePaymentIntent` object. Use `@JsonProperty` if necessary.

src/test/kotlin/ecommerce/controller/AdminControllerTest.kt
```diff
@@ -315,15 +306,13 @@ class AdminControllerTest {
...
-        val product =
-            objectMapper.readValue(
-                createdProductResponse.response.contentAsString,
-                Product::class.java,
-            )
+//        val product =
+//            objectMapper.readValue(
+//                createdProductResponse.response.contentAsString,
+//                Product::class.java,
+//            )
```
Wordbe: If it’s not needed, go ahead and delete it! 🗑️

src/test/kotlin/ecommerce/service/OrderServiceTest.kt
```diff
@@ -0,0 +1,78 @@
...
+                paymentMethod = "pm_visa_card",
+                currency = "USD",
+                declineCode = null,
+            )
+
+        assertThrows<EntityNotFoundException> {
+            orderService.createOrder(999L, paymentResponse, option, 1)
+        }
+    }
+}
```
Wordbe: It seems like the payment API is being called and data is being generated every time you test. Try using a stub to test without calling the API directly.

Wordbe (changes requested):
Hello Apoorva! 

Thank you for writing the code so carefully.
I also tested it using the information you sent via DM. Thanks!
I’ve written the review, so please check it and apply the parts you can.🙂

Thank you!

.dockerignore
Apoorva-S-Natesh: The corrected .dockerignore file follows a two-step process: first, ignores the entire build directory, and then uses a negation pattern ! to re-include the single JAR file that the Dockerfile needs. This approach ensures that only the required file is sent to the Docker daemon.
Thank you for mentioning about ! - I wasn't aware of this

build.gradle.kts
Apoorva-S-Natesh: No, will remove the dependency. Thank you for mentioning it. I was trying to implement timeouts with WebClient and the dependency was required. But it is unnecessary now.

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
Apoorva-S-Natesh: Thank you for the suggestion, it is a great one for improving code readability and maintainability. Refactored the method to use an enum for these general API errors -StripeApiErrors

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
Apoorva-S-Natesh: Sorry for the oversight. It is a redundant when statement that always returns HttpStatus.BAD_REQUEST. It is removed.

src/main/kotlin/ecommerce/config/ApplicationConfig.kt
Apoorva-S-Natesh: No, the `ApplicationConfig.kt` file is not being used and can be safely removed.

src/main/kotlin/ecommerce/controller/OrderController.kt
Apoorva-S-Natesh: You are right, The try-catch block in this method is redundant and can be removed. The exception handling for OrderProcessingException and other runtime exceptions is already managed by the GlobalControllerAdvice, which will handle the exceptions and return the appropriate HTTP status code and error message to the client.

src/main/kotlin/ecommerce/config/CorsConfig.kt
Apoorva-S-Natesh: This is a crucial piece of feedback, thank you . Allowing all origins with `.allowedOrigins("*")` is a significant security risk because it enables any website to make cross-origin requests to my API, I also read that this could potentially lead to cross-site request forgery (CSRF) or data exfiltration attacks. I replaced the replace the wildcard (*) with the allowed origin `http://localhost:3000`
Added a new property to application.properties file to define the allowed origins. The CorsConfig class reads the cors.allowed-origins property and uses it to configure the CorsRegistry.

src/main/kotlin/ecommerce/dto/OrderDtos.kt
Apoorva-S-Natesh: Thank you for the nudge, using enum for status is a good idea! Thought string works using enum for status makes the code more self-documenting and improves type safety.

Changes made - 
Created a new PaymentStatus Enum to define the possible payment statuses.
Modified the Payment.kt entity and the OrderResponse data class to use the new PaymentStatus enum instead of a nullable String. Since the status field will now always be one of the enum's values, it no longer needs to be nullable.
Updated the OrderService and ResponseMapper.kt to convert the string status from the Stripe API response into PaymentStatus enum.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
Apoorva-S-Natesh: Thank you for pointing that out. The idea was to create a polling system. But since its not being implement this method also is redundant. Removed it.

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: Yes this line is not used at all. Thank you for catching that. It is removed.

src/main/kotlin/ecommerce/model/OrderItem.kt
Apoorva-S-Natesh: That's a valid observation. I wanted to check if two Order items are the same (i.e if they have the same productOption for a same Order, then it should be the same orderItem.) But the implementation has a flaw due to lack of my knowledge. I understand what you mean - The hashCode method returns id?.hashCode() ?: 0. For any new OrderItem object before it is persisted (i.e., when id is null), this method will always return 0. This means that all newly created instances of OrderItem will have the same hash code. The equals method checks for equality based on id and will return false for any two new OrderItem instances because their IDs are both null. This contradicts the general contract of equals and hashCode, which states that if two objects are equal according to the equals() method, they must have the same hash code.

A better approach would be to use a business key for equality and hashing before the object is persisted, this would be the combination of its productOption and potentially the order it belongs to. But OrderItem entity doesn't have a direct reference to the Order entity; the relationship is unidirectional.

The simplest and cleanest solution that I can think of now is to remove the methods entirely. This avoids the logical pitfalls of an incorrect business key and removes code that is not functionally required.

src/main/kotlin/ecommerce/interceptor/AuthInterceptor.kt
Apoorva-S-Natesh: Thank you for pointing that out. The AuthInterceptor should be focused on its primary job: validating the token and throwing an exception if it's invalid. The responsibility of logging that a critical error occurred should be delegated to the global exception handler. This prevents duplicate log entries and centralizes the error-handling logic. I Refactored the code to delegate the logging and handling exception to `GlobalControllerAdvice` just like every other exception handling

src/main/kotlin/ecommerce/interceptor/AuthInterceptor.kt
Apoorva-S-Natesh: [Resolved here](https://github.com/herotechcourse/spring-ecommerce-order/pull/30#discussion_r2296258624)

src/main/kotlin/ecommerce/model/Member.kt
Apoorva-S-Natesh: I agree while a unidirectional relationship is simpler, choosing a bidirectional one was a deliberate decision by my teamMate and I to allow for easy and direct object-oriented navigation from the Member to the Cart without needing an additional database query. The idea was member should know about the cart and cart should know about member. For example, when an authenticated user (a Member entity) wants to add an item to their cart, our service logic can directly access the cart using member.cart. This is more intuitive than having to query the CartRepository by the memberId to find the associated cart.

However you are correct, our current implementation has an important configuration error: both `Member` and `Cart` entities define `@JoinColumn`, which can create issues with JPA's schema generation and persistence.

Correction: Making `Member.kt` the Owning Side, this will define the cart_id column and will handle cascading operations. `Cart.kt` inverse Side will use mappedBy to indicate that the relationship is managed by the member field in the Member entity.

src/main/kotlin/ecommerce/service/MemberService.kt
Apoorva-S-Natesh: Yes now CascadeType.ALL tells JPA that any persistence operations (like save, update, delete) performed on the Member entity should be automatically applied to its associated Cart entity. The JPA cascade ensures that the new Cart is saved along with its associated Member. Refactored.

src/main/kotlin/ecommerce/model/Order.kt
Apoorva-S-Natesh: The column name member_orders_id was chosen to be more descriptive however a simpler name like order_id could also be used.
The reason there is no separate DDL (Data Definition Language) file for the schema is because the application.properties file is configured to automatically manage the database schema. The property `spring.jpa.hibernate.ddl-auto=create-drop` tells the application to:
Create the database schema from the JPA entities at startup.
Drop the entire schema when the application shuts down.
I thought that this automatic schema generation is very convenient for rapid development and testing, as it saves from having to manually create or update the schema.

src/main/kotlin/ecommerce/model/Payment.kt
Apoorva-S-Natesh: Thank you for the suggestion. I understand why its better to have it as an enum instead of relying on a string that could have typos or invalid values (e.g., "usd" vs. "USD"), an enum restricts the field's value to a predefined set of valid constants.

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: Thank you for this valuable suggestion. I have refactored the code to include an OrderSortOption Enum

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: thank you, Refactored to include an [enum ](https://github.com/herotechcourse/spring-ecommerce-order/pull/30#discussion_r2296297028)

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: I agree the calculation should be a responsibility of the ProductOption. This puts the business logic closer to the data it operates on, making the code more cohesive.

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: Thank you for this suggestion, separating these concerns is a sound architectural decision. I understand how this allows for better testability and makes it easier to swap out or add new payment providers in the future without changing the core order logic. I guess in real world scenario there might be different payment providers and complex payment-related business logic. Separating this into its own service makes the system more modular and easier to maintain.

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: While the current implementation is simple and works it can make the object's state inconsistent if not handled carefully. I understand a better approach is to use a builder pattern or a constructor that creates a fully-initialized Order object in a single step with the help of entities that have immutable constructors,. This ensures that the object is always in a valid state and prevents unintended modifications.

src/test/kotlin/ecommerce/service/OrderServiceTest.kt
Apoorva-S-Natesh: This was challenging since I was stuck with NullPointerException which was happening because Mockito's any() matcher, when used alone in Kotlin, returns null by default. But my paymentService.processPayment(...) method expects a non-nullable PaymentRequest object as its argument, causing a NullPointerException when any() provides null.

src/main/kotlin/ecommerce/stripe/StripeClient.kt
Apoorva-S-Natesh: Thank you for pointing this out. This change simplifies the code, improves type safety, and is a more idiomatic way to use Spring's RestClient

changes made: changed the .toEntity(String::class.java) call to directly request a StripePaymentIntent object. This tells the RestClient to use a message converter to handle the JSON deserialization.

.dockerignore
Wordbe: Great. thank you.

src/main/kotlin/ecommerce/config/CorsConfig.kt
Wordbe: Good

src/main/kotlin/ecommerce/model/Member.kt
Wordbe: For example, think about the relationship between `order` and `order_item`. If you put `order_item_id` in `order`, the same order details would keep piling up redundantly across multiple `order_item`s. That’s why we usually put `order_id` in `order_item`. In other words, since `order_item` belongs to `order`, we make `order_item` the owner of the relationship and place the foreign key there.
Now, if we look at `member` and `cart`, it also seems like the `cart` belongs to the `member`, so it feels more natural to put `member_id` in `cart`. Could you revisit the idea of the relationship owner and explain why you thought differently from me? 🤔

src/main/kotlin/ecommerce/model/Order.kt
Wordbe: Even if you use the convenient JPA feature `spring.jpa.hibernate.ddl-auto`, it’s still advisable to manage DDL files that can preserve the current DB metadata. This will make maintenance smoother. How about giving it a try? 🙂

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,143 @@
...
+        productOption: ProductOption,
+        quantity: Int,
+    ): OrderResponse {
+        validateMember(memberId)
+        val orderItem =
+            OrderItem(
+                quantity = quantity,
+                price = productOption.product.price,
+                productOption = productOption,
+            )
```
Wordbe: There’s a critical issue where `orderItem` isn’t being used. ⚠️

src/main/kotlin/ecommerce/model/OrderItem.kt
Wordbe: Check out the material below, and if you’re convinced, try generating the `equals()` and `hashCode()` methods using the JPA Buddy plugin. 
https://jpa-buddy.com/blog/hopefully-the-final-article-about-equals-and-hashcode-for-jpa-entities-with-db-generated-ids/

src/test/kotlin/ecommerce/service/OrderServiceTest.kt
Wordbe: Thank you for writing the test code so well. 👍

Wordbe (changes requested):
Hello Apoorva!
There were a lot of reviews, but thank you for thinking them through, sharing your thoughts, and reflecting them well.
Most of it was done very well, but I left another comment regarding the JPA relationships that might be worth revisiting.
Please take a look. Thank you! 🙏

src/main/kotlin/ecommerce/service/OrderService.kt
Apoorva-S-Natesh: You are right, this is a logical bug. The orderItem object is created but never added to the order object's list of items. 
Change made:
added `order.orderItems.add(orderItem)` -  adds the OrderItem to the Order's collection before the order is saved to the repository. This ensures the order items are persisted correctly along with the order itself.

src/main/kotlin/ecommerce/model/Member.kt
Apoorva-S-Natesh: Thank you for patiently explaining this relationship concept. I was a bit confused but I understand now that it's a good JPA practice for the child or dependent entity to own the relationship. In my case, a Cart belongs to a Member, so Cart should hold the foreign key (member_id), making it the owning side. This is consistent with the Order and OrderItem relationship. I have refactored to follow this.

src/main/kotlin/ecommerce/model/Order.kt
Apoorva-S-Natesh: Thank you for this suggestion again. I read more about the DDL file and If my understanding is correct a single DDL file for schema management is a more robust practice than relying solely on `spring.jpa.hibernate.ddl-auto=create-drop or auto`. While convenient for development, a dedicated DDL file ensures schema stability and allows for controlled migrations in production environments. I have tried to create a `schema.sql`. Please let me know if its correct. the order in which the tables were dropped at the beginning was tricky for me because of the bidirectional relationships and foreign keys, which makes me wonder if there are better ways to manage entities without using bidirectional relationships in a real life project. Please let me know your thoughts on this.

src/main/kotlin/ecommerce/model/OrderItem.kt
Apoorva-S-Natesh: Thank you for providing the resource, it helped me understand the equals() and hashcode(). Also the tip for using `JPA Buddy plguin` is pretty amazing! thank you 🙂. I learnt that javaClass is a special Kotlin property that gives the runtime class of the object. In the generated code, `if (javaClass != other?.javaClass) return false` is a crucial first check that ensures the two objects being compared are of the exact same type.
Also  in`result = 31 * result + order.hashCode()`, I was wondering why 31 which felt random but the number 31 is a special value used in hashCode() implementations for several reasons. It's an odd prime number, which is good for creating a more uniform distribution of hash codes.

Apoorva-S-Natesh:
Thank you for explaining the relationship between entities and for helping me better the code logic and follow standard JPA practices. Also grateful for all the tips and for the review. 😁 🙏🏼

src/main/kotlin/ecommerce/model/Order.kt
Wordbe: Yes, I think you wrote it well. That’s why it’s best to use bidirectional relationships and foreign keys carefully. Apply them only when necessary. 👍

Wordbe (approved):
Thank you for giving your best effort throughout the missions.
You’ve carefully considered and applied JPA relationships, and you’ve also shared your thoughts well and improved other areas.
It feels like we’ve reached the end of this mission journey. You’ve worked really hard.
I hope the process you went through here. Writing code, reflecting, and thinking will help you in the future when you create or review code.
Wishing you the best of luck ahead. Thank you. 😀

PR #31, opened 2025-08-19
[Spring-Ecommerce-Order Step 2] Submit Petra's Mission

pebencze:
# Spring E-Commerce Order Step 2

Implement a new endpoint `/api/orders` and Stripe Payment integration. Deploy the application in an EC2 instance on AWS.

## Description

**1. Order Placement & Retrieval**
- User selects:
    - Product Option
    - Quantity
    - Payment method (Stripe test card in sandbox)
- API calculates total price and sends request to Stripe's Payment Intent Create API
- Possibility for `LoginMembers` to view all placed orders at: `/api/orders`
    -  contains useful information about the orders: e.g. status, date

**2. Stripe**
- Use `RestClient` for Stripe Payment Integration
   - POST https://api.stripe.com/v1/payment_intents
   - `confirm=true`
-  Stock management and Cart cleanup with two scenarios
   - Payment fails: nothing changes
   - Payment succeeds: remove items, decrease stock
- Error management:
    - 4xx throws 400 BadRequestException() 
    - 5xx throws 503 ExternalServiceException()
    - catch both types of exceptions and send back a message to the client with: message, status, orderId?

**3. Deployment**
- Simple deployment script to automate the deployment process (`deploy.sh`)
     - manual build 
     - script looks for the JAR file, kills existing process if the application is running and starts the application
- Cors Configuration
  - Allow all origins.
  - Allow “simple” methods GET, HEAD and POST.
  - Allow all headers.
  - [Set max age to 1800 seconds (30 minutes).

I am attaching some screenshots from testing the deployment.
Screenshot 2025-08-19 at 09 45 34
Screenshot 2025-08-19 at 09 50 46
Screenshot 2025-08-19 at 09 58 13

## Questions for the Reviewer

Q1: I have tried implementing separate transactions for order placement and payment processing, using 2 Services with PropagationType. REQUIRES_NEW, but I did not succeed. The goal was to be able to save Order even if the payment fails and I throw an exception. I was trying to pass the Order entity from one transaction to the other and that did not work. Now, I am catching exceptions and just returning a message to the user, so I am able to persist the failed Order. I was wondering if there is a better solution to handle stripe errors, e.g. by creating separate Transactions and if yes, how. related to that: is it wiser to throw an exception or to return an error message (and maybe log the error)?

Q2: When and why are data classes useful for DTOs?


README.md
```diff
@@ -83,12 +83,150 @@ Design and implement the feature considering the relationship between the Produc
...
 - [x] decide on where and how to use Models
     - Entity == Model
-- [ ] dont use Cascade.All but Persist, Merge etc.
+- [x] don't use Cascade.All but Persist, Merge etc.
 - [ ] move product-option mapping to the constructor
-- [ ] effective test: Create InMemory fake repos -> service uses the fake repos
\ No newline at end of file
+- [ ] effective test: Create InMemory fake repos -> service uses the fake repos
+- [ ] add HTTPS
+- [ ] add webhook endpoint (/webhooks/stripe)
```
pkeugine: I see there are still some [left over considerations](https://github.com/herotechcourse/spring-ecommerce-order/pull/14#discussion_r2257163060) and even new ones as well! 😮 

Since we're now past the time of this mission, I think it would be great to remove these considerations to make it not feel like it's not finished. You've done a great job with this project already 👍 

(I found another unchecked box under `Orders` and `Features`)

src/test/kotlin/ecommerce/controller/StatisticsE2ETest.kt
```diff
@@ -39,6 +43,9 @@ class StatisticsE2ETest {
     @Autowired
     private lateinit var cartItemRepository: CartItemRepository
 
+    @Autowired
+    private lateinit var optionRepository: OptionRepository
+
```
pkeugine: This might not seem like a big deal, but there are two whitespace characters in this file, in line 48.
Due to these whitespace characters, lint checking fails when I try to clean-build the project.
Why don't we remove these? :)

build.gradle.kts
```diff
@@ -37,6 +37,8 @@ dependencies {
     runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.12.6")
     testImplementation("org.mockito.kotlin:mockito-kotlin:5.2.1")
     implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+    implementation("com.stripe:stripe-java:29.4.0")
```
pkeugine: Wow, I didn't expect you'd find this Stripe library 😮 👍  
Well done!

However, I'm not sure if we're using it.
If we're not, I think it would be better for us to remove it.

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,57 @@
...
+                "payment_method=${req.paymentMethod}",
+                "confirm=true",
+                "automatic_payment_methods[enabled]=true",
+                "automatic_payment_methods[allow_redirects]=never",
+            ).joinToString("&")
+
+        return try {
+            val response =
+                restClient.post()
+                    .uri("https://api.stripe.com/v1/payment_intents")
```
pkeugine: I recommend managing 3rd party api URL in a different file such as in properties file (`application.properties` could be an option), and inject the value in application logic.

This is because by separating the URL from application logic, we can easily handle the changes when the URL changes, and we can also mock the API when testing.

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
```diff
@@ -27,4 +30,16 @@ class WebMvcConfiguration(
             .addPathPatterns("/api/**")
             .excludePathPatterns("/api/members/login", "/api/members/register")
     }
+
+    override fun addCorsMappings(registry: CorsRegistry) {
+        registry.addMapping("/api/**")
+            .allowedOrigins("*")
```
pkeugine, line 36: This is just a TMI that I think would be great for us to cover, rather than actually changing the logic.

For now, `allowedOrigins("*")` works fine, but if we're making APIs that are only for our team or product, we need to narrow down the allowed origins to expected client information. If we're making public API, then the situation is different, but in most of the times, we make APIs that are used in only certain places. We'll get to that situation at some point I guess (it might even be in team mission 🙂 )

src/main/kotlin/ecommerce/dto/PaymentIntentResponse.kt
```diff
@@ -0,0 +1,13 @@
+package ecommerce.dto
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
+
+data class PaymentIntentResponse(val id: String)
```
pkeugine, line 6: **Q. When and why are data classes useful for DTOs?**

If we look at the documentation for [Kotlin's data class](https://kotlinlang.org/docs/data-classes.html), we'll be able to find the clue to this question.

It gives us features without boilerplate codes, and help us hold certain data values together while allowing us to compare them with ease (see how the `equals()` method works for data class).

After checking out what data classes do, I think decision making process becomes simple.
If we need those features for our DTOs, we consider them useful and make DTOs with data class.
If not, it's not necessary.

I see you've made most of the DTO classes without data class.
Do you think changing them to data class could help us manage this project? 😄

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,57 @@
...
+//            val sessionId = createPaymentIntent(PaymentRequest(order.paymentAmount, order.currency, order.paymentMethod))
+//            order.status = OrderStatus.PAID
+//            order.checkoutSessionId = sessionId
+//            orderRepository.save(order)
+//        } catch (e: Exception) {
+//            order.status = OrderStatus.FAILED
+//            orderRepository.save(order)
+//            throw e
+//        }
+//    }
```
pkeugine: Let's remove commented out code in PR.

src/test/kotlin/ecommerce/ApplicationTests.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.test.context.ActiveProfiles
+
+@SpringBootTest
+@ActiveProfiles("test")
+class ApplicationTests {
+    @Test
+    fun contextLoads() {
+    }
+}
```
pkeugine: I don't think this test is necessary in our current state of this project.
Why don't we remove it?

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,57 @@
...
+import ecommerce.model.Order
+import ecommerce.model.OrderStatus
+import ecommerce.repository.OrderRepository
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+
+@Service
+class PaymentService(
+    private val stripeClient: StripeClient,
+    private val orderRepository: OrderRepository,
```
pkeugine: Just like commented out code,
I think it would be nice to remove unused injection as well.

It would be great to check other classes too!

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            val response = stripeClient.createCheckoutSession(req)
+            return response
+        } catch (e: BadRequestException) {
+            throw e
+        } catch (e: ExternalServiceException) {
+            throw e
+        } catch (e: Exception) {
+            throw RuntimeException(e)
+        }
```
pkeugine: If we're going to throw exceptions right away without wrapping it to another exception or handling them, I'm not sure if we'll need to catch them here.

Same goes to `Exception` handling part. If we're going to wrap it, why don't we use a custom exception so we can manage it in a more expected manner? This second part is just a suggestion so I'm open to other options :)

src/main/kotlin/ecommerce/service/PaymentService.kt
```diff
@@ -0,0 +1,57 @@
...
+    @Transactional
+    fun processPayment(order: Order) {
+        try {
+            val request = PaymentRequest(order.paymentAmount, order.currency, order.paymentMethod)
+            order.checkoutSessionId = createPaymentIntent(request).id
+            order.setStatus(OrderStatus.PAID, "Payment successful. Order has been placed")
+        } catch (e: Exception) {
+            order.setStatus(OrderStatus.FAILED, "Payment failed. Reason: ${e.message}")
+        }
+    }
```
pkeugine, line 29: **Q. I was wondering if there is a better solution to handle stripe errors**

I think it's good like this for now 👍 
Usually, when trying to manage payment, keeping the order status as `pending`, and managing it asynchronously could be an option. However, I think that kind of implementation is far from what we're targeting in this mission, so let's just go over it as `ah, so there are some other way` :)

README.md
```diff
@@ -1,11 +1,10 @@
-# spring-ecommerce-order
+# Spring Ecommerce Order
```
pkeugine, line 1: It would've been great if I took a look on your deployment script!
Did you deploy the project manually, or did you use a script?
Since it's the first time we've actually deployed something, I'm curious about your experience related to it. 😄

pkeugine (changes requested):
Hello Petra, and nice work continuing on with this mission.
You must've been busy, but still managed to work on this! 👍 

I left some simple comments, and added some questions as well.
Please check on them when you have the time.

I think we'll be able to merge this PR next time, so let's give it a look one last time.

I hope you're doing fine with team project, and as always,
Happy Hacking! 🔥 🔥 🔥

src/test/kotlin/ecommerce/controller/StatisticsE2ETest.kt
pebencze: Oh, yes! I am sorry about those. They must have appeared during the resolution of merge conflicts.

build.gradle.kts
pebencze: Fair. I was using it initially and it was very convenient, but then I decided to take another path to be able to test and mock exceptions better.

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
pebencze: Thank you for clarifying this, PK!

src/main/kotlin/ecommerce/dto/PaymentIntentResponse.kt
pebencze: I have read many times about data classes before the project and wanted to gain hands-on experience, but I could not find the right places to use them. Comparing and validating data would be much easier, I think. Logging as well. I will try to use them more in further projects or find projects that use them.

src/main/kotlin/ecommerce/service/PaymentService.kt
pebencze: Thanks! I was just curious if the idea of using @Transactional(propagation = Propagation.REQUIRES_NEW, noRollbackFor = [Exception::class]) was worth it to further develop or not. Have you ever used "propagation = Propagation.REQUIRES_NEW"?

src/main/kotlin/ecommerce/service/PaymentService.kt
pebencze: You are right. It is leftover code from my initial logic. I could have paid more attention to this file. I have resolved this and now I am only wrapping the exceptions inside` StripeClient.kt` in the following manner:
```kotlin
 throw when {
                statusCode.is4xxClientError -> BadRequestException("An error occurred during payment: ${stripeError.error.message}")
                statusCode.is5xxServerError -> ExternalServiceException("Stripe down: ${stripeError.error.message}")
                else -> RuntimeException("Unexpected error: ${e.message}")
            }
```

I like to differentiate here between Client (4xx), External Service (503 or 5xx) and other unexpected RuntimeExceptions.

README.md
pebencze: Screenshot 2025-08-26 at 15 22 56
My deployment script looks like this, pretty simple. I have built and moved the JAR file manually. 😄 In our team project we are discovering automated build and deployment with pipelines and that is pretty cool!

src/main/kotlin/ecommerce/service/PaymentService.kt
pkeugine: I've used it to try it out, but I haven't used it in real production logic yet.
There were always other options I could choose, and I (+ my team) thought the other option was better at that time.
However, if I had to use it, I definitely would :)

README.md
pkeugine: I recommend going over what `nohup`, `2>&1`, and `&` does in the last line, since those are the main parts of the deployment. They are used quite often, and are also important questions in tech interviews :)

pkeugine (approved):
Hello Petra, and nice work refactoring the project 👍 
As I said before, I think we are good to go and are ready to merge this PR.

I hope you are doing great on your team project,
and the knowledge that you've gained through these missions help you along the way.

Let's keep this momentum going, and as always,
Happy Hacking! 😄 😄 😄

PR #32, opened 2025-08-20
[Spring-Ecommerce-Order Step 2] Submit knopgm's Mission

knopgm:
I'm sorry for the delay of this PR Mission! I really appreciate your time reviewing my code!🙏 

# In this step, I have implemented the following:

## Step 2-1: Place Order Feature
- Integrated Stripe's payment system.
- Built the functionality for users to finalize their order, including their chosen product options and quantities.
- The system now automatically reduces the product stock after a successful purchase.
- If the ordered product was in the user's cart, it gets removed automatically.
- Added safeguards to handle instances where a payment might be approved but the subsequent process fails.
- Handle error messages are now shown for payment issues.

## Step 2-2: Orders Feature
- Implemented Orders API to retrieve detailed order information, including:
- Order date/time
- Order status
- Purchased items
- Stripe checkout session ID
- Payment amount

## Step 2-3: Deployment
- Deployed the order service to AWS.
- Created a deployment script to automate build, upload, and restart.


src/main/kotlin/ecommerce/Application.kt
```diff
@@ -1,8 +1,11 @@
 package ecommerce
 
+import ecommerce.config.StripeProperties
 import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.boot.runApplication
 
+@EnableConfigurationProperties(StripeProperties::class)
```
da-nyee, line 8: Is this necessary here? 👀

src/main/kotlin/ecommerce/service/payment/StripeClientService.kt
```diff
@@ -0,0 +1,38 @@
+package ecommerce.service.payment
+
+import com.stripe.model.PaymentIntent
+import com.stripe.param.PaymentIntentCreateParams
+import org.springframework.stereotype.Service
+
+@Service
+class StripeClientService {
```
da-nyee:
```suggestion
@Component
class StripeClient {
```

Well, I think it comes down to personal preference.
To me, it seems more like a component than a service, since it doesn't have business logic but only handles API calls 🤔

src/main/kotlin/ecommerce/service/payment/StripeClientService.kt
```diff
@@ -0,0 +1,38 @@
...
+                .setPaymentMethod(paymentMethod)
+                .setAutomaticPaymentMethods(
+                    PaymentIntentCreateParams.AutomaticPaymentMethods.builder()
+                        .setEnabled(true)
+                        .setAllowRedirects(PaymentIntentCreateParams.AutomaticPaymentMethods.AllowRedirects.NEVER)
+                        .build(),
+                )
+                .build()
+
+        return PaymentIntent.create(params)
```
da-nyee: What about using `RestClient` rather than the library?

da-nyee: I'd recommend setting timeouts for the client and throwing exceptions according to the Stripe docs 👀

src/test/kotlin/ecommerce/service/OrderServiceTest.kt
```diff
@@ -0,0 +1,370 @@
...
+    lateinit var paymentRepository: PaymentRepository
+
+    @Autowired
+    lateinit var stripeClientService: StripeClientService // injected mock
+
+    @TestConfiguration
+    class TestConfig {
+        @Bean
+        fun stripeClientService(): StripeClientService = mock(StripeClientService::class.java)
+    }
```
da-nyee, line 73: 👍

src/test/kotlin/ecommerce/service/OrderServiceTest.kt
```diff
@@ -0,0 +1,370 @@
...
+            memberRepository.save(
+                MemberEntity(
+                    email = "john.doe@john.com",
+                    password = "123456",
+                    role = "user",
+                    name = "John Doe",
+                ),
+            )
+        val cart = cartRepository.save(CartEntity(memberId = member.id!!))
+//        member.cart = cart
```
da-nyee: What about removing this code if it's not used?

scripts/deploy.sh
```diff
@@ -0,0 +1,40 @@
...
+nohup java -jar "$APP_JAR_LINK" > "$APP_DIR/app.out" 2>&1 &
+
+echo "Waiting 20 seconds..."
+sleep 20
+
+curl -sS http://localhost:8080/actuator/health || true
+EOF
+
+echo "Deployed: $JAR"
+
```
da-nyee, line 40: Deployment script 👍

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,171 @@
...
+
+            if (payment == null || payment.stripePaymentIntentId == null) {
+                throw NoSuchElementException("Payment not found")
+            }
+
+            if (payment.status == OrderAndPaymentStatus.PAID) {
+                throw PaymentFailedException("Payment already made, status = ${payment.status}")
+            }
+
+            stripeClientService.confirmPaymentIntent(payment.stripePaymentIntentId)
```
da-nyee: Is it okay to place an external API call in the middle of a transaction?
If not, what potential problems might it cause? 🤔

da-nyee (changes requested):
Hi Gabi! 🙌

I know things have been hectic with the project, but you did a great job 👍 

I've gone through the code and left a few comments, so please take a look when you have time.

If anything's unclear, feel free to reach out — I'll be happy to help ☺️

src/main/kotlin/ecommerce/Application.kt
knopgm: Since I have a StripeProperties file that is configuring Stripe properties, it isn't necessary, indeed. xD
Thank you for catching this!

src/main/kotlin/ecommerce/service/payment/StripeClientService.kt
knopgm: Good point! I'll change it to keep the code clear. :)

src/main/kotlin/ecommerce/service/OrderService.kt
knopgm: Interesting! After your review, I discover in my researches that calling an external API inside the transaction can cause several potential problems as long transaction durations, database connection exhaustion for waiting too long, inconsistence in the state because stripe could be succeeded but if database commit fail is a risk that the order will not be marked as paid.
I removed the external API call from the transaction and break the transactional logic into specific responsibility concerns. 
Thanks for question me with this important topic! 🙏

src/main/kotlin/ecommerce/service/payment/StripeClientService.kt
knopgm: Thanks for another good point of revision in my code! :)
I replaced Stripe Java SDK with Spring's RestClient to handle all Stripe API communications, better consistency with Spring ecosystem, added connect timeout of 10 seconds and read timeout of 30 seconds, implemented custom StripeApiException with proper HTTP status code.

Let me know if I need more refactoring here.

src/main/kotlin/ecommerce/Application.kt
knopgm: After replacing Stripe library with the RestClient solution I created a StripeProperties data class that is annotated with a @configurationProperties bean.
Because of that I had to added the @EnableConfigurationProperties bean back to the Application file to bind my application files values to these classes and use it as beans in the Spring context. 

I hope now it is ok to have it there. 🙏

knopgm:
Hi Danye :D
I apply all the reviews you suggested.
Thank for going through my code again and raising all these questions that, while coding in a rush, I may have skipped important points and concepts.
I'll make sure to revisit and remade this project as a way to study knowledges that I couldn't dive myself into now.

I really appreciate your kindness and time! 🙏 😄

src/main/kotlin/ecommerce/service/payment/StripeRestClient.kt
```diff
@@ -0,0 +1,81 @@
...
+class StripeRestClient(
+    private val stripeProperties: StripeProperties,
+) {
+    private val restClient: RestClient
+
+    init {
+        val requestFactory =
+            SimpleClientHttpRequestFactory().apply {
+                setConnectTimeout(10000) // 10 seconds
+                setReadTimeout(30000) // 30 seconds
```
da-nyee, line 26: 👍

da-nyee: Is there any particular reason you set the timeouts to 10s and 30s respectively?

src/main/kotlin/ecommerce/Application.kt
da-nyee: Okay, I got it 👍

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,217 @@
+package ecommerce.service
+
+import ecommerce.dto.CreateOrderResponse
+import ecommerce.dto.MemberResponse
+import ecommerce.entity.Order
+import ecommerce.entity.OrderItem
+import ecommerce.entity.Payment
+import ecommerce.enums.OrderAndPaymentStatus
+import ecommerce.handler.OrderCreationException
+import ecommerce.handler.PaymentFailedException
```
da-nyee: You handled the exceptions well, but what about handling them within the client?
I think that would feel more natural.

da-nyee: I'd also recommend handling exceptions in more detail according to the [docs](https://docs.stripe.com/error-handling?lang=java#payment-errors) 👀

src/main/kotlin/ecommerce/service/OrderService.kt
da-nyee: You nailed it 💯

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,217 @@
...
+    private val orderItemRepository: OrderItemRepository,
+    private val paymentRepository: PaymentRepository,
+    private val memberRepository: MemberRepositoryJpa,
+    private val stripeClientService: StripeClientService,
+    private val optionService: OptionService,
+    private val applicationContext: ApplicationContext,
+) {
+    // Get proxy instance for transactional methods
+    private val self: OrderService
+        get() = applicationContext.getBean(OrderService::class.java)
```
da-nyee: I think you added this code because of AOP (specifically `@Transactional`), since it doesn't support self-invocation.
That's fine, but what about separating the transaction-handling code into another class?
Injecting `ApplicationContext` for this feels a bit awkward 🤔

da-nyee (changes requested):
Hi Gabi! 🙌

Thanks for applying the feedback — I can see your growth 👍 

However, there are still a few things to improve, so I left some additional comments.
Please check them out when you get a chance.

It's almost there — keep it up 💪

src/main/kotlin/ecommerce/service/payment/StripeRestClient.kt
knopgm: Hey Danye!
In my research, the `setConnectTimeout` is the time to wait for a TCP connection to be established with the remote server and a slow connection handshake almost always indicates a fundamental network problem, which is pointless to wait more for this. 
A shorter connect timeout allows your application to fail fast. If Stripe's API is unreachable, you'll know within 10 seconds instead of holding up application threads and user requests for longer.
It's also the recommended Industry Standard range for connect timeouts for external APIs.

For the `setReadTiemout`, I learned that this is the time for Stripe's servers to receive your request, process it, and start sending the response back. Setting this too low (e.g., 5 seconds) would cause handling redirects for secure 2 authentication longer-running processes to fail with a timeout error, leading to a poor user experience and failed payments.
Nevertheless, 30 seconds prevents a hung request from consuming a thread in your application indefinitely, which could lead to resource exhaustion. Stripe's client libraries and documentation often suggest or use longer read timeouts for this exact reason.

Is that a good decision? I didn't found any other arguments against it, but I can change it at any time. :D

src/main/kotlin/ecommerce/service/OrderService.kt
knopgm: That was another good lesson of Stripe documentation! 👀 
I added another handle exception file just for StripeApiException, refactor StripeRestClient, StripeClientService and also cleaned OrderService from dealing with stripe error catches.

Thanks for linking the docs! 🙏 😄

src/main/kotlin/ecommerce/service/OrderService.kt
knopgm: Thank you for this review! 🙏 
I added the separation of concerns as you suggested.
Now OrderTransactionService has only transactional commands (can I say that this is state changes? 🤔 ), and OrderService is managing business orchestration (coordinates commands + external services).

knopgm:
Thank you for those catches Danye! 
I really appreciate your time and the knowledge you shared on your reviews! 😄 

I'm asking for another review of the review implementation now.

src/main/kotlin/ecommerce/service/payment/StripeRestClient.kt
```diff
@@ -0,0 +1,110 @@
...
+                null,
+                e,
+            )
+        } catch (e: RestClientException) {
+            // This is a lower-level exception (e.g., connection timeout, read timeout, network issue)
+            throw StripeConnectionException(
+                "Failed to communicate with Stripe: ${e.message}",
+                e,
+            )
+        }
```
da-nyee, line 108: Great improvement on the error handling 👍

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,123 @@
...
+            stripeClientService.confirmPaymentIntent(stripePaymentIntentId)
+            orderTransactionService.completeSuccessfulPayment(stripePaymentIntentId)
+            "Payment successful, order $orderId marked as PAID"
+        } catch (e: StripePaymentFailedException) {
+            orderTransactionService.markPaymentAsFailed(stripePaymentIntentId, "Payment failed: ${e.stripeErrorCode ?: "Unknown code"}")
+            throw PaymentFailedException("Payment was declined: ${e.message ?: "Unknown error"}")
+        } catch (e: StripeConnectionException) {
+            orderTransactionService.markPaymentAsFailed(stripePaymentIntentId, "Network error: ${e.message}")
+            throw PaymentFailedException("Temporary payment issue. Please try again.")
+        }
```
da-nyee, line 121: Well done 👍

src/main/kotlin/ecommerce/service/OrderService.kt
```diff
@@ -0,0 +1,123 @@
...
+            ),
+        )
+
+        val paymentIntentId =
+            try {
+                stripeClientService.createPaymentIntent(
+                    amountInCents,
+                    currency,
+                    paymentMethod,
+                )
```
da-nyee, line 81: Looks like you might have missed separating this external API call from the transaction 👀

src/main/kotlin/ecommerce/service/payment/StripeRestClient.kt
da-nyee: Perfect 💯

src/main/kotlin/ecommerce/service/OrderTransactionService.kt
```diff
@@ -0,0 +1,98 @@
...
+import ecommerce.enums.OrderAndPaymentStatus
+import ecommerce.handler.PaymentFailedException
+import ecommerce.repository.CartRepositoryJpa
+import ecommerce.repository.OrderRepository
+import ecommerce.repository.PaymentRepository
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+
+@Service
+class OrderTransactionService(
```
da-nyee, line 13: 👍

da-nyee (approved):
Hi Gabi! 🙌

Thanks for thoughtfully addressing all the feedback 🙇‍♀️ 
Everything looks solid — finally time to wrap up the last mission!
I'm sure you'll keep doing great, because you've already made it through such tough times.

Wishing you all the best — fingers crossed 🤞🏻🍀

PR #34, opened 2025-08-25
[spring-ecommerce-order - Step-2] Submit miooo0o's mission

miooo0o:
Hi, Dani. I'm very, very late :(
I rushed through the development and couldn't pay attention to documenting the process, so I failed to write a proper TODO list.

Screenshot 2025-08-25 at 16 20 40

### I encountered several major issues initially:
1. During development, I discovered problems with `CartItem` not properly connecting to `Option`/`Product`.
2. Tables weren't being created properly, which consumed a lot of time.

Since many problems were caused by bi-directional relationships, I came to prefer uni-directional entity relations after resolving these issues.`Payment` and `Order` were designed as uni-directional based on my experience implementing the entire flow in a team project. Since I'm handling payment-order flows in the team project, I tried to create a structure that's as concise yet practical as possible for this mission.

Regarding Lock usage, I temporarily applied it based on my concurrency control experience from the team project, but reverted it because it made testing complex. I wanted to keep the mission as simple as possible.

### 1. Changed CartItem to use Option instead of Product
```kotlin
@ManyToOne(fetch = FetchType.LAZY)
var option: Option
```

### 2. Added Stripe payment system
- Execute `CheckoutController` endpoints through `StripeProperties`/`StripeClient`
- Designed to simply receive and reflect results from `Stripe` (due to time constraints)

### 3. Changed to use `.env` file
```properties
security.jwt.token.secret-key=${TOKEN_SECRET_KEY}
stripe.secret-key=${STRIPE_SECRET_KEY}
```
I will send you this .env file via Slack

### 4. Exception management using GlobalException
```kotlin
    @ExceptionHandler(
        value = [
            ProcessingException::class,
            RuntimeException::class,
            IllegalArgumentException::class,
        ],
    )
    fun handleBadRequest(e: Exception): ResponseEntity<ErrorResponse> {
        return buildErrorResponse(HttpStatus.BAD_REQUEST, e.message, e.extractErrors())
    }
```


src/main/kotlin/ecommerce/Application.kt
```diff
@@ -1,8 +1,11 @@
 package ecommerce
 
+import ecommerce.client.StripeProperties
 import org.springframework.boot.autoconfigure.SpringBootApplication
+import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.boot.runApplication
 
+@EnableConfigurationProperties(StripeProperties::class)
```
da-nyee, line 8: I think it would make more sense to annotate `StripeClient` instead 🤔

src/main/kotlin/ecommerce/client/StripeProperties.kt
```diff
@@ -0,0 +1,10 @@
+package ecommerce.client
+
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+@ConfigurationProperties("stripe")
+data class StripeProperties(
+    val secretKey: String,
+    // TODO: .env setting
```
da-nyee: Thanks for sharing the `.env` values 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,54 @@
...
+import ecommerce.dto.PaymentIntent
+import ecommerce.exception.StripePaymentException
+import org.springframework.http.HttpHeaders
+import org.springframework.http.MediaType
+import org.springframework.stereotype.Component
+import org.springframework.web.client.RestClient
+
+@Component
+class StripeClient(private val stripeProperties: StripeProperties) {
+    private val restClient = RestClient.create()
```
da-nyee: What about setting timeouts for the client?

da-nyee: Have you heard of `WebClient`? If so, how is it different from this one?

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,54 @@
...
+            val paymentIntent = requireNotNull(response.body)
+
+            validatePaymentIntent(request, paymentIntent)
+            paymentIntent
+        } catch (e: Exception) {
+            throw StripePaymentException("Stripe error: ${e.message}")
+        }
+    }
+
+    fun validatePaymentIntent(
```
da-nyee: You might consider making it `private` 👀

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,54 @@
...
+                    .retrieve()
+                    .toEntity(PaymentIntent::class.java)
+
+            val paymentIntent = requireNotNull(response.body)
+
+            validatePaymentIntent(request, paymentIntent)
+            paymentIntent
+        } catch (e: Exception) {
+            throw StripePaymentException("Stripe error: ${e.message}")
+        }
```
da-nyee: What about handling exceptions in more detail, following the [docs](https://docs.stripe.com/error-handling?lang=java#payment-errors)?

src/main/kotlin/ecommerce/service/CheckoutService.kt
```diff
@@ -0,0 +1,58 @@
...
+    private val orderService: OrderService,
+    private val paymentRepository: PaymentRepository,
+) {
+    @Transactional
+    fun checkout(
+        member: RegisteredMember,
+        request: CheckoutRequest,
+    ): CheckoutResponse {
+        val order = processCheckout(member, request)
+        val payment = executePayment(request)
```
da-nyee, line 27: Is it okay to have an external API call within a transaction? If not, what potential problems could occur?

README.md
```diff
@@ -1,253 +1,21 @@
...
-- [x] This logic should be implemented in the Service class or Entity class for future reuse.
-- [x] check inside DTO or addOption()
-  - [x] A product must always have at least one option.
-  - [x] Duplicate option names are not allowed within the same product to prevent confusion during purchase.
-
-#### Test
-- [x] test Option
\ No newline at end of file
+### Step 2
+#### todo
```
da-nyee: Could you add a shell script for deployment? 👀

da-nyee (changes requested):
Hi Mina! 🙌

It's totally fine to request a late review, and you did a great job 👍 
But there are things to improve, so I left a few comments — please take a look once you have a moment.

If anything's unclear, feel free to let me know 🙇‍♀️

src/main/kotlin/ecommerce/Application.kt
miooo0o: Hello. I don't quite understand this comment. 
in my understand that `@EnableConfigurationProperties` makes properties into Spring beans and automatically fills in configuration values, but if I put Client in there, what would it fetch and fill?

src/main/kotlin/ecommerce/client/StripeClient.kt
miooo0o:
```kotlin
    private val restClient = RestClient.builder()
        .requestFactory(SimpleClientHttpRequestFactory().apply {
            setConnectTimeout(Duration.ofMillis(500))
            setReadTimeout(Duration.ofSeconds(10))
        })
        .build()
```

I tried applying it like this!
Regarding WebClient, no, I heard about it for the first time. I looked it up and found out it's asynchronous and non-blocking! 
...
WebClient was introduced in Spring 5 and RestClient in Spring 6.1. After looking into the pros and cons of both, I think RestClient would have been more suitable for my project.

src/main/kotlin/ecommerce/service/CheckoutService.kt
miooo0o: Hello! Thank you for the great question.

If an external API is called within a transaction, that part is outside of my control, so even if other parts are rolled back, the external API's data won't be rolled back. I didn't think about this aspect. My team project probably has the same issue (haha).

I think the solution would be to break down the current simplified process into more detailed steps, handling transactions in smaller units so that parts using external APIs are processed outside of transactions, and using event-driven asynchronous processing... but the problem would still persist. 

So combining everything, breaking transactions into smaller units while processing asynchronously to improve response speed and redesigning retry/compensation mechanisms! would be the one of approaches? 

I'm curious about your opinion...

src/main/kotlin/ecommerce/client/StripeClient.kt
miooo0o: I applied! 

```kotlin
sealed class StripeException(message: String, override val cause: Throwable? = null) : RuntimeException(message) {
    class Server(info: StripeErrorInfo, cause: Throwable? = null) :
        StripeException("Stripe server error: ${info.message} (code: ${info.code})", cause)

    class Client(info: StripeErrorInfo, cause: Throwable? = null) :
        StripeException("Stripe client error: ${info.message} (code: ${info.code})", cause)

    class Other(override val message: String, override val cause: Throwable? = null) : StripeException(message, cause)
}
```

I'm not sure if this turned out well. Actually, I think it would have been better if I had structured it a bit more carefully.

README.md
miooo0o: I updated my readme! Thank you 👍

miooo0o:
Hi, Dani. Thank you for understanding the late submission.

I tried to apply the review you mentioned as much as possible. However, I couldn't understand the first comment well, so I wasn't able to apply it (to be more precise, it's closer to saying I tried but failed!)

I really want to hear your answer about [one of the comments](https://github.com/herotechcourse/spring-ecommerce-order/pull/34#discussion_r2301380614) you gave this time. it's a part where I could have done better in the team project... Since the project is already wrapping up, I probably won't be able to change it, but I'm still really curious.

Thank you once again for understanding the late submission.

src/main/kotlin/ecommerce/client/StripeClient.kt
miooo0o: Thank you! Applied!

README.md
```diff
@@ -1,253 +1,68 @@
 # spring-ecommerce-order
+## eCommerce: Orders - ORM, integration with external APIs<br>
+Simple e-commerce checkout flow implementation using external API integration with Stripe for payment processing.<br>
+Basic deployment setup on Amazon AWS with Spring Boot application.
 
+### Step 2
 
-## Feature List
-### Step 1-2 - first Refactoring
+#### start.sh
```
da-nyee, line 8: 👍

src/main/kotlin/ecommerce/Application.kt
da-nyee: I thought it would be better to keep the configuration close to where it's used.
However, as you mentioned, it still registers as a global bean, so that wouldn't really make sense.
The current version is a more appropriate approach 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,90 @@
...
+import java.time.Duration
+
+@Component
+class StripeClient(private val stripeProperties: StripeProperties) {
+    private val restClient =
+        RestClient.builder()
+            .requestFactory(
+                SimpleClientHttpRequestFactory().apply {
+                    setConnectTimeout(Duration.ofMillis(500))
+                    setReadTimeout(Duration.ofSeconds(10))
```
da-nyee: Just curious — is there any particular reason you set the timeouts to 500ms and 10s respectively?

src/main/kotlin/ecommerce/exception/StripeException.kt
```diff
@@ -0,0 +1,13 @@
...
+import ecommerce.dto.StripeErrorInfo
+
+sealed class StripeException(message: String, override val cause: Throwable? = null) : RuntimeException(message) {
+    class Server(info: StripeErrorInfo, cause: Throwable? = null) :
+        StripeException("Stripe server error: ${info.message} (code: ${info.code})", cause)
+
+    class Client(info: StripeErrorInfo, cause: Throwable? = null) :
+        StripeException("Stripe client error: ${info.message} (code: ${info.code})", cause)
+
+    class Other(override val message: String, override val cause: Throwable? = null) : StripeException(message, cause)
```
da-nyee: Nice use of `sealed class` 👍

src/main/kotlin/ecommerce/client/StripeClient.kt
da-nyee: You refactored the structure well.
However, what I intended was to handle exceptions such as `CardException`, `InvalidRequestException`, or others mentioned in the docs I shared.

src/main/kotlin/ecommerce/service/CheckoutService.kt
da-nyee: Yes, I think you understood it very well 👍 

That could cause data inconsistency problems, as you mentioned.
It would also lead to long response times, which may result in a poor user experience or even timeout issues.
So, it's better to separate business logic from external API call logic.

There are good approaches for this, as you commented — such as separating them into different classes or adopting an event-driven structure.

da-nyee (changes requested):
Hi Mina! 🙌

I left a few more comments and replied to your comments, so please check them out when you have time.

It seems there's only one more step to go — keep it up until then 💪

src/main/kotlin/ecommerce/Application.kt
miooo0o: I see! Thank you. 
I used this annotation without much thought, was nice to be able to look it up and think about it while looking at your question :)

src/main/kotlin/ecommerce/client/StripeClient.kt
miooo0o: Oh, I see! If I had known this earlier, it could have helped my team project! Good to learn this now.

```kotlin
        } catch (e: StripeException) {
            val info = parseStripeError(e)
            when (e) {
                is CardException,
                is PermissionException,
                is RateLimitException,
                is InvalidRequestException,
                is AuthenticationException,
                is IdempotencyException,
                is SignatureVerificationException,
                -> throw StripeApiException.Client(info, e)

                is ApiConnectionException,
                is ApiException,
                -> throw StripeApiException.Server(info, e)

                else -> throw StripeApiException.Other(info, e)
            }
        } catch (e: Exception) {
            throw StripeApiException.Other(parseStripeError(e), e)
        }
```

I applied this!

src/main/kotlin/ecommerce/service/CheckoutService.kt
miooo0o: Thank you for sharing! 😆 There are often things I wish I had known a bit earlier, and this is definitely one of them. But I'm still happy to learn about it and expand my existing knowledge! If you have any related resources or references you'd recommend, please share them...!

src/main/kotlin/ecommerce/client/StripeClient.kt
miooo0o: This isn't exactly 'the reason,' but I thought the timeout duration should be very short (especially since this checkout process wraps the entire flow and users can't interact with anything). 

Looking through the docs, it seems there are separate recommended timeouts. I fixed it by referring to this [stripe issues/817](https://github.com/stripe/stripe-java/issues/817)

```kotlin
setConnectTimeout(Duration.ofSeconds(30))
setReadTimeout(Duration.ofSeconds(80))
```

src/main/kotlin/ecommerce/exception/StripeException.kt
miooo0o: Thank you :)

miooo0o: I updated this to apply your review [here](https://github.com/herotechcourse/spring-ecommerce-order/pull/34#discussion_r2304159078)

miooo0o:
Hello Dani!
Thank you for taking the time to review. It was a huge help. 
Especially with the asynchronous, transaction, and event-driven! I was able to directly experience why such design is necessary.

### Changes Made
1. Added Dependency
```kotlin
dependencies {
    implementation("com.stripe:stripe-java:24.16.0")
    ...
```
Used this to import Stripe Exceptions.

2. Applied StripeException
```kotlin
   } catch (e: StripeException) {
            val info = parseStripeError(e)
            when (e) {
                is CardException,
                is PermissionException,
...
```
Properly parses errors from Stripe and passes them as messages.

3. Changed Stripe Client Timeout Criteria
Since it wasn't clearly documented in Stripe's documentation, I looked into it and applied it. While it's a default value and might not need to be exposed, I thought it aligned with the direction of the modifications, so I left it in.

### Brief Reflection
I really came to think that I should read documentation more carefully. Spring made me already overwhelming, and since this was my first time working with external APIs, I feel like I missed important parts -- Stripe itself. But I was genuinely happy to have learned this time.

Thanks to the feedback, I learned a lot :)

src/main/kotlin/ecommerce/client/StripeClient.kt
```diff
@@ -0,0 +1,133 @@
...
+
+                is ApiConnectionException,
+                is ApiException,
+                -> throw StripeApiException.Server(info, e)
+
+                else -> throw StripeApiException.Other(info, e)
+            }
+        } catch (e: Exception) {
+            throw StripeApiException.Other(parseStripeError(e), e)
+        }
```
da-nyee, line 80: 👍

da-nyee (approved):
Hi Mina! 🙌

It's finally time to wrap up the last mission — you nailed it 💯 
I hope the feedback was helpful for your growth, and that you keep your passion for software development.

Congratulations on your graduation! 🎉

